PR #4, opened 2015-11-17
YD-50 embed buddies in user

aristotelos:


appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -119,6 +135,24 @@ public static Link getUserLink(UUID userID, boolean includePrivateData) {
...
+			if(getContent().getPrivateData() != null)
+			{
+				Optional<Set<BuddyDTO>> buddies = getContent().getPrivateData().getBuddies();
+				if(buddies.isPresent()) {
+					return Collections.singletonMap(UserDTO.BUDDIES_REL_NAME,
+						new BuddyController.BuddyResourceAssembler(getContent().getID()).toResources(buddies.get())); 
+				}					
+			}
+			return Collections.emptyMap();
+		}
```
Bert-R: Remove the nesting here by moving the return of the empty collection up into the first if statement.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -119,6 +135,24 @@ public static Link getUserLink(UUID userID, boolean includePrivateData) {
 		public UserResource(UserDTO user) {
 			super(user);
 		}
+		
+		@JsonProperty("_embedded")
+		public Map<String, List<BuddyController.BuddyResource>> getEmbeddedResources() {
+			if(getContent().getPrivateData() != null)
+			{
+				Optional<Set<BuddyDTO>> buddies = getContent().getPrivateData().getBuddies();
```
Bert-R: Don't use ``Optional`` on a collection but use an empty collection instead.

core/src/main/java/nu/yona/server/subscriptions/service/UserPrivateDTO.java
```diff
@@ -33,7 +36,8 @@
 	private UUID anonymousMessageSourceID;
 	private UUID anonymousMessageDestinationID;
 	private Set<UUID> buddyIDs;
-
+	private Optional<Set<BuddyDTO>> buddies;
```
Bert-R: Don't use optional on collections.

PR #6, opened 2015-11-17
YD-93 Refactor design as per wiki

Bert-R:
The code now reflects the design as per [this wiki comment](http://wiki.yona.nu/display/DEV/Architecture?focusedCommentId=13107331#comment-13107331)
YD-93 #under review


core/src/main/java/nu/yona/server/analysis/entities/GoalConflictMessage.java
```diff
@@ -53,25 +48,19 @@ public String getURL() {
...
 	}
 
-	public UUID getAccessorID() {
-		return accessorID;
-	}
-
 	@Override
 	public void encrypt(Encryptor encryptor) {
-		accessorIDCiphertext = encryptor.encrypt(accessorID);
 		goalIDCiphertext = encryptor.encrypt(goalID);
```
aristotelos: Does the goal connection still need to be encrypted?

PR #7, opened 2015-11-17
YD-82 load goals from file

aristotelos:


adminservice/data/goals.json
```diff
@@ -0,0 +1,16 @@
...
+    ]
+  },
+  {
+    "name": "news",
+    "categories": [
+      "news/media",
+      "newsgroups/forums"
+    ]
+  }
+]
```
Bert-R: If you put the content here that's being created by the setup test of the Admin service, then our tests continue to work without any changes.

adminservice/src/intTest/groovy/nu/yona/server/InitialGoalCreationTest.groovy
```diff
@@ -1,6 +1,7 @@
 package nu.yona.server
 
 import groovyx.net.http.RESTClient
+import groovyx.net.http.HttpResponseException
```
Bert-R: This test might go away. The initialization work is done through your new code. The testing work needs to happen against the /goals/ service in the appservice. The update feature will disappear, as we now have the file to load the goals.

aristotelos: Ah, I did not know that you intended to remove the initial functionality of
add/remove goals from the adminservice. In that case I can indeed remove a
lot of code. Do you want to remove the entire GoalController, or leave only
the method getAllGoals() there? Do you want to remove the public
add/delete/update methods from the GoalService too?

Bert-R: Let's for now retain the ``getAllGoals``.

adminservice/src/intTest/groovy/nu/yona/server/InitialGoalCreationTest.groovy
```diff
@@ -14,6 +15,26 @@ class InitialGoalCreationTest extends Specification {
...
 	YonaServer yonaServer = new YonaServer(baseURL)
 
+	@Shared
+	def initialGoalsCount
+	@Shared
+	def gamblingURL
+	@Shared
+	def programmingURL
+
+	def 'Get all goals'(){
```
Bert-R: This test was already there and is now a duplicate (error).

core/src/main/java/nu/yona/server/goals/service/GoalService.java
```diff
@@ -63,4 +73,39 @@ private Goal getEntityByID(UUID id) {
...
+	@Transactional
+	public void importGoals(Set<GoalDTO> goalDTOs) {
+		Set<Goal> goalsInRepository = getAllGoalEntities();				
+		deleteRemovedGoals(goalsInRepository, goalDTOs);
+		addOrUpdateNewGoals(goalDTOs, goalsInRepository);
+	}
+
+	private void addOrUpdateNewGoals(Set<GoalDTO> goalDTOs, Set<Goal> goalsInRepository) {
+		Map<String, Goal> goalsInRepositoryMap = new HashMap<String, Goal>();
+		for(Goal goalInRepository : goalsInRepository) goalsInRepositoryMap.put(goalInRepository.getName(), goalInRepository);
```
Bert-R: These two lines can easily be replaced with streaming:
```
		Map<String, Goal> goalsInRepositoryMap = goalsInRepository.stream()
				.collect(Collectors.toMap(g -> g.getName(), g -> g));
```

core/src/main/java/nu/yona/server/goals/service/GoalService.java
```diff
@@ -63,4 +73,39 @@ private Goal getEntityByID(UUID id) {
...
+
+	private void deleteRemovedGoals(Set<Goal> goalsInRepository, Set<GoalDTO> goalDTOs) {
+		Map<String, GoalDTO> goalDTOsMap = new HashMap<String, GoalDTO>();
+		for(GoalDTO goalDTO : goalDTOs) goalDTOsMap.put(goalDTO.getName(), goalDTO);
+		
+		for(Goal goalInRepository : goalsInRepository) {
+			if(!goalDTOsMap.containsKey(goalInRepository.getName())) {
+				deleteGoal(goalInRepository);
+			}
+		}
```
Bert-R: This method can be replace with two streaming operations:
```
		Map<String, GoalDTO> goalDTOsMap = goalDTOs.stream().collect(Collectors.toMap(g -> g.getName(), g -> g));

		goalsInRepository.stream().filter(g -> !goalDTOsMap.containsKey(g.getName())).forEach(this::deleteGoal);
```

PR #14, opened 2015-11-19
YD-54 add remove device

aristotelos:


appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -90,6 +91,93 @@
 		return CryptoSession.execute(password, () -> createResponse(userService.addUser(user), true, HttpStatus.CREATED));
 	}
 
+	private void CheckPassword(Optional<String> password, UUID id)
```
Bert-R: In Java, methods start with lowercase.

Bert-R: Rename id into ``userID``

core/src/main/java/nu/yona/server/subscriptions/entities/NewDeviceRequest.java
```diff
@@ -0,0 +1,83 @@
...
+@Entity
+@Table(name = "NEW_DEVICE_REQUESTS")
+public class NewDeviceRequest extends EntityWithID
+{
+	public static NewDeviceRequestRepository getRepository()
+	{
+		return (NewDeviceRequestRepository) RepositoryProvider.getRepository(NewDeviceRequest.class, UUID.class);
+	}
+
+	private Date expirationDateTime;
```
Bert-R: Please make this a ``creationDateTime``. That makes changing the cleanup policy more flexible and provides more information when debugging the system.

core/src/main/java/nu/yona/server/subscriptions/service/NewDeviceRequestDTO.java
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	public String getUserPassword()
+	{
+		return userPassword;
+	}
+
+	public static NewDeviceRequestDTO createInstance(NewDeviceRequest newDeviceRequestEntity)
+	{
+		if (newDeviceRequestEntity == null)
```
Bert-R: Why do you support creating a DTO when you don't have an entity?

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
```diff
@@ -0,0 +1,145 @@
...
+		given:
+
+		when:
+			def response = appService.setNewDeviceRequest(richardQuinURL, richardQuinPassword, """{
+				"userSecret":"unknown secret"
+			}""")
+			newDeviceRequestExpirationDateTime = response.responseData.expirationDateTime
+
+		then:
+			response.status == 200
```
Bert-R: This should be 201 when a new request is created.

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
```diff
@@ -0,0 +1,145 @@
...
+	def 'Try set new device request with wrong password'(){
+		given:
+
+		when:
+			try {
+				appService.setNewDeviceRequest(richardQuinURL, "foo", """{
+					"userSecret":"known secret"
+				}""")
+				assert false;
+			} catch (HttpResponseException e) {
```
Bert-R: Phillip did something in the test to prevent the HTTP user agent from throwing an exception upon a failure status code. That would be helpful here too.

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
```diff
@@ -0,0 +1,145 @@
...
+	def 'Add user Richard Quin'(){
+		given:
+
+		when:
+			def response = appService.addUser("""{
+				"firstName":"Richard",
+				"lastName":"Quin",
+				"nickName":"RQ",
+				"emailAddress":"rich@quin.net",
+				"mobileNumber":"+12345678",
```
Bert-R: Set the mobile number as a timestamp. For debuggability, it's good to add that to the first and last name too.

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
```diff
@@ -0,0 +1,145 @@
...
+	}
+	
+	def 'Set new device request'(){
+		given:
+
+		when:
+			def response = appService.setNewDeviceRequest(richardQuinURL, richardQuinPassword, """{
+				"userSecret":"unknown secret"
+			}""")
+			newDeviceRequestExpirationDateTime = response.responseData.expirationDateTime
```
Bert-R: Add an if to ensure the status is OK

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
```diff
@@ -0,0 +1,145 @@
...
+
+		when:
+			def response = appService.getNewDeviceRequest(richardQuinURL, richardQuinPassword, "unknown secret")
+			
+		then:
+			response.status == 200
+			response.responseData.userPassword == richardQuinPassword
+	}
+	
+	def 'Try set new device request with wrong password'(){
```
Bert-R: Adding a test for GET with wrong password would make sense

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -90,6 +91,55 @@
...
+	}
+
+	public static class NewDeviceRequestResource extends Resource<NewDeviceRequestDTO>
+	{
+		public NewDeviceRequestResource(NewDeviceRequestDTO newDeviceRequest)
+		{
+			super(newDeviceRequest);
+		}
+	}
+
```
Bert-R: Add a resource assembler, so a self-link is created.

core/src/main/java/nu/yona/server/subscriptions/service/NewDeviceRequestNotPresentException.java
```diff
@@ -0,0 +1,17 @@
...
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(value = HttpStatus.NOT_FOUND, reason = "No new device request present")
+public class NewDeviceRequestNotPresentException extends RuntimeException
+{
+	private static final long serialVersionUID = -7070143633618007280L;
+
+	public NewDeviceRequestNotPresentException(UUID id)
```
Bert-R: Rename ``id`` into ``userID``.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -75,6 +80,28 @@ public UserDTO addUser(UserDTO userResource)
...
+	{
+		User originalUserEntity = getEntityByID(forUserID);
+		Set<String> deviceNames = new TreeSet<String>(originalUserEntity.getDeviceNames());
+		deviceNames.remove(deviceName);
+		originalUserEntity.setDeviceNames(deviceNames);
+		User updatedUserEntity = User.getRepository().save(originalUserEntity);
+		return updatedUserEntity.getDeviceNames();
+	}
+
+	@Transactional
```
Bert-R: The methods ``addDeviceForUser`` and ``deleteDeviceForUser`` can be removed.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -230,4 +257,63 @@ public void addBuddy(UserDTO user, BuddyDTO buddy)
...
+	{
+		User userEntity = getEntityByID(id);
+		Date expirationDateTime = getNewDeviceRequestExpirationDateTime();
+		NewDeviceRequest newDeviceRequestEntity = NewDeviceRequest.createInstance(userPassword, expirationDateTime);
+		newDeviceRequestEntity.encryptUserPassword(userSecret);
+		userEntity.setNewDeviceRequest(newDeviceRequestEntity);
+		return NewDeviceRequestDTO.createInstance(User.getRepository().save(userEntity).getNewDeviceRequest());
+	}
+
+	private Date getNewDeviceRequestExpirationDateTime()
```
Bert-R: This method can be removed.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -230,4 +257,63 @@ public void addBuddy(UserDTO user, BuddyDTO buddy)
...
+		c.setTime(now);
+		c.add(Calendar.DATE, 1);
+		Date expirationDateTime = c.getTime();
+		return expirationDateTime;
+	}
+
+	@Transactional
+	public NewDeviceRequestDTO getNewDeviceRequestForUser(UUID id, String userSecret)
+	{
+		if (StringUtils.isNotBlank(userSecret))
```
Bert-R: make this ``isNotBlank`` and swap the if-else. Positive conditions read a lot easier.

aristotelos: Done, this was a typo.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
aristotelos: Done

core/src/main/java/nu/yona/server/subscriptions/service/NewDeviceRequestNotPresentException.java
aristotelos: Done

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
aristotelos: Done

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
aristotelos: Done

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
aristotelos: Done

core/src/main/java/nu/yona/server/subscriptions/service/NewDeviceRequestDTO.java
aristotelos: Reason was that the client can do a valid call to check if there are any live requests to display it on the app to the user. Now throwing exception and returning 404 (the client can use that too).

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
aristotelos: Done

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
aristotelos: Done

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
aristotelos: Checked HTTP standard, this is right, so did it.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
aristotelos: Done, but you don't need a resource assembler, only changed the resource constructor.

core/src/main/java/nu/yona/server/subscriptions/entities/NewDeviceRequest.java
aristotelos: Reject. The app may need to display the expiration date if they want to display open requests to add new devices.
What can be done to make it more flexible is to let the app post the expirationDateTime. That way the app designers can choose for how long the request is valid. (However that may impose a security risk.)

Bert-R: The date stored should be the creation date. You can return the expiration
date by adding the interval to it.

Groetjes, Bert

aristotelos: But then you get a difference between the DTO and the entity, which is not nice. Moreover the cleanup policy needs the expiration date to cleanup.

Bert-R: The app doesn't need to know when the request expires. We won't show that
to the user. The cleanup process should use the configuration setting to
see what to remove. Storing the expiration date is like storing someone's
age instead of birthday.

Groetjes, Bert

aristotelos: OK, so the app does not need to know if there is a request still open or not? Could you validate that with the PO?
I disagree that expiration date is like age. It is like date of death.

Bert-R: The app simply can use it if it exists. If the expired items sweeper has
taken it, it's gone. The app doesn't need to worry about expiration.

Groetjes,
Bert

aristotelos:
OK thanks for refactoring and merging! I thought we could discuss it
tomorrow but when I see the changes you made it becomes more clear to me
what you intended. My perspective was more from security viewpoint: it is
important that such a request is not open for a long time because it makes
the user vulnerable. Therefore my design was different.

PR #16, opened 2015-11-19
YD-101 remove email address from user

Bert-R:


appservice/src/intTest/groovy/nu/yona/server/BasicBuddyTest.groovy
```diff
@@ -132,11 +134,16 @@ class BasicBuddyTest extends Specification {
...
 			def response = appService.getDirectMessages(bobDunnURL, bobDunnPassword)
-			bobDunnBuddyMessageAcceptURL = response.responseData._embedded.buddyConnectRequestMessages[0]._links.accept.href
+			if (response.responseData._embedded && response.responseData._embedded.buddyConnectRequestMessages) {
+				bobDunnBuddyMessageAcceptURL = response.responseData._embedded.buddyConnectRequestMessages[0]._links.accept.href
+			}
 
 		then:
 			response.status == 200
 			response.responseData._links.self.href == bobDunnURL + appService.DIRECT_MESSAGE_PATH_FRAGMENT
+			response.responseData._embedded
```
aristotelos: The test results already shows the per path fragment what its value is, so this is not necessary

core/src/main/java/nu/yona/server/subscriptions/service/UserDTO.java
```diff
@@ -53,6 +51,11 @@ public UserDTO(@JsonProperty("firstName") String firstName, @JsonProperty("lastN
 		this(null, firstName, lastName, emailAddress, mobileNumber, privateData);
 	}
 
+	private UserDTO(UUID id, String firstName, String lastName, String mobileNumber, UserPrivateDTO privateData)
+	{
+		this(id, firstName, lastName, null, mobileNumber, privateData);
+	}
+
 	private UserDTO(UUID id, String firstName, String lastName, String emailAddress, String mobileNumber,
```
aristotelos: I would like less constructors. Maybe remove email from the constructors and add a setter for this purpose?

core/src/main/java/nu/yona/server/subscriptions/service/UserDTO.java
```diff
@@ -30,19 +30,17 @@
 	private final String mobileNumber;
```
aristotelos: I would like to rename emailAddress to emailAddressForInvitation or something like that to prevent misunderstanding.

PR #21, opened 2015-11-23
YD-51 add new user on buddy request

aristotelos:
Things to do later:
- finish EmailServer
- think about scenario with second buddy adding request
- think about if scenario for forgetting the invite email and signing up yourself is complete


appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -0,0 +1,344 @@
...
+				"nickName":"RQ ${timestamp}",
+				"mobileNumber":"+${timestamp}11",
+				"devices":[
+					"Nexus 6"
+				],
+				"goals":[
+					"news"
+				]
+			}""", richardQuinPassword)
+			richardQuinURL = appService.stripQueryString(response.responseData._links.self.href)
```
Bert-R: Only do this when the status is 201 and ``_links`` exists.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -0,0 +1,344 @@
...
+			response.responseData.firstName == "Bob ${timestamp}"
+			response.responseData.lastName == "Dun ${timestamp}"
+			response.responseData.mobileNumber == "+${timestamp}12"
+			response.responseData.nickName == "BD ${timestamp}"
+			response.responseData.devices.size() == 1
+			response.responseData.devices[0] == "iPhone 6"
+			//TODO: updating of goals is not yet supported
+			//response.responseData.goals.size() == 1
+			//response.responseData.goals[0] == "gambling"
+			//TODO: buddy is not added yet, needs to be fixed
```
Bert-R: This TODO is addressed already

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -0,0 +1,344 @@
...
+
+		then:
+			response.status == 200
+			response.responseData.firstName == "Bob ${timestamp}"
+			response.responseData.lastName == "Dun ${timestamp}"
+			response.responseData.mobileNumber == "+${timestamp}12"
+			response.responseData.nickName == "BD ${timestamp}"
+			response.responseData.devices.size() == 1
+			response.responseData.devices[0] == "iPhone 6"
+			//TODO: updating of goals is not yet supported
```
Bert-R: These TODOs aren't relevant here.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -0,0 +1,344 @@
...
+	def richardQuinPassword = "R i c h a r d"
+	def bobDunnPassword = "B o b"
+	@Shared
+	def richardQuinURL
+	@Shared
+	def richardQuinLoginID
+	@Shared
+	def bobDunnGmail = "bobdunn325@gmail.com"
+	@Shared
+	def bobDunnGmailPassword = "bobbydunn"
```
Bert-R: Someone will find this password and hijack the account. Let's add a configuration property that indicates whether we're running production or test. In case of test, return the URL in the buddy DTO.

appservice/src/main/java/nu/yona/server/subscriptions/rest/BuddyController.java
```diff
@@ -89,11 +92,19 @@ static ControllerLinkBuilder getAllBuddiesLinkBuilder(UUID requestingUserID)
...
-				() -> createResponse(requestingUserID, buddyService.addBuddyToRequestingUser(requestingUserID, buddy),
+		return CryptoSession.execute(
+				password,
+				() -> userService.canAccessPrivateData(requestingUserID),
+				() -> createResponse(requestingUserID, buddyService.addBuddyToRequestingUser(requestingUserID, buddy, this),
 						HttpStatus.CREATED));
 	}
 
+	@Override
+	public String getInviteURL(UUID newUserID, String tempPassword)
```
Bert-R: Please rename in ``getInviteURLString``, as it's not returning an URL object.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -185,12 +210,19 @@ public void deleteUser(@RequestHeader(value = Constants.PASSWORD_HEADER) Optiona
...
 	}
 
+	public static Link getUserLinkWithTempPassword(UUID userID, String tempPassword)
+	{
+		ControllerLinkBuilder linkBuilder = linkTo(methodOn(UserController.class).updateUser(Optional.empty(), tempPassword,
+				userID, null));
+		return linkBuilder.withSelfRel();
+	}
+
 	public static Link getUserLink(UUID userID, boolean includePrivateData)
```
Bert-R: Make this private. Nobody else uses it.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -185,12 +210,19 @@ public void deleteUser(@RequestHeader(value = Constants.PASSWORD_HEADER) Optiona
 		return createResponse(user, includePrivateData, HttpStatus.OK);
 	}
 
+	public static Link getUserLinkWithTempPassword(UUID userID, String tempPassword)
```
Bert-R: Rename into ``getUserSelfLinkWithTempPassword`` or allow specifying the ``rel`` as parameter.

core/src/main/java/nu/yona/server/subscriptions/entities/User.java
```diff
@@ -101,6 +101,11 @@ public boolean isCreatedOnBuddyRequest()
 		return createdOnBuddyRequest;
 	}
 
+	public void removeIsCreatedOnBuddyRequest()
```
Bert-R: Call this ``unset...`` or ``clear...``. It's not removing anything.

core/src/main/java/nu/yona/server/subscriptions/entities/User.java
```diff
@@ -151,7 +156,7 @@ public void setNewDeviceRequest(NewDeviceRequest newDeviceRequest)
 		this.newDeviceRequest = newDeviceRequest;
 	}
 
-	private UserPrivate getUserPrivate()
+	public UserPrivate getUserPrivate()
```
Bert-R: What do you use this for? Currently, ``User`` hides the existence of ``UserPrivate``.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -59,21 +60,36 @@
 	@RequestMapping(value = "{id}", params = { "includePrivateData" }, method = RequestMethod.GET)
 	@ResponseBody
 	public HttpEntity<UserResource> getUser(@RequestHeader(value = PASSWORD_HEADER) Optional<String> password,
+			@RequestParam(value = "tempPassword", required = false) String tempPassword,
```
Bert-R: Given that the temp password is optional, can't you use ``Optional``?

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -59,21 +60,36 @@
...
+	{
+		if (password.isPresent())
+		{
+			return password;
+		}
+		if (StringUtils.isNotBlank(tempPassword))
+		{
+			return Optional.of(tempPassword);
+		}
+		return null;
```
Bert-R: Don't return null for an optional, as that defeats the purpose of ``Optional``.

core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
```diff
@@ -31,6 +32,9 @@
 
 	private static final String DECRYPTION_CHECK_STRING = "Decrypted properly#";
 
+	@Column(nullable = true)
```
Bert-R: Why is this nullable?

core/src/main/java/nu/yona/server/subscriptions/entities/Buddy.java
```diff
@@ -32,6 +33,9 @@ public static BuddyRepository getRepository()
 		return (BuddyRepository) RepositoryProvider.getRepository(Buddy.class, UUID.class);
 	}
 
+	@Column(nullable = true)
```
Bert-R: Why is this nullable?

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -76,22 +83,25 @@ public UserDTO addUser(UserDTO userResource)
 		return UserDTO.createInstanceWithPrivateData(userEntity);
 	}
 
+	@Autowired
+	UserServiceTempEncryptionContextExecutor tempEncryptionContextExecutor;
+
+	public User addUserCreatedOnBuddyRequest(UserDTO buddyUserResource, String tempPassword)
```
Bert-R: Shouldn't this be transactional?

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -76,22 +83,25 @@ public UserDTO addUser(UserDTO userResource)
...
 		if (originalUserEntity.isCreatedOnBuddyRequest())
 		{
-			// TODO: the password needs to come from the URL, not from the
-			// payload
-			savedUser = handleUserSignupUponBuddyRequest("TODO", userResource, originalUserEntity);
-		}
-		else
-		{
-			savedUser = handleUserUpdate(userResource, originalUserEntity);
+			throw new YonaException("User is created on buddy request, use other method");
```
Bert-R: Throw ``IllegalArgumentException``, as this is a programming error, and change the text into something like this: "Cannot update 'user created on buddy request' through this method".

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -105,54 +115,63 @@ static User findUserByMobileNumber(String mobileNumber)
...
-	private UserDTO handleUserSignupUponBuddyRequest(String password, UserDTO userResource, User originalUserEntity)
+	@Transactional
+	public UserDTO updateUserCreatedOnBuddyRequest(UUID id, String tempPassword, UserDTO userResource)
 	{
-		UpdatedEntities updatedEntities = updateUserWithTempPassword(userResource, originalUserEntity);
-		return saveUserWithDevicePassword(password, updatedEntities);
+		User originalUserEntity = getEntityByID(id);
+		if (!originalUserEntity.isCreatedOnBuddyRequest())
+		{
+			throw new YonaException("User is not created on buddy request");
```
Bert-R: Throw ``IllegalArgumentException``, as this is a programming error, and change the text into something like this: "Can only update 'user created on buddy request' through this method".

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -76,22 +83,25 @@ public UserDTO addUser(UserDTO userResource)
...
+				.addUserCreatedOnBuddyRequestFlush(buddyUserResource).getID());
+		return getEntityByID(savedUserID);
+	}
+
 	@Transactional
 	public UserDTO updateUser(UUID id, UserDTO userResource)
 	{
 		User originalUserEntity = getEntityByID(id);
-		UserDTO savedUser;
 		if (originalUserEntity.isCreatedOnBuddyRequest())
```
Bert-R: Replace these 4 lines with ``assert originalUserEntity.isCreatedOnBuddyRequest()``.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -105,54 +115,63 @@ static User findUserByMobileNumber(String mobileNumber)
...
 	}
 
-	private UserDTO handleUserSignupUponBuddyRequest(String password, UserDTO userResource, User originalUserEntity)
+	@Transactional
+	public UserDTO updateUserCreatedOnBuddyRequest(UUID id, String tempPassword, UserDTO userResource)
 	{
-		UpdatedEntities updatedEntities = updateUserWithTempPassword(userResource, originalUserEntity);
-		return saveUserWithDevicePassword(password, updatedEntities);
+		User originalUserEntity = getEntityByID(id);
+		if (!originalUserEntity.isCreatedOnBuddyRequest())
```
Bert-R: Replace these 4 lines with ``assert ~originalUserEntity.isCreatedOnBuddyRequest()``.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -76,22 +83,25 @@ public UserDTO addUser(UserDTO userResource)
...
-			// payload
-			savedUser = handleUserSignupUponBuddyRequest("TODO", userResource, originalUserEntity);
-		}
-		else
-		{
-			savedUser = handleUserUpdate(userResource, originalUserEntity);
+			throw new YonaException("User is created on buddy request, use other method");
 		}
-		return savedUser;
+		return handleUserUpdate(userResource, originalUserEntity);
```
Bert-R: Given that ``handleUserUpdate`` is only called here, while this method is otherwise empty, you can inline it here (<ctrl><shift>I).

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -105,54 +115,63 @@ static User findUserByMobileNumber(String mobileNumber)
...
+		}
+		UserEncryptedEntitySet retrievedEntitySet = retrieveUserEncryptedData(originalUserEntity, tempPassword);
+		return saveUserEncryptedDataWithNewPassword(retrievedEntitySet, userResource);
 	}
 
-	static class UpdatedEntities
+	/*
+	 * Gathers all entities that contain encrypted data from the database.
+	 */
+	static class UserEncryptedEntitySet
```
Bert-R: This name confuses me. What about ``EncryptedUserData``?

Bert-R: Can this class be private?

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -105,54 +115,63 @@ static User findUserByMobileNumber(String mobileNumber)
...
-	{
-		// TODO: the password needs to come from the URL, not from the
-		// payload
-		return CryptoSession.execute(Optional.of("TODO"), () -> canAccessPrivateData(userDTO.getID()), () -> {
-			User updatedUserEntity = userDTO.updateUser(originalUserEntity);
-			MessageSource touchedNameMessageSource = touchMessageSource(updatedUserEntity.getNamedMessageSource());
-			MessageSource touchedAnonymousMessageSource = touchMessageSource(updatedUserEntity.getAnonymousMessageSource());
-			return new UpdatedEntities(updatedUserEntity, touchedNameMessageSource, touchedAnonymousMessageSource);
-		});
+		public void loadLazyEncryptedData()
```
Bert-R: Can be package scoped rather than public.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
aristotelos: It's used in BuddyController, so has to be internal.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
Bert-R: Move this class to the beginning or end of ``UserService``.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -212,6 +231,14 @@ public void addBuddy(UserDTO user, BuddyDTO buddy)
 		User.getRepository().save(userEntity);
 	}
 
+	private SecureRandom random = new SecureRandom();
```
Bert-R: Declare all attributes at the begin of the class.

core/src/main/java/nu/yona/server/subscriptions/service/UserServiceTempEncryptionContextExecutor.java
```diff
@@ -0,0 +1,41 @@
...
+	public User addUserCreatedOnBuddyRequestFlush(UserDTO buddyUserResource)
+	{
+		return User.getRepository().save(
+				User.createInstanceOnBuddyRequest(buddyUserResource.getFirstName(), buddyUserResource.getLastName(),
+						buddyUserResource.getPrivateData().getNickName(), buddyUserResource.getMobileNumber()));
+	}
+
+	// use a separate transaction to read within the crypto session
+	@Transactional(value = TxType.REQUIRES_NEW)
+	public UserEncryptedEntitySet retrieveUserEncryptedDataFlush(User originalUserEntity)
```
Bert-R: What does 'flush' mean here?

core/src/main/java/nu/yona/server/subscriptions/entities/User.java
aristotelos: Used in:
- UserDTO for updating the User (with the nested UserPrivateDTO)
- UserService for retrieving the user as a whole
I think these are required and OK. Let's discuss if you disagree.

core/src/testutils/groovy/nu/yona/server/YonaServer.groovy
```diff
@@ -191,10 +193,41 @@ class YonaServer {
 			contentType:'application/json',
 			headers: headers)
 	}
+	
+	def putJson(path, jsonString, headers = [:], parameters = [:])
```
Bert-R: Use tabs to indent.

core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
aristotelos: When I left that out I got a NOT NULL constraint violation from HSQLDB when trying to insert it. I tried to initialize the field with `private int touchVersion = 1;` but that didn't work. If you have a different suggestion that works please let me know.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
aristotelos: Yes I think so.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
aristotelos: Notice the app (or any other hacker) can arrive here if it uses the temp password as Yona-Password and could use the temp password forever...
So it might be a security issue: if someone intercepts the email, she may update the user without storing a new password. Maybe not an issue.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
aristotelos: On this one I think I disagree. If assertion checking is off the password of an existing user may be replaced. I consider this a security issue..

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
aristotelos: Renamed and moved. Cannot be private because used in `UserServiceTempEncryptionContextExecutor`.

core/src/main/java/nu/yona/server/subscriptions/service/UserServiceTempEncryptionContextExecutor.java
aristotelos: Flush means that a save is done in a new required transaction...

aristotelos: First method: saves in new required transaction.
Second method: retrieves in new required transaction. Is maybe not so nice here.

PR #22, opened 2015-11-23
YD-104 implement proper exceptions

pgussow:


core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -74,7 +75,7 @@ private UserAnonymized getUserAnonymizedByID(UUID id)
 		UserAnonymized entity = UserAnonymized.getRepository().findOne(id);
 		if (entity == null)
 		{
-			throw new LoginIDNotFoundException(id);
+			throw InvalidDataException.analysisUserNotFound(id);
```
Bert-R: Please stick with "Login ID", a that's the term we use throughout the code base.

PR #25, opened 2015-11-23
YD-51 Added some hacking tests

aristotelos:


appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -126,6 +129,57 @@ class CreateUserOnBuddyRequestTest extends Specification {
...
 	}
 	
+	def 'Hacking attempt: Try to get Bob Dunn with an invalid temp password'(){
+		given:
+
+		when:
+			def response = appService.getResource(bobDunnURL, [:], ["tempPassword": "hack", "includePrivateData": "true"])
+
+		then:
+			response.status != 200
```
Bert-R: Now the test passes when the response is another success code (2xx). What about adding a method ``YonaServer.isSuccess()`` and use that in all places were we only care about success or failure?

PR #29, opened 2015-11-25
YD-65 Implement delete of a message

2gmpoyhonen:


Bert-R:
Please sync with Master.

core/src/main/java/nu/yona/server/analysis/service/GoalConflictMessageDTO.java
```diff
@@ -28,6 +30,7 @@
 @JsonRootName("goalConflictMessage")
 public class GoalConflictMessageDTO extends MessageDTO
 {
+	private static final String DELETE = "delete";
```
Bert-R: The delete should be done the REST way: doing a DELETE on the message URL.

appservice/src/intTest/groovy/nu/yona/server/DeleteDirectMessageTest.groovy
```diff
@@ -0,0 +1,365 @@
...
+
+	def 'Richard tries to delete Bob\'s buddy acceptance before it is processed'(){
+		given:
+
+		when:
+			def response = appService.deleteResourceWithPassword(richardQuinBuddyMessageDeleteURL, richardQuinPassword)
+
+		then:
+			response.status == 400
+			response.responseData?.code == "error.cannot.delete.unprocessed.message"
```
aristotelos: Maybe you would like to not reply to a connect request and let it expire instead of rejecting it. This is what I sometimes do with Facebook/LinkedIn to avoid disappointing people...

Bert-R: That still doesn't imply that you can delete it. The cleanup program should
take care of that.

aristotelos: Agreed.

PR #30, opened 2015-11-25
YD-53 unsubscribe, YD-59 remove buddy

aristotelos:


core/src/main/java/nu/yona/server/subscriptions/entities/Buddy.java
```diff
@@ -133,10 +138,10 @@ public Status getSendingStatus()
 		return getBuddyAnonymized().getSendingStatus();
 	}
 
-	public void setLoginID(UUID loginID)
+	public void setUserAnonymizedID(UUID userAnonymizedID)
```
Bert-R: This should be login ID, as that's the functional term.

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyAnonymized.java
```diff
@@ -81,13 +81,8 @@ public void setReceivingStatus(Status receivingStatus)
 		this.receivingStatus = receivingStatus;
 	}
 
-	public UUID getLoginID()
+	public UUID getUserAnonymizedID()
```
Bert-R: Same here.

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectRemoveMessage.java
```diff
@@ -0,0 +1,47 @@
+package nu.yona.server.subscriptions.entities;
+
+import java.util.UUID;
+
+import javax.persistence.Entity;
+
+import nu.yona.server.subscriptions.service.BuddyService.DropBuddyReason;
+
+@Entity
+public class BuddyConnectRemoveMessage extends BuddyMessage
```
Bert-R: This name is not logical. What about ``BuddyDisconnectMessage``?

core/src/main/java/nu/yona/server/subscriptions/service/BuddyConnectRemoveMessageDTO.java
```diff
@@ -0,0 +1,110 @@
...
+import nu.yona.server.messaging.service.MessageActionDTO;
+import nu.yona.server.messaging.service.MessageDTO;
+import nu.yona.server.messaging.service.MessageService.DTOManager;
+import nu.yona.server.messaging.service.MessageService.TheDTOManager;
+import nu.yona.server.subscriptions.entities.BuddyConnectRemoveMessage;
+import nu.yona.server.subscriptions.entities.User;
+import nu.yona.server.subscriptions.service.BuddyService.DropBuddyReason;
+
+@JsonRootName("buddyConnectRemoveMessage")
+public class BuddyConnectRemoveMessageDTO extends BuddyMessageDTO
```
Bert-R: This name is not logical. What about ``BuddyDisconnectMessageDTO``?

PR #31, opened 2015-11-25
YD-77 Number check as part of the sign up process

2gmpoyhonen:


appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
```diff
@@ -42,6 +42,25 @@ class UserTest extends Specification {
 			appService.deleteUser(appService.stripQueryString(response.responseData._links.self.href), password)
 	}
 
+	def 'Send sign-in confirmation code'(){
```
Bert-R: Rename "sign-in confirmation" into "mobile number confirmation".

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -112,6 +112,17 @@ private void checkPassword(Optional<String> password, UUID userID)
 		CryptoSession.execute(password, () -> userService.canAccessPrivateData(userID), () -> null);
 	}
 
+	@RequestMapping(value = "{userID}/confirm", method = RequestMethod.POST)
+	@ResponseBody
+	public HttpEntity<UserResource> addUserConfimation(
+			@RequestHeader(value = Constants.PASSWORD_HEADER) Optional<String> password, @PathVariable UUID userID,
+			@RequestBody UserSignInConfirmationDTO newDeviceRequestCreation)
```
Bert-R: Rename ``UserSignInConfirmationDTO`` into ``MobileNumberConfirmationDTO``

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -112,6 +112,17 @@ private void checkPassword(Optional<String> password, UUID userID)
...
+	{
+		checkPassword(password, userID);
+		userService.confirmSignIn(userID, newDeviceRequestCreation.getCode());
```
Bert-R: Rename ``confirmSignInUser`` into ``confirmMobileNumber``.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserSignInConfirmationDTO.java
```diff
@@ -0,0 +1,22 @@
+package nu.yona.server.subscriptions.rest;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonRootName;
+
+@JsonRootName("userSignInConfirmation")
+public class UserSignInConfirmationDTO
```
Bert-R: Rename into ``MobileNumberConfirmationDTO`` (don't forget the ``JsonRootName``).

core/src/main/java/nu/yona/server/crypto/CryptoUtil.java
```diff
@@ -25,7 +25,7 @@ public static String getRandomString(int length)
 		return randomString.substring(0, Math.min(length, randomString.length()));
 	}
 
-	static SecureRandom getSecureRandomInstance()
+	public static SecureRandom getSecureRandomInstance()
```
Bert-R: Make this private again and add a method ``int generateRandomNumber(int digits)`` instead.

core/src/main/java/nu/yona/server/exceptions/SignInValidationException.java
```diff
@@ -0,0 +1,34 @@
+package nu.yona.server.exceptions;
+
+/**
+ * This exception is to be used in case user sign-in code is wrong.
+ */
+public class SignInValidationException extends YonaException
```
Bert-R: Rename into ``MobileNumberConfirmationException``.

core/src/main/java/nu/yona/server/exceptions/SignInValidationException.java
```diff
@@ -0,0 +1,34 @@
...
+	}
+
+	private SignInValidationException(Throwable t, String messageId, Object... parameters)
+	{
+		super(t, messageId, parameters);
+	}
+
+	public static SignInValidationException signInCodeMismatch()
+	{
+		return new SignInValidationException("error.sign-in.code.mismatch");
```
Bert-R: Rename into ``error.mobile.number.confirmation.code.mismatch`` and use a similar pattern for the other resource IDs.

core/src/main/java/nu/yona/server/exceptions/MobileNumberConfirmationException.java
```diff
@@ -0,0 +1,34 @@
...
+	{
+		super(messageId, parameters);
+	}
+
+	private MobileNumberConfirmationException(Throwable t, String messageId, Object... parameters)
+	{
+		super(t, messageId, parameters);
+	}
+
+	public static MobileNumberConfirmationException signInCodeMismatch()
```
aristotelos: Forgot one! :)

core/src/main/java/nu/yona/server/exceptions/MobileNumberConfirmationException.java
```diff
@@ -0,0 +1,34 @@
...
+	{
+		return new MobileNumberConfirmationException("error.mobile.number.confirmation.code.mismatch");
+	}
+
+	public static MobileNumberConfirmationException signInCodeNotSet()
```
aristotelos: And this one...

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -83,10 +91,51 @@ public UserDTO addUser(UserDTO user)
 		user.getPrivateData().getVpnProfile().setVpnPassword(generatePassword());
 
 		User userEntity = user.createUserEntity();
+		userEntity.setConfirmationCode(CryptoUtil.getRandomDigits(5));
```
aristotelos: Maybe the number of 5 should be a setting?

core/src/main/java/nu/yona/server/subscriptions/entities/User.java
```diff
@@ -60,6 +68,7 @@ private User(UUID id, byte[] initializationVector, String firstName, String last
 			UserPrivate userPrivate, MessageDestination messageDestination)
 	{
 		super(id);
+		this.status = Status.UNCONFIRMED;
```
aristotelos: Currently UNCONFIRMED users still can create buddy requests. This maybe should be disabled because it is a threat for buddies if someone signs up with a false identity and invites them?

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -112,6 +112,17 @@ private void checkPassword(Optional<String> password, UUID userID)
 		CryptoSession.execute(password, () -> userService.canAccessPrivateData(userID), () -> null);
 	}
 
+	@RequestMapping(value = "{userID}/confirm", method = RequestMethod.POST)
+	@ResponseBody
+	public HttpEntity<UserResource> addUserConfimation(
+			@RequestHeader(value = Constants.PASSWORD_HEADER) Optional<String> password, @PathVariable UUID userID,
+			@RequestBody MobileNumberConfirmationDTO newDeviceRequestCreation)
```
aristotelos: newDeviceRequestCreation: please rename

core/src/main/java/nu/yona/server/subscriptions/entities/User.java
aristotelos: Added new task http://jira.yona.nu/browse/YD-113 for this.

core/src/main/java/nu/yona/server/subscriptions/entities/User.java
```diff
@@ -25,6 +25,11 @@
 @Table(name = "USERS")
 public class User extends EntityWithID
 {
+	public enum Status
```
aristotelos: Suppose the user quits the app before the confirmation. Then a user has been created but not confirmed. How will the app know that the user should still be confirmed when opening the app the second time? There is no GET request to get the status, and no error will be thrown.

(Resending the confirmation SMS is also not implemented yet. I think this may be necessary because SMSes may be lost on the way. But if we can drop the user and recreate it that is maybe sufficient for version 1.)

core/src/main/java/nu/yona/server/subscriptions/entities/User.java
```diff
@@ -25,6 +25,11 @@
...
+	{
+		UNCONFIRMED, INACTIVE, ACTIVE
```
aristotelos: INACTIVE and ACTIVE currently mean nothing. Please leave them out of the code for now until it is clear how we are going to use this. (Use boolean isConfirmed)

core/src/main/java/nu/yona/server/subscriptions/entities/User.java
aristotelos: If you change status to a boolean isConfirmed, the solution is to add this to the UserDTO.

PR #32, opened 2015-11-25
YD-105 Removed unneccesary code

aristotelos:


Bert-R:
That's a good cleanup!

PR #33, opened 2015-11-25
YD-48 Create VPN account

Bert-R:


core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
```diff
@@ -156,6 +160,11 @@ public UUID getLoginID()
 		return getUserAnonymized().getLoginID();
 	}
 
+	public String getPassword()
+	{
+		return password;
```
aristotelos: This is really confusing. We everywhere use the app stored password as "password", this is the VPN password I guess?

PR #37, opened 2015-11-27
YD-109 HTML email rendering by template

2gmpoyhonen:


core/src/main/java/nu/yona/server/email/EmailService.java
```diff
@@ -0,0 +1,49 @@
...
+
+				MimeMessageHelper message = new MimeMessageHelper(mimeMessage);
+				message.setFrom(new InternetAddress(senderAddress, senderName));
+				message.setTo(receiverAddress);
+				message.setSubject(subjectText);
+				message.setText(bodyText, true);
+			}
+		};
+
+		// TODO queue and continue because it may take time to send
```
Bert-R: Please create a ticket for this and mention that here.

core/src/main/java/nu/yona/server/email/EmailService.java
```diff
@@ -0,0 +1,49 @@
...
+	@Autowired
+	private VelocityEngine velocityEngine;
+	@Value("${yona.email.sender.address}")
+	private String senderAddress;
+
+	public void sendEmail(String senderName, InternetAddress receiverAddress, String subjectTemplateName, String bodyTemplateName,
+			Map<String, Object> templateParameters)
+	{
+		MimeMessagePreparator preparator = new MimeMessagePreparator() {
+			public void prepare(MimeMessage mimeMessage) throws Exception
```
Bert-R: Remove the ``throws`` here and put the try/catch of ``BuddyService`` inside.

core/src/main/java/nu/yona/server/email/EmailService.java
2gmpoyhonen: I have created task http://jira.yona.nu/browse/YD-114 for this. I also mention that in the comment.

aristotelos: I added this TODO, but on further consideration I think it is not necessary. It is OK if sending the email takes a little longer, and if it fails it is better to have that info in the app directly as Philip pointed out.

Bert-R: Agreed. I'll remove it, including the ticket.

Bert-R:
As I'm preparing the demo scenario for Tuesday today, I'll merge the property changes into this, address the review comments and pull the change set.

PR #38, opened 2015-11-27
YD-107 unify the property handling

Bert-R:
Along with this:
* LDAP can now be disabled
* Corrected ``UserDTO`` (didn't assign ``isConfirmed``, which caused a test case failure).
* Finished the rename from "confirm user" to "confirm mobile number"


Bert-R:
As I'm going to prepare the demo scenario for Tuesday today, I've pulled this myself.

PR #41, opened 2015-11-30
YD-116 Require mobile number for users invited through a buddy request

2gmpoyhonen:


core/src/main/java/nu/yona/server/subscriptions/service/BuddyDTO.java
```diff
@@ -139,4 +143,21 @@ public String getUserCreatedInviteURL()
...
 	}
+
+	/*
+	 * Only intended for test purposes.
+	 */
+	public String getConfirmationCode()
+	{
+		return confirmationCode;
+	}
+
```
Bert-R: Use the ``confirmationCode`` property of ``UserDTO`` instead.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -143,7 +143,9 @@ public User addUserCreatedOnBuddyRequest(UserDTO buddyUser, String tempPassword)
 	{
 		UUID savedUserID = CryptoSession.execute(Optional.of(tempPassword), null,
 				() -> tempEncryptionContextExecutor.addUserCreatedOnBuddyRequest(buddyUser).getID());
-		return getEntityByID(savedUserID);
+		User userEntity = getEntityByID(savedUserID);
+		sendMobileNumberConfirmationMessage(userEntity);
```
aristotelos: Should we not use a different SMS template here? Because the user may not have read her email yet, the SMS may be quite unexpected.

Bert-R: The SMS should be sent after the saved the data under the device password.
In that stage, the app should have informed the user.

PR #42, opened 2015-11-30
YD-121 Disconnecting remote Buddy on remove, removing anonymous messages

aristotelos:
Also checking for the case that the buddy connect request was not yet
accepted

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
@@ -145,9 +145,9 @@ public class UserPrivate extends EntityWithID
 		UserAnonymized.getRepository().save(userAnonymized);
 	}
 
-	public void removeBuddiesFromUser(UUID fromUserLoginID)
+	public void removeBuddiesFromUser(UUID fromUserID)
 	{
-		buddies.removeIf(buddy -> buddy.getVPNLoginID().equals(fromUserLoginID));
+		buddies.removeIf(buddy -> buddy.getUserID().equals(fromUserID));
 	}
 
 	public MessageSource getAnonymousMessageSource()
--- a/core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
@@ -145,9 +145,9 @@ public class UserPrivate extends EntityWithID
 		UserAnonymized.getRepository().save(userAnonymized);
 	}
 
-	public void removeBuddiesFromUser(UUID fromUserID)
+	public void removeBuddyForUserID(UUID userID)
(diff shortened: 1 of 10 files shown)


core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
```diff
@@ -145,9 +145,9 @@ public void removeBuddy(Buddy buddy)
 		UserAnonymized.getRepository().save(userAnonymized);
 	}
 
-	public void removeBuddiesFromUser(UUID fromUserLoginID)
+	public void removeBuddiesFromUser(UUID fromUserID)
```
Bert-R: There will be only one buddy object for a given user ID (these days). I'm renaming this into ``removeBuddyForUserID(UUID userID)``.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -103,14 +105,21 @@ public void removeBuddy(UUID idOfRequestingUser, UUID buddyID, Optional<String>
...
 
+	private void removeMessagesSentByBuddy(User user, Buddy buddy)
+	{
+		removeNamedMessagesSentByUser(user, buddy.getVPNLoginID());
+		removeAnonymousMessagesSentByUser(user.getAnonymized(), buddy.getVPNLoginID());
+	}
+
 	@Transactional
-	void clearMessagesAndSendDropBuddyMessage(User requestingUser, Buddy requestingUserBuddy, Optional<String> message,
+	void removeBuddyInfoFromBuddy(User requestingUser, Buddy requestingUserBuddy, Optional<String> message,
```
Bert-R: You're not removing the buddy from the buddy but for the given buddy. Renaming this into ``removeBuddyInfoForBuddy``.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -118,15 +127,29 @@ void clearMessagesAndSendDropBuddyMessage(User requestingUser, Buddy requestingU
...
+
+			UserAnonymized userAnonymized = UserAnonymized.getRepository().findOne(requestingUserBuddy.getVPNLoginID());
+			disconnectBuddy(userAnonymized, requestingUser.getVPNLoginID());
+			removeAnonymousMessagesSentByUser(userAnonymized, requestingUser.getVPNLoginID());
+			sendDropBuddyMessage(requestingUser, requestingUserBuddy, message, reason);
+		}
+		else
+		{
+			// remove sent buddy request message
+			removeNamedMessagesSentByUser(requestingUserBuddy.getUser(), requestingUser.getVPNLoginID());
```
Bert-R: This line is identical to the first one if the if-clause, so it can go before the if-statement.

PR #49, opened 2015-12-01
YD-123 Added translator

aristotelos:


core/src/main/java/nu/yona/server/Translator.java
```diff
@@ -0,0 +1,42 @@
...
+
+@Service
+public class Translator
+{
+	/** The source for the messages to use */
+	@Autowired
+	private MessageSource msgSource;
+
+	/**
+	 * This method returns the message for this exception based on the given locale
```
pgussow: What exception?

core/src/main/java/nu/yona/server/Translator.java
```diff
@@ -0,0 +1,42 @@
...
+	 * 
+	 * @return The actual message based on the default locale.
+	 */
+	public String getLocalizedMessage(String messageId, Object... parameters)
+	{
+		return getLocalizedMessage(messageId, parameters, null);
+	}
+
+	/**
+	 * This method returns the message for this exception based on the given locale
```
pgussow: here too, not an exception

PR #54, opened 2015-12-03
Yd 104 implement proper eceptions

pgussow:
All service classes have their input parameters checked and avoid null pointer exceptions


Bert-R:
Looks good. Thanks!

PR #55, opened 2015-12-03
YD-128 Run the goal file loader as part of the Docker image creation

Bert-R:


Bert-R:
Don't pull this yet. The Dockerfile needs a change.

Bert-R:
I made the change in the Dockerfile. If there's someone who can review and pull this, that would be great. If not, I'll do it myself later today.

pgussow:
I will take a look

dbinit/src/main/docker/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM yonadev/hsqldb:2.3.2
+
+ADD DatabaseInitializer.jar DatabaseInitializer.jar
+ADD data data
+
+RUN ["java","-jar","DatabaseInitializer.jar","--spring.datasource.url=jdbc:hsqldb:file:thedatabase"]
```
pgussow: Shouldn't the connection string be externalized? Because once MariaDB is in place this will change again.

Bert-R: For MariaDB we'll anyway change this. Note that this image is derived from the HSQLDB image.

PR #56, opened 2015-12-04
YD-52 resubscribe

aristotelos:


appservice/src/intTest/groovy/nu/yona/server/OverwriteUserTest.groovy
```diff
@@ -0,0 +1,394 @@
+package nu.yona.server
+
+import groovy.json.*
+import spock.lang.Shared
+import spock.lang.Specification
+
+class OverwriteUserTest extends Specification {
```
pgussow: This one should extend the new AbstractYonaIntegrationTest class

aristotelos:
There are many whitespace changes here, I do not know why, they seem to be introduced with merging the master into the branch.

PR #59, opened 2015-12-04
Yd 126 add action link

pgussow:


appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -248,6 +246,13 @@ static Link getUserSelfLinkWithTempPassword(UUID userID, String tempPassword)
 		return linkBuilder.withSelfRel();
 	}
 
+	private static Link getConfirmMobileLink(UUID userID)
+	{
+		ControllerLinkBuilder linkBuilder = linkTo(
+				methodOn(UserController.class).confirmMobileNumber(Optional.empty(), userID, null));
+		return linkBuilder.withSelfRel();
```
aristotelos: This is not correct. The link is inserted in the user json object so "self" must refer to the user href. Suggestion: return a LinkBuilder here and use the method `.withRel("confirmMobileNumber")` in the UserResourceAssembler.

PR #61, opened 2015-12-05
YD-74 Changes in DOS protection properties.

Bert-R:
* Renamed dosMemorySeconds into dosProtectionWindowSeconds
* Renamed isDOSProtectionEnabled into dosProtectionEnabled (in alignment with the other 'enabled' properties)
* Added maxCreateUserAttemptsPerTimeWindow (that was a fixed value at first).


core/src/main/java/nu/yona/server/properties/SecurityProperties.java
```diff
@@ -5,8 +5,9 @@
 	private int newDeviceRequestExpirationDays = 1;
 	private int passwordLength = 32;
 	private long bruteForceBlockMinutes = 60;
-	private long dosMemorySeconds = 300;
-	private boolean isDOSProtectionEnabled = false;
+	private long dosProtectionWindowSeconds = 300;
+	private boolean isDosProtectionEnabled = false;
+	private int maxCreateUserAttemptsPerTimeWindow = 1;
```
aristotelos: You updated the number to 2 in application.properties. I think 2 is OK, addUser may indeed be called twice in the resubscribe scenario (the first time it will give an exception and the app discovers that the user exists).

Bert-R: I did that intentionally. The other scenario would be running into a failure and retrying. Twice in 5 minutes sounds reasonable for that. Agreed?

aristotelos: Yes, but should you not also update the line in SecurityProperties.java from 1 to 2?

Bert-R: Not necessarily. The two defaults don't need to be identical. I'd rather make the one on the code very very strict, so you know that you're not working with realistic data.

aristotelos: OK, no problem.

PR #63, opened 2015-12-11
YD-85 Caching UserAnonymized

aristotelos:
Added UserAnonymizedDTO
Added MessageDestinationDTO
Reversed relation between Message and MessageDestination
Sending messages via MessageService
Getting UserAnonymizedDTO from UserService
Updating UserAnonymized from UserService
Removed unused destinationID in BuddyConnectResponseMessage


aristotelos:
Wait a moment. I forgot something: the `getBuddyDestinations` is not recomputed when a buddy accepts. Let me fix that.

analysisservice/src/main/java/nu/yona/server/AnalysisServiceApplication.java
```diff
@@ -51,6 +51,6 @@ public Docket yonaApi()
 	@Bean
 	public CacheManager cacheManager()
 	{
-		return new ConcurrentMapCacheManager("goalConflictMessages");
+		return new ConcurrentMapCacheManager("goalConflictMessages", "usersAnonymized");
```
Bert-R: To prevent the duplication of information, I propose to define the cache manager in ``CoreConfiguration``.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineCacheService.java
```diff
@@ -30,11 +31,12 @@ public GoalConflictMessage fetchLatestGoalConflictMessageForUser(UUID userAnonym
 	}
 
 	@CachePut(value = "goalConflictMessages", key = "{#message.relatedUserAnonymizedID,#message.goalID,#destination.ID}")
-	public GoalConflictMessage updateLatestGoalConflictMessageForUser(GoalConflictMessage message, MessageDestination destination)
+	public GoalConflictMessage updateLatestGoalConflictMessageForUser(GoalConflictMessage message,
+			MessageDestinationDTO destination)
```
Bert-R: This way, you don't need the ``destination`` anymore, right?

aristotelos: No you do. See the `@CachePut` directive which uses `#destination.ID`.

analysisservice/src/main/java/nu/yona/server/AnalysisServiceApplication.java
aristotelos: Accept. Will change this later. However for the app service you do not strictly need the `"goalConflictMessages"` cache.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineCacheService.java
Bert-R: I see. I only looked inside the method body.

analysisservice/src/main/java/nu/yona/server/AnalysisServiceApplication.java
Bert-R: Agreed. However, I expect the overhead to be negligible as long as we don't populate the cache.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -107,7 +110,7 @@ private void updateLastGoalConflictMessage(PotentialConflictDTO payload, Date me
 		message.setEndTime(messageEndTime);
 	}
 
-	private Set<Goal> determineConflictingGoalsForUser(UserAnonymized userAnonymized, Set<String> categories)
+	private Set<Goal> determineConflictingGoalsForUser(UserAnonymizedDTO userAnonymized, Set<String> categories)
 	{
 		Set<Goal> allGoals = goalService.getAllGoalEntities();
```
Bert-R: As a next step, we should change this too. The goal service should return goal DTOs and cache them.

core/src/main/java/nu/yona/server/messaging/entities/MessageDestination.java
```diff
@@ -5,21 +5,16 @@
 package nu.yona.server.messaging.entities;
```
Bert-R: I'm not happy with the changes in this file. This breaks down the domain model for message destinations and reduces the value of the ``MessageDestination`` entity to nearly nothing.

core/src/main/java/nu/yona/server/messaging/service/MessageService.java
```diff
@@ -18,9 +20,9 @@
 import org.springframework.stereotype.Component;
 import org.springframework.stereotype.Service;
 
+import nu.yona.server.crypto.PublicKeyEncryptor;
```
Bert-R: This knowledge should not go here but remain hidden inside the message destination.

core/src/main/java/nu/yona/server/messaging/entities/Message.java
```diff
@@ -20,7 +21,11 @@
 @Table(name = "MESSAGES")
 public abstract class Message extends EntityWithID
 {
+	@Column(nullable = false)
+	private UUID destinationID;
```
Bert-R: If you make the relationship between ``Message`` and ``MessageDestination`` two way navigable, you can still easily find messages for a destination, without having to abandon the relationship in the domain model.

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectResponseMessage.java
```diff
@@ -58,24 +46,22 @@ public void setProcessed()
...
-			UUID destinationID, String nickname, String message, UUID buddyID, BuddyAnonymized.Status status)
+			String nickname, String message, UUID buddyID, BuddyAnonymized.Status status)
 	{
 		return new BuddyConnectResponseMessage(UUID.randomUUID(), respondingUserID, respondingUserAnonymizedID, nickname, message,
-				buddyID, destinationID, status);
+				buddyID, status);
 	}
 
 	@Override
 	public void encrypt(Encryptor encryptor)
```
Bert-R: This method can now be removed, as there is nothing encrypted in this class anymore.

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectResponseMessage.java
```diff
@@ -15,11 +14,6 @@
 @Entity
 public class BuddyConnectResponseMessage extends BuddyConnectMessage
 {
-
-	@Transient
-	private UUID destinationID;
```
Bert-R: The change of moving the backward relationship up from ``BuddyConnectResponseMessage`` to ``Message`` is good.

core/src/main/java/nu/yona/server/subscriptions/entities/UserAnonymized.java
```diff
@@ -66,8 +67,13 @@ public MessageDestination getAnonymousDestination()
 
 	public Set<MessageDestination> getBuddyDestinations()
 	{
-		return buddiesAnonymized.stream().filter(ba -> ba.getSendingStatus() == Status.ACCEPTED).map(ba -> ba.getUserAnonymized())
-				.map(ua -> ua.getAnonymousDestination()).collect(Collectors.toSet());
+		return mapAvailableDestinations(buddiesAnonymized.stream()).collect(Collectors.toSet());
+	}
+
+	public static Stream<MessageDestination> mapAvailableDestinations(Stream<BuddyAnonymized> buddiesAnonymized2)
```
Bert-R: A couple of comments on this method signature:
* The name does not tell what the method does.
* The parameter name with a "2" in it is not very logical and it is an indicator of something else: ``UserAnonymized`` already knows the ``buddiesAnonymized``, so passing that into it is strange.
* Passing streams through public APIs is uncommon. We commonly use collections in such a case.

My proposal is to solve this in a different way: link anonymized buddy DTOs to the anonymized user DTOs, so you don't need to go to the database to find out where to send the messages to.

Bert-R:
I've provided some review feedback but need to stop now and don't have more time today. Hope this is helpful to make the next step.

core/src/main/java/nu/yona/server/messaging/entities/MessageDestination.java
aristotelos: This is because I am caching `MessageDestinationDTO` instances. These should not contain the set of messages. The `MessageService` should be able to use the message destination using a `MessageDestinationDTO`.

I didn't like the `MessageDestination` structure at all. It required `MessageDestination.getRepository().save(messageDestination)` every time when sending a message.

core/src/main/java/nu/yona/server/messaging/service/MessageService.java
aristotelos: I don't think that is useful. Entities should not contain too much logic, logic should take place in the services.

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectResponseMessage.java
aristotelos: OK, will do that.

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectResponseMessage.java
aristotelos: I thought that these columns accidentally had the same name. This column was used nowhere!

core/src/main/java/nu/yona/server/subscriptions/entities/UserAnonymized.java
aristotelos: The proposed solution is hard because these anonymized buddy DTOs need to expire when the buddy status changes (the `BuddyService` needs to expire them). Maybe the solution is to add a cache for anonymized buddy DTOs maintained by the `BuddyService` to accomplish this.

Bert-R: Indeed, that's the way to do it.

core/src/main/java/nu/yona/server/messaging/entities/MessageDestination.java
Bert-R: A couple of questions :
* why should these not contain the messages? If you don't include these, you would need to go to the database every time you need them.
* the message destination is the container of the messages. With your current design, you're implementing a relational  model and need to take care of deleting every message individually.
* Why do you not like that save on the destination repository?  Afraid of the overhead?  JPA keeps track of the changes and won't safe things that aren't changed

core/src/main/java/nu/yona/server/messaging/service/MessageService.java
Bert-R: By concept, the ``MessageSource`` / ``MessageDestination`` pair is a store of encrypted messages.  That responsibility can well be implemented in these two entities. The logic is very limited..

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectResponseMessage.java
Bert-R: I don't know where these come from.

core/src/main/java/nu/yona/server/messaging/entities/MessageDestination.java
aristotelos: - I think caching ALL messages is too much overhead.
- No you can delete using a query on the destination column.
- The extra overhead is: pulling the `MessageDestination` entity from the database by ID, and pushing the `MessageDestination` entity again.

analysisservice/src/main/java/nu/yona/server/AnalysisServiceApplication.java
aristotelos: Can I achieve this by just moving the method to `CoreConfiguration` and using `@Autowired
	CacheManager cacheManager;` here combined with the `@EnableCaching` directive?

aristotelos: Done

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectResponseMessage.java
aristotelos: Done

core/src/main/java/nu/yona/server/messaging/entities/MessageDestination.java
Bert-R: * We can do some math on the number of messages, based on [the tables created by Rine](http://wiki.yona.nu/display/DEV/Cluster+architecture). The other alternative that I still need to look in: would it be possible to use a distributed in memory database? That way, we wouldn't need to bother about caching at all.
* So it's not individually, but it's still relational. That's not the design we are applying.
* The ``MessageDestination`` is not being pushed, as that's not updated, but only the ``List`` inside it.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
aristotelos: Created [http://jira.yona.nu/browse/YD-143]

core/src/main/java/nu/yona/server/messaging/entities/MessageDestination.java
aristotelos: I think the pull of the `MessageDestination` is already too much overhead for sending a message. If you don't agree I can revert the change and implement a two-way relation. I'm not going to combine caching of all messages in this change set because it is altogether uncertain if we are going to do that.

Bert-R: I understand caching all messages is too much of a leap. Please revert ``MessageDestination`` to the original design. We might need a discussion in front of a whiteboard to conclude on what change is needed here.

aristotelos: OK, already started on that, will revert my changes.

yonabuild:

Bert-R:
ok to test

Groetjes, Bert

core/src/main/java/nu/yona/server/subscriptions/entities/UserAnonymized.java
aristotelos: Inlined the method. Please create an issue for caching buddy anonymized.

adminservice/src/main/java/nu/yona/server/AdminServiceApplication.java
```diff
@@ -38,8 +40,11 @@
 @SpringBootApplication
 @EnableSwagger2
 @EnableHypermediaSupport(type = HypermediaType.HAL)
+@EnableCaching
 public class AdminServiceApplication
 {
+	@Autowired
+	CacheManager cacheManager;
```
Bert-R: Is this used? It's not private, so it's possible, but I guess it's not used.

analysisservice/src/main/java/nu/yona/server/AnalysisServiceApplication.java
```diff
@@ -31,6 +31,8 @@
 @EnableCaching
 public class AnalysisServiceApplication
 {
+	@Autowired
+	CacheManager cacheManager;
```
Bert-R: Same question here: Is this used? It's not private, so it's possible, but I guess it's not used.
 Add a line note

adminservice/src/main/java/nu/yona/server/AdminServiceApplication.java
Bert-R: Scrolling further down, I see you're defining the bean on ``CoreConfiguration``, but not instantiating it there. I propose to centralize the instantiation of the cache manager in the core component layer, so everyone can rely in its availability.

Bert-R: Looking even further :)
Why do we need to create a cache manager ourselves? Spring Boot does that automatically, right? See [this article](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-caching.html) and also [this one](https://spring.io/blog/2014/06/16/further-cache-improvements-in-spring-4-1).

core/src/main/java/nu/yona/server/subscriptions/service/UserAnonymizedCacheService.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.cache.annotation.CachePut;
+import org.springframework.cache.annotation.Cacheable;
+import org.springframework.stereotype.Service;
+
+import nu.yona.server.exceptions.InvalidDataException;
+import nu.yona.server.subscriptions.entities.UserAnonymized;
+
+@CacheConfig(cacheNames = "usersAnonymized")
+@Service
+class UserAnonymizedCacheService
```
Bert-R: Omit ``Cache`` from the name. For consumers, it's simply the service that deals with ``UserAnonymized``.

Bert-R:
Looks pretty good now. I've added a few comments.

core/src/main/java/nu/yona/server/subscriptions/service/UserAnonymizedCacheService.java
aristotelos: Done

adminservice/src/main/java/nu/yona/server/AdminServiceApplication.java
aristotelos: Removed the `@Autowired CacheManager cacheManager` from the 3 files, it was not used indeed. Removing it from `CoreConfiguration` doesn't work, it gives:

...
java.lang.IllegalStateException: No CacheResolver specified, and no bean of type CacheManager found. Register a CacheManager bean or remove the @EnableCaching annotation from your configuration.
2015-12-22 06:57:03.027 ERROR 5624 --- [           main] o.s.boot.SpringApplication               : Application startup failed
...

Bert-R: OK. Created [YD-147](http://jira.yona.nu/browse/YD-147) for this.

PR #64, opened 2015-12-12
YD-135 Replaced vpnloginid by useranonymizedid

aristotelos:
Only retained where it belongs in AnalysisEngineService,
LDAPUserService, VPNProfileDTO etc


core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
```diff
@@ -180,4 +180,9 @@ public void touch()
 	{
 		touchVersion++;
 	}
+
+	public UUID getUserAnonymizedID()
```
Bert-R: Shouldn't this be the replacement of ``getVPNLoginID``? In other words: shouldn't ``getVPNLoginID`` be removed?

core/src/main/resources/messages/messages.properties
```diff
@@ -47,7 +47,7 @@ error.buddy.user.cannot.be.null=User cannot be null
 error.buddy.requesting.user.cannot.be.null=Requesting user cannot be null
 error.buddy.action.unknown=Action ''{0}'' is unknown
 error.buddy.message.entity.cannot.be.null=Message entity cannot be null
-error.buddy.vpnloginid.cannot.be.null=VPN Login Id cannot be null 
+error.buddy.useranonymizedid.cannot.be.null=Anonymized user id cannot be null
```
Bert-R: ID should be in capitals.

Bert-R:
That has been quite some work. Good action!
I've added two minor comments.

core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
aristotelos: It's referenced by `User.getVPNLoginID` which is referenced by `VPNProfileDTO.createInstance` and standing side by side to `getVPNPassword`. I don't think it should be removed.

core/src/main/resources/messages/messages.properties
aristotelos: All the messages have id in lowercase. It's more consistent now ;). Do you want me to do a global replace?

aristotelos:
This action was quite easy, just global search and using `Alt+Shift+R` in Eclipse...

core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
Bert-R: I see. In that case, ``getVPNLoginID`` should call ``getUserAnonymizedID`` (which would be faster, as currently ``getVPNLoginID`` first loads the object and then fetches its ID.

core/src/main/resources/messages/messages.properties
Bert-R: Yes, please. Be careful to only do it in the text, not the message ID.

aristotelos: Done

core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
aristotelos: Done

PR #66, opened 2015-12-12
YD-78 Sending invite SMS

aristotelos:
Also added invite message to buddy invite email as it was missing

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
@@ -26,6 +26,7 @@ import nu.yona.server.email.EmailService;
 import nu.yona.server.exceptions.EmailException;
 import nu.yona.server.messaging.entities.MessageDestination;
 import nu.yona.server.properties.YonaProperties;
+import nu.yona.server.sms.SmsService;
 import nu.yona.server.subscriptions.entities.Buddy;
 import nu.yona.server.subscriptions.entities.BuddyAnonymized;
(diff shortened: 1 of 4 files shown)


core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -253,14 +257,22 @@ private void sendInvitationMessage(UserDTO requestingUser, User buddyUserEntity,
 			String bodyTemplateName = "buddy-invitation-body";
 			String requestingUserName = StringUtils
 					.join(new Object[] { requestingUser.getFirstName(), requestingUser.getLastName() }, " ");
+			String requestingUserMobileNumber = requestingUser.getMobileNumber();
```
Bert-R: What about making this number the "from number"? Plivo allows specifying the" from", but I'm not sure whether we can do that per request.

core/src/main/resources/templates/sms/buddy-invitation.vm
```diff
@@ -0,0 +1 @@
+You have been invited to Yona by ${requestingUserName}. Please check your email!
```
Bert-R: What about mentioning the e-mail address in the text?

Bert-R: I've created YD-139 for this.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
Bert-R: I've created YD-140 for this.

PR #65, opened 2015-12-12
YD-122 Refactored messages, added tests

aristotelos:
Fixed bug: BuddyConnectRequestMessage had status NOT_REQUESTED
Disclosing status on all messages (have to add some tests still for updating messages after action)

PR #67, opened 2015-12-14
YD-131 Refactor tests

Bert-R:


core/src/main/java/nu/yona/server/subscriptions/service/BuddyDisconnectMessageDTO.java
```diff
@@ -106,7 +106,15 @@ private MessageActionDTO handleAction_Process(UserDTO actingUser, BuddyDisconnec
 		{
 			buddyService.removeBuddyAfterBuddyRemovedConnection(actingUser.getID(), messageEntity.getUserID());
 
+			updateMessageStatusAsProcessed(messageEntity);
```
Bert-R: This fixes a bug: the message status didn't flip to "processed" after processing it.

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
```diff
@@ -1,80 +1,65 @@
...
+	@Shared
+	private String baseTimestamp = createBaseTimestamp()
 
-abstract class AbstractAppServiceIntegrationTest extends AbstractYonaIntegrationTest {
 	@Shared
-	def timestamp = YonaServer.getTimeStamp()
+	private int sequenceNumber = 0
+
 	@Shared
 	def userNumber = 0;
```
aristotelos: This can be removed as it is no longer used.

Bert-R: Done.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyDisconnectMessageDTO.java
aristotelos: Nice catch! By the way, in https://github.com/yonadev/yona-server/pull/65 I
noticed that the status was not exposed on the message DTO at all... That
means that the status of no message was checked after doing some action on
it.

An idea: can we not embed the updated message DTO inside the
MessageActionDTO response? That will be more efficient for the app to
update existing messages.

Bert-R: I'm not sure I fully get what you mean.
The HATEOAS pattern normally communicates the possible actions through the URLs, so when the message can be processed, the ``/process`` URL exists. Once it's processed, that URL is gone.

Embedding the resulting message in the action response DTO is a good idea, as it saves a round trip.

core/src/testUtils/groovy/nu/yona/server/test/Buddy.groovy
```diff
@@ -0,0 +1,25 @@
...
+	final String receivingStatus
+	final String sendingStatus
+	final User user;
+	final String url
+	Buddy(def json)
+	{
+		this.nickname = json.nickname
+		this.receivingStatus = json.receivingStatus
+		this.sendingStatus = json.sendingStatus
+		// TODO: Is it appropriate to return the user with all null data? I'd rather omit the user in that case.
```
aristotelos: It's better to create Tasks instead of TODO comments.
For this case we maybe should return a user with most `null` data and `firstName` equal to `"(removed)"?

analysisservice/src/intTest/groovy/nu/yona/server/RelevantCategoriesTest.groovy
```diff
@@ -1,11 +1,20 @@
...
 import groovy.json.*
-import nu.yona.server.test.AbstractYonaIntegrationTest
 
-class RelevantCategoriesTest extends AbstractYonaIntegrationTest {
+import spock.lang.Shared
+import spock.lang.Specification
 
-	def 'Get relevant categories'(){
+import nu.yona.server.test.AnalysisService
+import nu.yona.server.test.Service
```
aristotelos: This import is no longer needed.

adminservice/src/intTest/groovy/nu/yona/server/GoalTest.groovy
```diff
@@ -1,11 +1,21 @@
...
 
-	def 'Get all goals loaded from file'(){
+import nu.yona.server.YonaServer
+import nu.yona.server.test.Service
+
+class GoalTest extends Specification
+{
+	@Shared
+	def adminServiceBaseURL = Service.getProperty('yona.adminservice.url', "http://localhost:8080")
+	def YonaServer adminService = new YonaServer(adminServiceBaseURL)
```
aristotelos: Maybe nice to refactor this to an AdminService as with the other two services.

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
```diff
@@ -0,0 +1,406 @@
...
+		// Have the other user fetch the buddy connect request
+		def response = getDirectMessages(user)
+		assert response.status == 200
+		assert response.responseData._embedded
+
+		def message = response.responseData._embedded?.buddyConnectRequestMessages[0]?.message ?: null
+		def acceptURL = response.responseData._embedded?.buddyConnectRequestMessages[0]?._links?.accept?.href ?: null
+		def rejectURL = response.responseData._embedded?.buddyConnectRequestMessages[0]?._links?.reject?.href ?: null
+
+		def result = [ : ]
```
aristotelos: I would prefer to return the `responseData._embedded.buddyConnectRequestMessages[0]` here. It makes the tests contain a little less logic.

core/src/testUtils/groovy/nu/yona/server/test/User.groovy
```diff
@@ -0,0 +1,101 @@
...
+		this(json, hasPrivateData)
+		this.password = password
+	}
+	User(def json)
+	{
+		this(json, false)
+	}
+	private User(def json, boolean hasPrivateData)
+	{
+		this.firstName = json.firstName
```
aristotelos: I would prefer to just have a property `user` storing the JSON and a property `password`. It makes the tests contain a little less logic.

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
Bert-R: Possible, but then we would need to pass an asserted. I propose to do that after the pull.

adminservice/src/intTest/groovy/nu/yona/server/GoalTest.groovy
Bert-R: Let's do that next time we extend this test.

core/src/testUtils/groovy/nu/yona/server/test/Buddy.groovy
Bert-R: I knew you were reviewing this. Wanted to discuss before creating tasks.

analysisservice/src/intTest/groovy/nu/yona/server/RelevantCategoriesTest.groovy
Bert-R: Done

core/src/testUtils/groovy/nu/yona/server/test/User.groovy
Bert-R: I doubt that. The json implies having _embedded and_links. That's a lot more in tests.

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
```diff
@@ -0,0 +1,406 @@
...
+		assertUserWithPrivateData(response.responseData)
+	}
+
+	def assertUserOverwriteResponseDetails(def response)
+	{
+		assertResponseStatusCreated(response)
+		assertUserWithPrivateData(response.responseData)
+	}
+
+	def assertUserGetResponseDetailsWithPrivateDataCreatedOnBoddyRequest(def response)
```
aristotelos: In my opinion the methods related to asserts should be in the tests they belong to. So this one should be in `CreateUserOnBuddyRequestTest` and should only be used there. The `CreateUserOnBuddyRequestTest` should contain logic about what to assert and what not, and should apply it on one place if possible (you do not need to assert things over and over - if it fails in the right test fixture this is sufficient).

appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
```diff
@@ -1,236 +1,227 @@
...
+			updatedJohn.mobileNumber = newMobileNumber
+			def userUpdateResponse = appService.updateUser(john.url, updatedJohn, john.password);
 
 		then:
-		response.status == 200
-		verifyUserDoesNotExist(userURL)
+			userUpdateResponse.status == 200
+			userUpdateResponse.responseData._links?.confirmMobileNumber?.href != null
+			userUpdateResponse.responseData.mobileNumber == newMobileNumber
+
```
aristotelos: I am missing the rest of the test here. In the original the new mobile number was confirmed afterwards and checked.

appservice/src/intTest/groovy/nu/yona/server/RemoveUserTest.groovy
```diff
@@ -1,159 +1,131 @@
...
+			getAnonMessagesResponse.responseData._embedded.goalConflictMessages[0].nickname == "<self>"
+			getAnonMessagesResponse.responseData._embedded.goalConflictMessages[0].goalName == "gambling"
+			getAnonMessagesResponse.responseData._embedded.goalConflictMessages[0].url =~ /poker/
+
+		cleanup:
+			appService.deleteUser(bob)
 	}
 
-	def 'Bob checks his direct messages and the messages of Richard are no longer there'(){
+	def 'Conflicts for Bob are still processed after the unsubscribe of Bob'()
```
aristotelos: You mean unsubscribe of Richard.

appservice/src/intTest/groovy/nu/yona/server/RemoveUserTest.groovy
```diff
@@ -1,159 +1,131 @@
...
-		response.responseData._embedded.buddyDisconnectMessages[0].reason == "USER_ACCOUNT_DELETED"
-		response.responseData._embedded.buddyDisconnectMessages[0].message == "Goodbye friends! I deinstalled the Internet"
-		response.responseData._embedded.buddyDisconnectMessages[0].nickname == richardQuin.nickname
-		response.responseData._embedded.buddyDisconnectMessages[0]._links.self.href.startsWith(bobDunn.url + appService.ANONYMOUS_MESSAGES_PATH_FRAGMENT)
-		bobDunnBuddyRemoveMessageProcessURL.startsWith(response.responseData._embedded.buddyDisconnectMessages[0]._links.self.href)
+			response.status == 200
 	}
 
-	def 'Bob processes the remove buddy message'(){
+	def 'Delete and recreate account'()
```
aristotelos: Nice test :). Good that you added some other tests on the way.

appservice/src/intTest/groovy/nu/yona/server/RemoveUserTest.groovy
```diff
@@ -1,159 +1,131 @@
...
 
 		when:
-		def response = appService.getBuddies(bobDunn.url, bobDunn.password);
+			def message = "Goodbye friends! I deinstalled the Internet"
+			appService.deleteUser(richard, message)
 
 		then:
-		response.status == 200
-		response.responseData._embedded == null
+			// TODO: when are the buddies actually removed? Not on "process" of the message?
```
aristotelos: The buddy is removed on process of the message indeed. That is covered by the next tests.

appservice/src/intTest/groovy/nu/yona/server/RejectBuddyTest.groovy
```diff
@@ -1,154 +1,45 @@
...
+			def getRichardBudddiesResponse = appService.getBuddies(richard);
+			!getRichardBudddiesResponse.responseData?._embedded?.buddies
 
-		then:
-		response.status == 201
-		response.responseData._embedded.user.firstName == "Bob"
-		richardQuinBobBuddyURL.startsWith(richardQuin.url)
+			// Verify that Richard is not in Bob's buddy list anymore
+			def getBobBudddiesResponse = appService.getBuddies(bob);
+			!getBobBudddiesResponse.responseData?._embedded?.buddies
```
aristotelos: In another test I saw just `def buddies = appService.getBuddies(bob); buddies.size() == 0`. That was easier to read.

appservice/src/intTest/groovy/nu/yona/server/RejectBuddyTest.groovy
```diff
@@ -1,154 +1,45 @@
...
-				},
-				"message":"Would you like to be my buddy?",
-				"sendingStatus":"REQUESTED",
-				"receivingStatus":"REQUESTED"
-			}""", richardQuin.password)
-		if (response.status == 201) {
-			richardQuinBobBuddyURL = response.responseData._links.self.href
-		}
+			// Verify that Bob is not in Richard's buddy list anymore
+			def getRichardBudddiesResponse = appService.getBuddies(richard);
```
aristotelos: Spelling error Budddies.

appservice/src/intTest/groovy/nu/yona/server/OverwriteUserTest.groovy
```diff
@@ -1,220 +1,184 @@
...
+			appService.deleteUser(richard)
+			appService.deleteUser(bob)
 	}
 
-	def 'Delete users'(){
+	def 'Hacking attempt: Brute force overwrite mobile number confirmation'()
+	{
+		given:
+			def userCreationMobileNumber = "+${timestamp}99"
+			def userCreationJSON = """{
```
aristotelos: Maybe nicer to use a template here.

appservice/src/intTest/groovy/nu/yona/server/OverwriteUserTest.groovy
```diff
@@ -1,220 +1,184 @@
...
-	def 'Request overwrite of the existing user'(){
-		when:
-		def response = appService.requestOverwriteUser("${richardQuin.mobileNumber}")
-		if(response.status == 200) {
-			richardQuinOverwriteConfirmationCode = response.responseData.confirmationCode
-		}
-		then:
-		response.status == 200
-		richardQuinOverwriteConfirmationCode
+	def userExistsAsserter(def response)
```
aristotelos: Can't this method be inlined?

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -2,448 +2,421 @@ package nu.yona.server
...
-			response.responseData.properties.status == "done"
-	}
-
-	def 'Bob\'s user data will contain Richard as buddy'(){
-		given:
-
-		when:
-			def response = appService.getUser(bobDunnURL, true, bobDunnPassword)
+			response.status == 400
+			// response.responseData.code == "TODO" // Why do we not get a response code here?
```
aristotelos: Maybe because it's an `InvalidArgumentException`.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -2,448 +2,421 @@ package nu.yona.server
...
+			bobWithBuddy.buddies[0].sendingStatus == "ACCEPTED"
+			bobWithBuddy.buddies[0].receivingStatus == "ACCEPTED"
 
+		cleanup:
+			appService.deleteUser(richard)
+			appService.deleteUser(updatedBob)
 	}
 
-	def 'Bob Dunn adjusts data and submits; app saves with new password'(){
+	def 'Richard processes Bob\'s buddy acceptance'()
```
aristotelos: I think this could be merged with the previous test.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -2,448 +2,421 @@ package nu.yona.server
...
+			richardWithBuddy.buddies[0].sendingStatus == "ACCEPTED"
+			richardWithBuddy.buddies[0].receivingStatus == "ACCEPTED"
+
+		cleanup:
+			appService.deleteUser(richard)
+			appService.deleteUser(updatedBob)
 	}
 
-	def 'Bob cannot read direct messages before confirming mobile number'(){
+	def 'Goal conflict of Richard is reported to Richard and Bob'()
```
aristotelos: This could also maybe be merged with the previous test.

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
```diff
@@ -2,108 +2,113 @@ package nu.yona.server
...
 
-class AddDeviceTest extends AbstractAppServiceIntegrationTest {
-
-	@Shared
-	def richardQuin
-
-	def 'Add user Richard'(){
+class AddDeviceTest extends AbstractAppServiceIntegrationTest
+{
+	def 'Set new device request'()
```
aristotelos: Rename to 'Set and get new device request'

appservice/src/intTest/groovy/nu/yona/server/BasicBuddyTest.groovy
```diff
@@ -1,536 +1,378 @@
...
-		response.responseData._embedded.buddyDisconnectMessages[0].nickname == "${richardQuin.nickname}"
-		response.responseData._embedded.buddyDisconnectMessages[0].message == "Bob, as you know our ways parted so I'll remove you as a buddy."
-		response.responseData._embedded.buddyDisconnectMessages[0]._links.self.href.startsWith(bobDunn.url + appService.ANONYMOUS_MESSAGES_PATH_FRAGMENT)
-		bobDunnBuddyRemoveMessageProcessURL.startsWith(response.responseData._embedded.buddyDisconnectMessages[0]._links.self.href)
-	}
+			response.status == 200
+			appService.getBuddies(richard).size() == 0 // Buddy removed for Richard`
+			appService.getBuddies(bob).size() == 1 // Buddy not yet removed for Bob (not processed yet)
 
-	def 'Test what happens if the classification engine detects a potential conflict for Bob (third conflict message) when the buddy disconnect is not yet processed'(){
```
aristotelos: Is this test still there? I can't find it inlined in any test.

appservice/src/intTest/groovy/nu/yona/server/DisclosureTest.groovy
```diff
@@ -1,182 +1,144 @@
...
-		goalDiscloseRequestMessage1AcceptURL
-		goalDiscloseRequestMessage2RejectURL
+			response.status == 200
+			def getRichardAnonMessagesResponse = appService.getAnonymousMessages(richard)
+			getRichardAnonMessagesResponse.status == 200
+			getRichardAnonMessagesResponse.responseData?._embedded?.discloseRequestMessages
+			def discloseRequestMessages = getRichardAnonMessagesResponse.responseData._embedded.discloseRequestMessages
+			discloseRequestMessages.size() == 1
+			discloseRequestMessages[0].targetGoalConflictMessage.goalName == "gambling"
+			discloseRequestMessages[0].targetGoalConflictMessage.creationTime > (System.currentTimeMillis() - 50000) // TODO Use standard date/time format
```
aristotelos: This IS the standard HATEOAS date format.

core/src/testUtils/groovy/nu/yona/server/test/User.groovy
aristotelos: Accepted. For User it indeed makes sense to optimize a little because it is so often used.

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
aristotelos: I'm not sure what you mean by "we would need to pass an asserted".

adminservice/src/intTest/groovy/nu/yona/server/GoalTest.groovy
aristotelos: Hesitatingly accepted.

aristotelos:
I'm ready reviewing. Overall the tests are good to read and have more asserts than before the refactoring. I also saw some extra added tests on the way. Sometimes there is more repeated code to do the test setup, but that is not blocking. One thing I don't like is the assert logic in AppService, in my opinion it belongs to the tests, but that has been mentioned in a comment. But the pull request seems good enough to merge after processing the comments anyway.

core/src/testUtils/groovy/nu/yona/server/test/Buddy.groovy
Bert-R: Done:  YD-136 - Make the user null when the buddy is removed

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
Bert-R: Nicely "corrected" by Android. I meant to write "asserter". However, looking closer at what you said, with more context: That means you don't have URLs but links. That makes the consumption a lot more cumbersome (e.g. ``message._links?.reject?.href`` instead of ``message.rejectURL``). That's why I'm mapping it down to a simple structure.

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
Bert-R: This is what I generally did, but not if it lead to code duplication. The difference between ``assertUserCreationResponseDetails`` and ``assertUserOverwriteResponseDetails`` is that the former in addition verifies that there is a confirmation code.

appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
Bert-R: This is omitted on purpose. Mobile number confirmation is already tested. If in this scenario the new number is set and the confirmation link is present, there is no reason to assume confirmation will be different, so no need to add a test for that.
Makes sense?

appservice/src/intTest/groovy/nu/yona/server/RemoveUserTest.groovy
Bert-R: Done

appservice/src/intTest/groovy/nu/yona/server/RemoveUserTest.groovy
Bert-R: Removed the comment

appservice/src/intTest/groovy/nu/yona/server/RejectBuddyTest.groovy
Bert-R: Nice catch!
This is simply wrong. The entire chain results in null, as the response doesn't contain ``responseData``, wich makes the test positive, but that would even occur when the buddy is still there. Corrected it.

appservice/src/intTest/groovy/nu/yona/server/RejectBuddyTest.groovy
Bert-R: Corrected it with the change of the previous comment.

appservice/src/intTest/groovy/nu/yona/server/OverwriteUserTest.groovy
Bert-R: I'm not sure what you mean with this. This JSON fragment is only used here. Do you propose to move it out of this method?

appservice/src/intTest/groovy/nu/yona/server/OverwriteUserTest.groovy
Bert-R: The test "Attempt to add another user with the same mobile number" passes it as asserter to the ``addUser`` call.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
Bert-R: That's not the case (searched for it). I've logged YD-137 for this and mentioned it in the TODOs

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
Bert-R: That would imply if-statements in the previous test: if the first step is successful, do the second one and verify the results. The good thing of the current test structure is that there are nearly no conditions in the tests. That increases the readability a lot.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
Bert-R: Same here. The alternative would be to do these actions in the ``then`` block. Till now, I didn't do that but only put verficiation/read actions in the ``then`` blocks.

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
Bert-R: The test is to verify whether "set" works (and that is in the ``when``). To verify that, "get" is done (as part of ``then``), but that's not what the test is made for, so I prefer to stick with the current name.

appservice/src/intTest/groovy/nu/yona/server/BasicBuddyTest.groovy
Bert-R: The test ``After Richard removed Bob as buddy, new goal conflicts are not reported to the buddies anymore`` does that.

appservice/src/intTest/groovy/nu/yona/server/DisclosureTest.groovy
Bert-R: Where do you take that from? Based on http://stackoverflow.com/a/15952652, I would think we can use a more readable format.

Bert-R:
For the comments where I said "done" or something like that, I just pushed a change set. The second commit of that change set adds a copyright header above each file and uses the formatting that you're used to. I tried to get the formatting in a (IMHO) more clear way, with an extra indent for each block, but that cannot be done automatically (even not with the Spock plug-in for Eclipse). Based on your earlier mail, I've opted for the automation.

To your point on the assertions: generally, the assertions are in the tests, but for reusable methods used from multiple tests, I've used closures in the methods in ``AppService`` and predefined (and used) several of them in that same class. Otherwise, we would need to duplicate code, and that's not a good practice (either).

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
aristotelos: OK, but I don't like to have to create a method like this for every Message type, and have to look inside this method when the link is renamed or something like that, and have to update this method when adding a new link type.

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
aristotelos: Then you should rename it to `assertValidVPNProfile` and have a method for the user created on buddy request in the `CreateUserOnBuddyRequestTest`.

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
aristotelos: OK, then add a separate test for get since you removed it.

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
Bert-R: Why not? It's better to do that here centrally then having to clutter all
tests for it. If we do other messages with this pattern, we might want to
introduce a class, like buddy and user.

This particular method will never support other message types.

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
Bert-R: Besides what's tested in the ``then`` here, what else should that test do?

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
Bert-R: I don't understand that. How would that reduce the code or make it better maintainable?

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
aristotelos: That sound like duplicating all your logic in tests right...

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
aristotelos: It's not very important but I would prefer to not distribute logic about the 'create user on buddy request' in two files.

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
aristotelos: You removed the test `def 'Get new device request with user secret'()`. I would expect a new test or a test that has the name `Set new device request and get it with user secret`.

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
Bert-R: Though I agree to that point, I don't know how we can prevent it. There are two options: have it in this class and use it to assert correct creation and overwrite, or have it in two classes: here to assert correct creation and in "create user on buddy request" to assert correct overwrite. I felt duplication is better than moving it out of context, so I centralized it here.

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
Bert-R: That's a matter of taste, I think. The "then" blocks at times contain get operations to validate the "when". This is not reflected in the test name. Otherwise, the test names would become very long. The "then" block however tests everything that was inside the original test.

Your comment might also be particular to the change that I'm making here, as we go from an old test set (with a separate test for "get") to a new one (that uses "get" in the "then" block).

aristotelos: Setting the device request (from your old device) and getting it (from your new device) is functionally different! So I would like to have two tests for that or a test method that reflects that they are both tested.

PR #68, opened 2015-12-16
YD-141 move test specific assertion

Bert-R:
Is this what you meant? That's quite a bit less code than comment. Sorry for not calling you earlier.

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/BasicBuddyTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/BasicBuddyTest.groovy
@@ -99,7 +99,7 @@ class BasicBuddyTest extends AbstractAppServiceIntegrationTest
 		buddies[0].sendingStatus == "ACCEPTED"
 		buddies[0].receivingStatus == "ACCEPTED"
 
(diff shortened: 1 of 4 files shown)


aristotelos:
This is what I meant. You can move `assertUserGetResponseDetailsWithPrivateDataCreatedOnBoddyRequest`, `assertUserWithPrivateDataCreatedOnBuddyRequest` to the `CreateUserOnBuddyRequestTest` class too.

Bert-R:
These are used in by two tests: ``CreateUserOnBuddyRequestTest`` and ``BasicBuddyTest``

Bert-R:
I did one more rename (to make it a test-case independent assertion): ``assertPrivateUserDataCreatedOnBuddyRequest`` into ``assertVpnProfile``. The other two cannot be moved into a test, as they are used in two tests.

Bert-R:
I've renamed them to make the names test case independent, just expressing what they assert. Is this better?

aristotelos:
Yes, a lot better! The "created on buddy request" logic is now only in the `CreateUserOnBuddyRequestTest`. and the "overwrite user" logic is now only in the `OverwriteUserTest`.

PR #72, opened 2015-12-19
YD-144 Added logging of key events

Bert-R:


Bert-R:
There are a few more important lifetime events, but before I do these, I want to handle (YD-138)[http://jira.yona.nu/browse/YD-138] and (YD-146)[http://jira.yona.nu/browse/YD-146].

aristotelos:
No comments, looks good.

PR #73, opened 2015-12-22
YD-138 Improve resource based exception

Bert-R:


core/src/main/java/nu/yona/server/exceptions/ResourceBasedException.java
```diff
@@ -1,21 +1,44 @@
...
  */
 public abstract class ResourceBasedException extends RuntimeException
 {
+	@Component
+	private static class TranslatedTextRetriever
+	{
+		@Autowired
+		private Translator translator;
+
+		@EventListener({ ContextRefreshedEvent.class })
```
aristotelos: Is it not possible to inline this method in the `Translator` class? In that case we do not need a `TranslatedTextRetriever` class.

core/src/main/java/nu/yona/server/exceptions/ResourceBasedException.java
```diff
@@ -73,6 +101,23 @@ protected ResourceBasedException(HttpStatus statusCode, Throwable t, String mess
...
 	}
 
+	@Override
+	public String getMessage()
+	{
+		return getLocalizedMessage();
+	}
+
+	@Override
+	public String getLocalizedMessage()
```
aristotelos: Please update the `GlobalExceptionMapping` class to use this method instead of directly calling `Translator`.

core/src/main/java/nu/yona/server/exceptions/ResourceBasedException.java
aristotelos: Hmm, maybe the current structure is better because then `Translator` does not need to know about `ResourceBasedException`.
I have a name suggestion though: `TranslatorInjector`, it could set a static `Translator` object.

core/src/main/java/nu/yona/server/exceptions/ResourceBasedException.java
Bert-R: Done.

core/src/main/java/nu/yona/server/exceptions/ResourceBasedException.java
Bert-R: Indeed, that was the reason to use a microcomponent. It's not possible to autowire a static field. This object is used to retrieve the text, not to inject the translator, so I feel the current name is more appropriate.

aristotelos: But the problem that this class solves is that the translator cannot be injected by autowiring. So you could choose to set a `private static Translator translator;` (instead of `private static TranslatedTextRetriever translatedTextRetriever;`) in `ResourceBasedException` and rename `TranslatedTextRetriever` to `TranslatorInjector` and remove the `getLocalizedText` method from it. It would be less code and less 'magic' to me.

Bert-R: I see. That indeed simplifies it. I'll change it that way.

PR #74, opened 2015-12-22
YD-146 switch from jul to slf4j

Bert-R:


aristotelos:
No comments, looks good

PR #77, opened 2015-12-23
YD-134 Properly handle decryption failures

Bert-R:
With this, the modified hacking test (see [the JIRA issue](http://jira.yona.nu/browse/YD-134)) succeeds even when run 1000 times.


core/src/main/java/nu/yona/server/crypto/LongFieldEncrypter.java
```diff
@@ -27,6 +27,19 @@ public Long convertToEntityAttribute(String dbData)
...
 		String decryptedString = stringFieldEncrypter.convertToEntityAttribute(dbData);
-		return (decryptedString == null) ? 0 : Long.parseLong(decryptedString);
+		return (decryptedString == null) ? 0 : tryToConvertToLong(decryptedString);
+	}
+
+	private long tryToConvertToLong(String decryptedString)
+	{
+		try
+		{
+
```
aristotelos: Empty line

Bert-R: Gone.

PR #79, opened 2015-12-28
YD-149 Goals design change

aristotelos:
- Renamed Goal to ActivityCategory
- Added Goal class which is a placeholder (currently all goals are still considered no-go goals)
- Added Activity class, caching and updating endTime on Activity instead of on the GoalConflictMessage (no example url aggregation yet)
- IntervalActivity not yet introduced


aristotelos:
Started first part of the refactoring as this seems to me highest prio on the backlog now. Kept this changeset small to keep it reviewable. Moreover, the app design is not completely ready yet so it is risky to apply the whole refactoring already. On the design page [http://wiki.yona.nu/display/DEV/Design+change+to+capture+app+activity] I posted some more questions that arose while implementing.

core/src/main/java/nu/yona/server/analysis/entities/Activity.java
```diff
@@ -0,0 +1,74 @@
...
+@Table(name = "ACTIVITIES")
+public class Activity extends EntityWithID
+{
+	public static ActivityRepository getRepository()
+	{
+		return (ActivityRepository) RepositoryProvider.getRepository(Activity.class, UUID.class);
+	}
+
+	private UUID userAnonymizedID;
+	private UUID goalID;
```
Bert-R: Why are these two IDs instead of regular associations?

core/src/main/java/nu/yona/server/analysis/entities/GoalConflictMessage.java
```diff
@@ -125,22 +119,17 @@ public static GoalConflictMessage createInstanceFromBuddy(UUID relatedUserAnonym
 	{
 		if (origin == null)
 		{
-			throw GoalServiceException.originCannotBeNull();
+			throw new IllegalArgumentException("origin cannot be null");
```
Bert-R: Let's stick to the pattern that Phillip introduced and use specific and localizable exceptions everywhere.

core/src/main/java/nu/yona/server/analysis/entities/Activity.java
aristotelos: In `GoalConflictMessage` these were IDs too. It would be OK to change them, but then we would have a problem with creating instances because `Goal` and `UserAnonymized` instances are not cached, but `GoalDTO` and `UserAnonymizedDTO` instances are. So if we change this from IDs to associations we need to retrieve the instances from the repository directly, isn't it?

core/src/main/java/nu/yona/server/analysis/entities/GoalConflictMessage.java
aristotelos: There are still 12 instances of `IllegalArgumentException` in the source code. I propose not to change all of those with this design change because it clutters up the Pull Request. It's OK to remove them, so please add a separate task.

core/src/main/java/nu/yona/server/analysis/entities/Activity.java
Bert-R: This change is already enormous, so let's not combine that change. I've logged [YD-150](http://jira.yona.nu/browse/YD-150) for this.

core/src/main/java/nu/yona/server/analysis/entities/GoalConflictMessage.java
Bert-R: I see. I agree to not growing the change set any further. Logged [YD-151](http://jira.yona.nu/browse/YD-151).

core/src/main/java/nu/yona/server/rest/GlobalExceptionMapping.java
```diff
@@ -23,10 +21,6 @@
 {
 	private static final Logger logger = LoggerFactory.getLogger(GlobalExceptionMapping.class);
 
-	/** The source that contains the actual messages for the codes */
-	@Autowired
-	private Translator translator;
-
```
Bert-R: Thanks!

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectRequestMessage.java
```diff
@@ -28,8 +25,8 @@ public BuddyConnectRequestMessage()
 		super();
 	}
 
-	private BuddyConnectRequestMessage(UUID id, UUID userID, UUID userAnonymizedID, Set<UUID> goalIDs, String nickname,
-			String message, UUID buddyID, boolean isRequestingSending, boolean isRequestingReceiving)
+	private BuddyConnectRequestMessage(UUID id, UUID userID, UUID userAnonymizedID, String nickname, String message, UUID buddyID,
+			boolean isRequestingSending, boolean isRequestingReceiving)
```
Bert-R: Good cleanup. In the past, the buddy connect request communicated the set of goals for the user. Nowadays these are simply part of the user anonymized.

Bert-R:
That was a major effort (more than just search/replace)!
I'm going to pull it.

PR #80, opened 2015-12-29
YD-142 embed message after action

aristotelos:


core/src/main/java/nu/yona/server/messaging/service/MessageService.java
```diff
@@ -209,4 +209,14 @@ public void removeMessagesFromUser(MessageDestinationDTO destination, UUID sentB
 		destinationEntity.removeMessagesFromUser(sentByUserAnonymizedID);
 		MessageDestination.getRepository().save(destinationEntity);
 	}
+
+	public static boolean getIsDirectMessage(MessageDTO message)
```
Bert-R: Please follow the common naming pattern: ``isDirectMessage(...)``.

core/src/main/java/nu/yona/server/messaging/service/MessageService.java
```diff
@@ -209,4 +209,14 @@ public void removeMessagesFromUser(MessageDestinationDTO destination, UUID sentB
...
+	{
+		if (message instanceof BuddyConnectRequestMessageDTO)
```
Bert-R: This approach is a little tricky. Over time, we might add more types of direct messages or, even worse, send some type of message through both message boxes. Let's for now stick with it, as it makes the overall code flow mor readable.

core/src/main/java/nu/yona/server/messaging/service/DiscloseRequestMessageDTO.java
```diff
@@ -122,33 +121,39 @@ public MessageActionDTO handleAction(UserDTO actingUser, Message messageEntity,
...
+		private MessageActionDTO handleAction_Reject(UserDTO actingUser, DiscloseRequestMessage discloseRequestMessageEntity,
 				MessageActionDTO payload)
 		{
-			return updateResult(rejectingUser, discloseRequestMessageEntity, GoalConflictMessage.Status.DISCLOSE_REJECTED,
+			return updateResult(actingUser, discloseRequestMessageEntity, GoalConflictMessage.Status.DISCLOSE_REJECTED,
 					payload.getProperty("message"));
 		}
 
-		private MessageActionDTO updateResult(UserDTO respondingUser, DiscloseRequestMessage discloseRequestMessageEntity,
+		private MessageActionDTO updateResult(UserDTO actingUser, DiscloseRequestMessage discloseRequestMessageEntity,
```
Bert-R: I know the method name is not changed, but still: This method name is not very logical to me. What about ``buildReturnValue``?

Bert-R:
Looks good. I added a few minor comments.

core/src/main/java/nu/yona/server/messaging/service/DiscloseRequestMessageDTO.java
aristotelos: Renamed to more speaking `updateGoalConflictMessageStatus`

core/src/main/java/nu/yona/server/messaging/service/MessageService.java
aristotelos: Indeed, done

aristotelos:
Reworked

Bert-R:
But not committed or not pushed...

aristotelos:
Oops, indeed not pushed correctly, done now

Bert-R:
Looks good. I'll merge it after the build of the previous one is completed (a second build while one is active already might break the test run, as that replaces the docker images while the tests are active).

PR #82, opened 2016-01-09
YD-147 autoconfigure cache

Bert-R:


aristotelos:
Looks good. I am not a cygwin expert so can't judge that part well :).

PR #88, opened 2016-01-16
YD-154 Introduced Hamcrest matchers

Bert-R:
This way of matching retains more of the semantics and shows the compared data in the output upon a mismatch.


aristotelos:
Nice, is indeed better

PR #89, opened 2016-01-19
YD-155 Added BudgetGoal subclass of Goal, fixed tests

aristotelos:


appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -347,7 +347,9 @@ class CreateUserOnBuddyRequestTest extends AbstractAppServiceIntegrationTest
 				],
 				"goals":[
 					{
-						"activityCategoryName":"news"
+						"@class":"budgetGoal",
```
aristotelos: Polymorphism has been implemented using a "@class" property. I am not sure how to do this better with HAL, I did not find good examples of it.

Bert-R: Let's stick with this for now. We anyway need to handle goal changes differently, see [YD-165](http://jira.yona.nu/browse/YD-165).

core/src/main/java/nu/yona/server/goals/entities/BudgetGoal.java
```diff
@@ -0,0 +1,50 @@
...
+
+@Entity
+public class BudgetGoal extends Goal
+{
+	private int maxDuration;
+
+	// Default constructor is required for JPA
+	public BudgetGoal()
+	{
+		super();
```
Bert-R: This line isn't necessary.

core/src/main/java/nu/yona/server/goals/service/GoalDTO.java
```diff
@@ -39,13 +37,15 @@ public String getActivityCategoryName()
...
 	}
 
-	public static GoalDTO createInstance(Goal entity)
-	{
-		return new GoalDTO(entity.getID(), entity.getActivityCategory().getName());
-	}
+	public abstract Goal createGoalEntity();
 
-	public Goal createGoalEntity()
+	public static GoalDTO createInstance(Goal g)
```
Bert-R: Just use the full name: ``goal``.

Bert-R:
I made two minor comments. For the rest it's fine.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
aristotelos: Agreed. It took me most time to implement the polymorphism like this but I realized it was lost time anyway :(. Maybe it is good to do YD-165 first or in combination with adding other goal types to take the impact as early as possible.

PR #90, opened 2016-01-20
YD-137 Extended and fixed some tests

aristotelos:
Exceptions were already added by @pgussow


Bert-R:
I refactored the tests after that, so I don't understand why I didn't get error codes.
I'm going to merge this one.

PR #92, opened 2016-01-21
YD-136 Set UserDTO to null when user removed

aristotelos:
Not returning a 'stub entity'

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/service/BuddyDisconnectMessageDTO.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/BuddyDisconnectMessageDTO.java
@@ -19,7 +19,6 @@ import nu.yona.server.messaging.service.MessageService.DTOManager;
 import nu.yona.server.messaging.service.MessageService.TheDTOManager;
 import nu.yona.server.messaging.service.MessageServiceException;
 import nu.yona.server.subscriptions.entities.BuddyDisconnectMessage;
-import nu.yona.server.subscriptions.entities.User;
 import nu.yona.server.subscriptions.service.BuddyService.DropBuddyReason;
 
 @JsonRootName("buddyDisconnectMessage")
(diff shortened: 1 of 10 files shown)


core/src/main/java/nu/yona/server/subscriptions/service/UserDTO.java
```diff
@@ -154,17 +154,12 @@ static UserDTO createInstance(User userEntity)
 	{
 		if (userEntity == null)
 		{
-			return createRemovedUserInstance();
+			return null;
```
Bert-R: Can't you fix this at the level of the caller? It's a weird contract to pass a null value to a ``createInstance`` method and then get null back.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyDisconnectMessageDTO.java
```diff
@@ -61,10 +61,9 @@ public boolean isProcessed()
 	public static BuddyDisconnectMessageDTO createInstance(UserDTO actingUser, BuddyDisconnectMessage messageEntity)
 	{
 		User userEntity = messageEntity.getUser(); // may be null if deleted
-		UserDTO user = userEntity != null ? UserDTO.createInstance(userEntity) : UserDTO.createRemovedUserInstance();
```
Bert-R: If you just retain the below line, the API contract for ``createInstance`` can be regular:
```
UserDTO user = userEntity == null ? null : UserDTO.createInstance(userEntity);
```

Bert-R:
Comments added.

core/src/main/java/nu/yona/server/subscriptions/service/UserDTO.java
aristotelos: There were 4 callers that needed the `null` check. I have created an alternative method `createInstanceIfNotNull` that returns `null` or an instance and added some comments to the methods to explain it.

Bert-R: Yes, there are more of them. I didn't notice.

PR #93, opened 2016-01-21
YD-157 Migrated to HttpClientBuilder

aristotelos:
Not yet tested because I do not have a Plivo setup, can somebody do that for me?


aristotelos:
Just tested with Plivo developer account and it works

aristotelos:
Another option would be to use `com.plivo.helper.api.client.RestAPI`, see https://www.plivo.com/docs/getting-started/send-a-single-sms/

Bert-R:
Tested it and found and corrected a small glitch.

PR #94, opened 2016-01-22
YD-140 Using optional source phone number

aristotelos:


aristotelos:
Oops, this does not work! 

  Unexpected status code received from SMS service: 400. Message: { "api_id": "effe4abd-c0db-11e5-a8a2-22000ac69cc8", "error": "Destination Phone Numbers need to be verified. Please go to https://manage.plivo.com/sandbox-numbers/ to verify them." }

aristotelos:
See https://www.plivo.com/faq/phone-numbers/can-a-source-number-set-for-an-outbound-text-message-be-a-non-plivo-number/

PR #96, opened 2016-01-22
YD-167 updated the URL

Bert-R:
The change is simple enough to be understandable for people without Perl knowledge.


aristotelos:
I thought to do it myself indeed :)

PR #97, opened 2016-01-23
YD-81 Periodically refresh the Smoothwall categories list from the server

Bert-R:


scripts/smoothwall/HandleDansGuardianLog.pl
```diff
@@ -59,44 +93,52 @@ sub fetch_relevant_url_categories {
...
-	handle_records_from_stream(*INPUT_FILE, $analysis_engine_url);
+	handle_records_from_stream(*INPUT_FILE);
 	close INPUT_FILE;
 }
 
-my $analysis_engine_url = 'http://localhost:8081/analysisEngine/';
-GetOptions ('analysisEngineURL=s' => \$analysis_engine_url) or die "Usage: $0 [--analysisEngineURL <URL>] [<input file>]";
+$| = 1; # Make STDOUT unbuffered
+GetOptions ('analysisEngineURL=s' => \$analysis_engine_url,
+	'categoriesRefreshInterval=i' => \$categories_refresh_interval) or die "Usage: $0 [--analysisEngineURL <URL>] [<input file>]";
```
aristotelos: I do not see a Usage instruction for the `categoriesRefreshInterval` parameter here? Is it required? If so, I would expect it in the start-yona script below.

Bert-R: It's not required. The default value is 5 minutes (the 300 value a few
lines above this one).

aristotelos:
Reviewed it, only one comment made

scripts/smoothwall/HandleDansGuardianLog.pl
aristotelos: I see. It can be added to this line:

    or die "Usage: $0 [--analysisEngineURL <URL>] [--categoriesRefreshInterval <seconds integer>] [<input file>]"

Bert-R:
Extended the usage.

PR #100, opened 2016-02-02
YD-143 implement activity category caching

aristotelos:


Bert-R:
Looks good.

What about adding a test that adds an manipulates an activity category to see whether the cache update and eviction works properly? We don't have such a test at the moment, right?

aristotelos:
Adding the test was useful because the `importActivityCategories` method indeed did not use the 
`@CachePut` annotation correctly.
Because there are no controllers using the `ActivityCategoryService` fully yet, I had to add a JUnit test instead of an integration test. So I think [YD-88|http://jira.yona.nu/browse/YD-88] has been done as well.

PR #102, opened 2016-02-08
YD-165 implement goals controller, YD-46 embed goals in user

aristotelos:
Updating of goals not supported yet

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
@@ -33,22 +33,20 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
 	def addRichard()
 	{
 		def richard = appService.addUser(appService.&assertUserCreationResponseDetails, "R i c h a r d", "Richard", "Quinn", "RQ",
-				"+$timestamp", ["Nexus 6"], [
-					BudgetGoal.createNoGoInstance("news"),
-					BudgetGoal.createNoGoInstance("gambling")
-				])
+				"+$timestamp", ["Nexus 6"])
 		appService.confirmMobileNumber(appService.&assertResponseStatusSuccess, richard)
+		appService.addBudgetGoal(richard, BudgetGoal.createNoGoInstance("news"))
+		appService.addBudgetGoal(richard, BudgetGoal.createNoGoInstance("gambling"))
 		return richard
 	}
 
 	def addBob()
 	{
 		def bob = appService.addUser(appService.&assertUserCreationResponseDetails, "B o b", "Bob", "Dunn", "BD",
-				"+$timestamp", ["iPhone 5"], [
-					BudgetGoal.createNoGoInstance("news"),
-					BudgetGoal.createNoGoInstance("gambling")
-				])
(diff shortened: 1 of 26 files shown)


appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
```diff
@@ -33,22 +33,20 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
...
 	{
 		def richard = appService.addUser(appService.&assertUserCreationResponseDetails, "R i c h a r d", "Richard", "Quinn", "RQ",
-				"+$timestamp", ["Nexus 6"], [
-					BudgetGoal.createNoGoInstance("news"),
-					BudgetGoal.createNoGoInstance("gambling")
-				])
+				"+$timestamp", ["Nexus 6"])
 		appService.confirmMobileNumber(appService.&assertResponseStatusSuccess, richard)
+		appService.addBudgetGoal(richard, BudgetGoal.createNoGoInstance("news"))
+		appService.addBudgetGoal(richard, BudgetGoal.createNoGoInstance("gambling"))
```
Bert-R: Thanks for choosing this approach. The earlier "inline" approach where the goals were given during user creation was just a convenience. This is indeed the way to go.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -94,7 +92,7 @@ class CreateUserOnBuddyRequestTest extends AbstractAppServiceIntegrationTest
 		response.responseData.nickname == newNickname
 		response.responseData.devices.size() == 1
 		response.responseData.devices[0] == "iPhone 6"
-		response.responseData.goals.size() == 0 //TODO: updating of goals is not yet supported
+		//response.responseData._embedded._budgetGoals.size() == 0 //TODO: updating of goals is not yet supported
```
Bert-R: Why is this commented? I would expect Bob to still have zero goals, so this line can be written as 
```
!response.responseData._embedded?._budgetGoals?.size
```

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
aristotelos: There was no other way actually to do this nicely with HAL. At least Spring cannot deserialize a JSON object containing _embedded items I guess :)

Bert-R: I want to give that a try later. I'm still hopeful :)

Bert-R:
Besides one comment on ``CreateUserOnBuddyRequestTest``, it's fine with me.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
aristotelos: Agreed, thanks for the suggestion

aristotelos:
Thanks for the fast review! Have fixed the TODO.

PR #103, opened 2016-02-08
YD-165 Fixed broken test!

aristotelos:


aristotelos:
The test was in analysis service, while I only tested the app service integration tests. Shame on me!

Bert-R:
I did a direct commit to fix the test.

PR #105, opened 2016-02-09
YD-174 add collection links

aristotelos:
Checked all _embedded resources, and refactored the embedded buddies to a collection resource.
Next step is to use curies to make these links custom, I propose that this is picked up together with YD-172 or otherwise this task can remain open.

### DIFF
--- a/core/src/testUtils/groovy/nu/yona/server/test/User.groovy
+++ b/core/src/testUtils/groovy/nu/yona/server/test/User.groovy
@@ -50,9 +50,9 @@ class User
 		{
 			this.nickname = json.nickname
 
-			this.buddies = (json._embedded?.buddies && json._embedded.buddies._embedded) ? json._embedded.buddies._embedded.buddies.collect{new Buddy(it)} : []
+			this.buddies = (json._embedded?.buddies?._embedded) ? json._embedded.buddies._embedded.buddies.collect{new Buddy(it)} : []
 			this.devices = json.devices.collect{"$it"}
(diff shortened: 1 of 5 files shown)


core/src/testUtils/groovy/nu/yona/server/test/User.groovy
```diff
@@ -50,7 +50,7 @@ class User
 		{
 			this.nickname = json.nickname
 
-			this.buddies = (json._embedded?.buddies) ? json._embedded.buddies.collect{new Buddy(it)} : []
+			this.buddies = (json._embedded?.buddies && json._embedded.buddies._embedded) ? json._embedded.buddies._embedded.buddies.collect{new Buddy(it)} : []
```
Bert-R: This can be written simpler:
```
this.buddies = (json._embedded?.buddies?._embedded) ? json._embedded.buddies._embedded.buddies.collect{new Buddy(it)} : []
```

Bert-R: The work in this PR is good, but it is not addressing the JIRA issue yet, so I prefer to keep it open. Once YD-172 is done, this ticket can be taken on again to add the requested links.

aristotelos: OK no problem, I will move it to Open again.

aristotelos: Simplified the mentioned line

aristotelos:
Will move the task to Open again

Bert-R:
OK. Will merge this one after the current build is complete.

PR #106, opened 2016-02-09
YD-62 Transfering direct messages to anonymous on get

aristotelos:
Following implementation suggestion of @Bert-R


Bert-R:
I've merged the change set.
You plan to do the rename from "anonymous messages" to just "messages" (on REST API level) as a next step, right?

aristotelos:
Hmm, that can be done. I was hesitating because I thought there might still
be a POST to /user/x/messages/direct/ to send a direct message. What do you
think?

Bert-R:
We're currently not supporting post on these URLs. The buddy-to-buddy messaging will be implemented in a different way. The /messages URL of a user needs the device password of that user, so it's not possible to post for someone else.

PR #107, opened 2016-02-09
YD-62 Remove /anonymous/ from message path

aristotelos:


Bert-R:
Wow, that goes blazing fast.

PR #108, opened 2016-02-09
YD-173 Edit links

aristotelos:


appservice/src/main/java/nu/yona/server/goals/rest/GoalController.java
```diff
@@ -155,6 +156,7 @@ public GoalResource toResource(GoalDTO goal)
 			GoalResource goalResource = instantiateResource(goal);
 			ControllerLinkBuilder selfLinkBuilder = getGoalLinkBuilder(userID, goal);
 			addSelfLink(selfLinkBuilder, goalResource);
+			addRemoveLink(selfLinkBuilder, goalResource);
```
Bert-R: Rename into ``addEditLink``, as that is what's added.

appservice/src/main/java/nu/yona/server/messaging/rest/MessageController.java
```diff
@@ -189,6 +194,11 @@ private void addSelfLink(ControllerLinkBuilder selfLinkBuilder, MessageResource
 			messageResource.add(selfLinkBuilder.withSelfRel());
 		}
 
+		private void addDeleteLink(ControllerLinkBuilder selfLinkBuilder, MessageResource messageResource)
```
Bert-R: Rename into ``addEditLink``, as that is what's added.

appservice/src/main/java/nu/yona/server/subscriptions/rest/BuddyController.java
```diff
@@ -202,5 +204,10 @@ private void addSelfLink(ControllerLinkBuilder selfLinkBuilder, BuddyResource bu
 		{
 			buddyResource.add(selfLinkBuilder.withSelfRel());
 		}
+
+		private void addRemoveLink(ControllerLinkBuilder selfLinkBuilder, BuddyResource buddyResource)
```
Bert-R: Rename into ``addEditLink``, as that is what's added.

core/src/main/java/nu/yona/server/messaging/service/DiscloseResponseMessageDTO.java
```diff
@@ -53,6 +53,12 @@ public String getNickname()
 		return nickname;
 	}
 
+	@Override
+	public boolean canBeDeleted()
```
Bert-R: We should implement a test to verify deletion of such messages. In the original code they could not be deleted, but that wasn't noticed.

Bert-R:
Looks good. A few minor comments added.

aristotelos:
Rework done

PR #112, opened 2016-02-10
YD-180 Disclose messages extend BuddyMessage

Bert-R:
* DiscloseRequestMessageDTO and DiscloseResponseMessageDTO now extend BuddyMessageDTO
* Made two methods private that were accidentally protected
* Extended the disclosure test to verify the messages


appservice/src/intTest/groovy/nu/yona/server/DisclosureTest.groovy
```diff
@@ -118,6 +121,9 @@ class DisclosureTest extends AbstractAppServiceIntegrationTest
 
 		def getBobMessagesResponse = appService.getMessages(bob)
 		getBobMessagesResponse.status == 200
+		getBobMessagesResponse.responseData?._embedded?.discloseResponseMessages?.size
```
aristotelos: Maybe merge these lines with the added lines in PR 111

appservice/src/intTest/groovy/nu/yona/server/DisclosureTest.groovy
```diff
@@ -162,6 +169,9 @@ class DisclosureTest extends AbstractAppServiceIntegrationTest
 
 		def getBobMessagesResponse = appService.getMessages(bob)
 		getBobMessagesResponse.status == 200
```
aristotelos: Idem dito

core/src/main/java/nu/yona/server/messaging/service/DiscloseRequestMessageDTO.java
```diff
@@ -77,8 +71,9 @@ public Status getStatus()
 	public static DiscloseRequestMessageDTO createInstance(UserDTO actingUser, DiscloseRequestMessage messageEntity)
 	{
 		GoalConflictMessage targetGoalConflictMessage = messageEntity.getTargetGoalConflictMessage();
-		return new DiscloseRequestMessageDTO(messageEntity.getID(), messageEntity.getCreationTime(), messageEntity.getNickname(),
-				messageEntity.getStatus(), targetGoalConflictMessage.getOriginGoalConflictMessageID(),
+		return new DiscloseRequestMessageDTO(messageEntity.getID(), messageEntity.getCreationTime(), actingUser,
```
aristotelos: You use `actingUser` as a parameter, while the other subclasses of `BuddyMessageDTO` use `UserDTO.createInstanceIfNotNull(messageEntity.getUser())`. Could you please make these uniform? (See `BuddyConnectRequestMessageDTO.createInstance`, `BuddyConnectResponseMessageDTO.createInstance`, `BuddyDisconnectMessageDTO.createInstance`.) If the user is deleted, does `actingUser` not become `null`? The message factory may construct these messages (by an inbox request) after the related user has been deleted.

aristotelos:
Nice uniformization! Something to consider is to move `BuddyMessageDTO` and `BuddyMessage` to the `nu.yona.server.messaging` namespace.

aristotelos:
No more review comments :)

appservice/src/intTest/groovy/nu/yona/server/DisclosureTest.groovy
Bert-R: You mean #108?

core/src/main/java/nu/yona/server/messaging/service/DiscloseRequestMessageDTO.java
Bert-R: That adds complexity and overhead for no reason. In this scenario, the user must still exist,  because they are sending a disclosure request.

Bert-R:
> Nice uniformization! Something to consider is to move `BuddyMessageDTO` and `BuddyMessage` to the `nu.yona.server.messaging` namespace.
Makes sense. There is one more rename/move I want to do: from "disclose request" to "disclosure request", to make that proper English.

If you agree, I will make a PR for both changes

appservice/src/intTest/groovy/nu/yona/server/DisclosureTest.groovy
Bert-R: Done

appservice/src/intTest/groovy/nu/yona/server/DisclosureTest.groovy
Bert-R: Done

core/src/main/java/nu/yona/server/messaging/service/DiscloseRequestMessageDTO.java
aristotelos: I don't agree. The message still needs to be rendered after the user sending this message does no longer exist... So I guess the `UserDTO.createInstanceIfNotNull(messageEntity.getUser())` is the only correct solution.

Bert-R: You're even more right than you think. The receiver of this message should not see themselves but the sender of it, so the user should indeed be taken from the message entity. I'll change it later this evening.

Bert-R:
Created [YD-187](http://jira.yona.nu/browse/YD-187) for the rename/move of the message classes.

PR #114, opened 2016-02-12
YD-87 Unit tests for activity category service

aristotelos:
Split up the Spring-based test (needed for testing caching) and the unit test
Added test for importActivityCategories


core/src/test/java/nu/yona/server/goals/service/ActivityCategoryServiceTest.java
```diff
@@ -0,0 +1,121 @@
...
+	@Mock
+	ActivityCategoryRepository mockRepository = mock(ActivityCategoryRepository.class);
+	@InjectMocks
+	private ActivityCategoryService service = new ActivityCategoryService();
+
+	private ActivityCategory gambling;
+	private ActivityCategory news;
+
+	@Before
+	public void setUp()
```
Bert-R: This setup is a straight copy of  the one in ``ActivityCategoryServiceIntegrationTest``. Can't you put it somewhere where it can be shared? Common base class?

core/src/test/java/nu/yona/server/goals/service/ActivityCategoryServiceIntegrationTest.java
```diff
@@ -30,11 +29,11 @@
...
 import nu.yona.server.goals.entities.ActivityCategoryRepository;
-import nu.yona.server.goals.service.ActivityCategoryServiceTests.TestConfiguration;
+import nu.yona.server.goals.service.ActivityCategoryServiceIntegrationTest.TestConfiguration;
 
 @RunWith(SpringJUnit4ClassRunner.class)
 @ContextConfiguration(classes = { TestConfiguration.class })
-public class ActivityCategoryServiceTests
+public class ActivityCategoryServiceIntegrationTest
 {
 	private Set<ActivityCategory> activityCategories = new HashSet<ActivityCategory>();
```
Bert-R: Comment applies to the ``mockRepository`` below. Does it make sense to add add mock calls in the setup if you are autowiring the repository?

Bert-R:
I like the idea of splitting the different test approaches, but added a few questions.

core/src/test/java/nu/yona/server/goals/service/ActivityCategoryServiceIntegrationTest.java
aristotelos: Yes it does, because I am returning a mocked `ActivityCategoryRepository` repository bean in the `TestConfiguration` class below instead of a real instance.
Maybe there is some way to return a real instance that works nicely, but the example I found [http://stackoverflow.com/questions/27856266/how-to-make-instance-of-crudrepository-interface-during-testing-in-spring] is far more complicated.

core/src/test/java/nu/yona/server/goals/service/ActivityCategoryServiceTest.java
aristotelos: Done

PR #116, opened 2016-02-16
YD-158 Automatically adding mandatory goal

aristotelos:
Renamed property on ActivityCategory, blocking remove, added/updated
tests

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
@@ -165,14 +165,14 @@ public class UserService
 
 	private void addMandatoryGoals(User userEntity)
 	{
-		for (ActivityCategoryDTO activityCategory : activityCategoryService.getAllActivityCategories())
-		{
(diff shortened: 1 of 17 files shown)


core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -155,6 +163,18 @@ public UserDTO addUser(UserDTO user, Optional<String> overwriteUserConfirmationC
 		return userDTO;
 	}
 
+	private void addMandatoryGoals(User userEntity)
```
Bert-R: A simpler way to write this would be:
```
	private void addMandatoryGoals(User userEntity)
	{
		activityCategoryService.getAllActivityCategories().stream().filter(c -> c.isMandatoryNoGo())
				.forEach(c -> addGoal(userEntity, c));
	}

	private void addGoal(User userEntity, ActivityCategoryDTO category)
	{
		userEntity.getAnonymized()
				.addGoal(BudgetGoal.createNoGoInstance(ActivityCategory.getRepository().findOne(category.getID())));
	}
```

PR #118, opened 2016-02-18
YD-164 Added app activity controller

aristotelos:
Added integration test
Fixed a random integration test in EditGoalsTest

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
@@ -45,33 +45,65 @@ class EditGoalsTest extends AbstractAppServiceIntegrationTest
 		response.status == 200
 		response.responseData._links?.self.href == richard.url + appService.GOALS_PATH_FRAGMENT
 		response.responseData._embedded.budgetGoals.size() == 2
+		def gamblingGoals = response.responseData._embedded.budgetGoals.findAll{ it.activityCategoryName == 'gambling'}
+		gamblingGoals.size() == 1
+		!gamblingGoals[0]._links.edit //mandatory goal
+		def newsGoals = response.responseData._embedded.budgetGoals.findAll{ it.activityCategoryName == 'news'}
+		newsGoals.size() == 1
+		newsGoals[0]._links.edit.href
 	}
 
 	def 'Add goal'()
 	{
 		given:
-		def richard = addRichard()
+		def richardAndBob = addRichardAndBobAsBuddies()
+		def richard = richardAndBob.richard
+		def bob = richardAndBob.bob
 		when:
-		def addedGoal = appService.addBudgetGoal(appService.&assertResponseStatusCreated, richard, BudgetGoal.createInstance("social", 60))
+		def addedGoal = appService.addBudgetGoal(appService.&assertResponseStatusCreated, richard, BudgetGoal.createInstance("social", 60), "Going to monitor my social time!")
+
 		then:
 		addedGoal
+
 		def responseGoalsAfterAdd = appService.getGoals(richard)
 		responseGoalsAfterAdd.status == 200
 		responseGoalsAfterAdd.responseData._embedded.budgetGoals.size() == 3
+
+		def bobMessagesResponse = appService.getMessages(bob)
+		bobMessagesResponse.responseData._embedded.goalChangeMessages.size() == 1
+		bobMessagesResponse.responseData._embedded.goalChangeMessages[0].change == 'GOAL_ADDED'
+		bobMessagesResponse.responseData._embedded.goalChangeMessages[0].changedGoal.activityCategoryName == 'social'
+		bobMessagesResponse.responseData._embedded.goalChangeMessages[0].user.firstName == 'Richard'
+		bobMessagesResponse.responseData._embedded.goalChangeMessages[0].nickname == 'RQ'
+		bobMessagesResponse.responseData._embedded.goalChangeMessages[0].message == "Going to monitor my social time!"
+		bobMessagesResponse.responseData._embedded.goalChangeMessages[0]._links.edit
 	}
 
 	def 'Delete goal'()
 	{
 		given:
-		def richard = addRichard()
+		def richardAndBob = addRichardAndBobAsBuddies()
+		def richard = richardAndBob.richard
+		def bob = richardAndBob.bob
 		def addedGoal = appService.addBudgetGoal(appService.&assertResponseStatusCreated, richard, BudgetGoal.createInstance("social", 60))
 		when:
-		def response = appService.removeBudgetGoal(richard, addedGoal)
+		def response = appService.removeBudgetGoal(richard, addedGoal, "Don't want to monitor my social time anymore")
+
 		then:
 		response.status == 200
+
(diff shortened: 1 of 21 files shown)


appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
```diff
@@ -45,10 +45,6 @@ class EditGoalsTest extends AbstractAppServiceIntegrationTest
 		response.status == 200
 		response.responseData._links?.self.href == richard.url + appService.GOALS_PATH_FRAGMENT
 		response.responseData._embedded.budgetGoals.size() == 2
-		response.responseData._embedded.budgetGoals[1].activityCategoryName == 'gambling'
```
aristotelos: This was random failing

Bert-R: Instead of removing it, it's better to still verify it:
```
response.responseData._embedded.budgetGoals.findAll{ it.activityCategoryName == 'gambling'}.size() == 1
```

Bert-R: I applied this change directly on the master, to get the tests green again. See this change set: https://github.com/yonadev/yona-server/commit/6215e34b37c48dd25e3f1cd00ce3600a3a9b886e

Bert-R: Along with this correction: https://github.com/yonadev/yona-server/commit/d5fcac5c2631e9c5624a0ab2395ce87ebb873536 :(

Bert-R: The PR now has merge conflicts because of these changes in the master.

appservice/src/main/java/nu/yona/server/analysis/rest/ActivityController.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+import nu.yona.server.analysis.service.AnalysisEngineService;
+import nu.yona.server.analysis.service.AppActivityDTO;
+import nu.yona.server.crypto.CryptoSession;
+import nu.yona.server.subscriptions.service.UserService;
+
+@Controller
+@RequestMapping(value = "/users/{userID}/activity")
```
Bert-R: Given that we post only app activities, please rename {{activity}} into {{appActivity}} and remove the value from {{addAppActivity}} and rename the class accordingly ({{{AppActivityControler}}).

appservice/src/main/java/nu/yona/server/analysis/rest/ActivityController.java
```diff
@@ -0,0 +1,45 @@
...
+	private AnalysisEngineService analysisEngineService;
+
+	@Autowired
+	private UserService userService;
+
+	@RequestMapping(value = "/appActivity/", method = RequestMethod.POST)
+	@ResponseBody
+	@ResponseStatus(HttpStatus.OK)
+	public void addAppActivity(@RequestHeader(value = PASSWORD_HEADER) Optional<String> password, @PathVariable UUID userID,
+			@RequestBody AppActivityDTO appActivity)
```
Bert-R: This method now takes a single activity, but we would be taking a set of application activity entries. If the app does not have a network connection, it will need to store the app activities and post them once it can connect again.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -100,43 +118,64 @@ private void updateLastActivity(PotentialConflictDTO payload, Date activityEndTi
...
+				.collect(Collectors.toSet());
 		Set<Goal> goalsOfUser = userAnonymized.getGoals();
 		Set<Goal> matchingGoalsOfUser = goalsOfUser.stream()
 				.filter(g -> matchingActivityCategoryIDs.contains(g.getActivityCategory().getID())).collect(Collectors.toSet());
 		return matchingGoalsOfUser;
 	}
+
+	class ActivityPayload
+	{
+		public String url;
```
Bert-R: Using a struct-like approach with all public member variables is fine, but then make all members final, so they cannot be modified after creation.

core/src/main/java/nu/yona/server/goals/service/ActivityCategoryService.java
```diff
@@ -143,4 +143,19 @@ private void deleteRemovedActivityCategories(Set<ActivityCategory> activityCateg
 		activityCategoriesInRepository.stream().filter(ac -> !activityCategoryDTOsMap.containsKey(ac.getName()))
 				.forEach(this::deleteActivityCategory);
 	}
+
+	public Set<ActivityCategoryDTO> getMatchingActivityCategories(Set<String> smoothwallCategories)
```
Bert-R: The method names ``getMatchingActivityCategories`` and ``getAppActivityCategories`` are inconsistent. I propose to name them ``getMatchingCategoriesForSmoothwallCategories`` and ``getMatchingCategoriesForApp``.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -74,24 +83,33 @@ private void addOrUpdateActivity(PotentialConflictDTO payload, UserAnonymizedDTO
...
 		{
-			updateLastActivity(payload, now, matchingGoal, activity);
+			updateLastActivity(payload, userAnonymized, matchingGoal, activity);
 			cacheService.updateLatestActivityForUser(activity);
 		}
+		else if (payload.startTime.before(activity.getStartTime()))
+		{
+			// can happen if app activity is posted after some network activity is posted during the use of the same app
+			// TODO: remove activities after start time and log a new activity?
+			// or just overwrite start time of the existing activity, creating possible overlap?
```
Bert-R: What could go wrong if we just update the start time? If that's nothing or nothing dramatic, than I'd go for that.

Bert-R:
Added a few comments. Generally not fundamental, except for the controller (but that's little work).

appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
aristotelos: Thanks, that is an improvement.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
aristotelos: The problem is that we cache the last activity to update it. It may be that there are multiple activities added for network activity while the app is open, e.g. there may be a network activity period A1 at start time t1, another network activity period A2 at start time t2, and the app activity period at start time t0 may span A1 and A2. Updating the start time of A2 to t0 will cause that A1 and A2 overlap.

A purer solution would be to set the start time of A2 to t0 and remove A1. However, then it may be the case that there is already a goal conflict message sent for A1, so that maybe  needs to be relinked to the extended A2.

aristotelos:
Done, except for the last TODO. Maybe have a call about that one?

PR #119, opened 2016-02-20
YD-184 make messages polymorphic

Bert-R:
It has been a while, but finally it's done.


appservice/src/intTest/groovy/nu/yona/server/BasicBuddyTest.groovy
```diff
@@ -76,11 +76,13 @@ class BasicBuddyTest extends AbstractAppServiceIntegrationTest
 
 		then:
 		response.status == 200
-		response.responseData._embedded.buddyConnectRequestMessages[0].user.firstName == "Richard"
```
aristotelos: This line has gone. Per accident?

appservice/src/intTest/groovy/nu/yona/server/DisclosureTest.groovy
```diff
@@ -33,8 +32,7 @@ class DisclosureTest extends AbstractAppServiceIntegrationTest
 		messagesRichard[0]._links.requestDisclosure == null
 
 		responseBob.status == 200
-		responseBob.responseData?._embedded?.goalConflictMessages
```
aristotelos: You could add the assert `responseBob.responseData._embedded?.messages` here

appservice/src/main/java/nu/yona/server/messaging/rest/MessageController.java
```diff
@@ -141,40 +140,32 @@ public MessageActionResource(MessageActionDTO messageAction, UUID userID)
...
+			super(MessageController.class, MessageDTO.class);
 			this.userID = userID;
 		}
 
 		@Override
-		public MessageResource toResource(MessageDTO message)
+		public MessageDTO toResource(MessageDTO message)
 		{
-			MessageResource messageResource = instantiateResource(message);
+			message.removeLinks();
```
aristotelos: Why is this?

core/src/main/java/nu/yona/server/messaging/service/DiscloseRequestMessageDTO.java
```diff
@@ -47,6 +47,12 @@ private DiscloseRequestMessageDTO(UUID id, Date creationTime, UserDTO user, Stri
 	}
 
 	@Override
+	public String getType()
+	{
+		return "DiscloseRequestMessage";
```
aristotelos: It should be "disclos**ure** request" ;)

core/src/main/java/nu/yona/server/rest/JsonRootRelProvider.java
```diff
@@ -25,9 +29,9 @@
 	@Override
 	public String getItemResourceRelFor(Class<?> type)
 	{
-		JsonRootName[] jsonRootNameAnnotations = type.getAnnotationsByType(JsonRootName.class);
-		JsonRootName rootName = (jsonRootNameAnnotations.length == 0) ? null : jsonRootNameAnnotations[0];
-		return (rootName == null) ? defaultRelProvider.getItemResourceRelFor(type) : rootName.value();
+		Class<?> baseType = determineBaseType(type);
```
aristotelos: Maybe good to add a comment what you are doing here?

aristotelos:
Great work! I think the refactoring of all those tests must have been enjoyable... I already get tired by looking at it ;)

appservice/src/intTest/groovy/nu/yona/server/BasicBuddyTest.groovy
Bert-R: Not by accident. I felt checking the nickname is sufficient. But when I think about it again, I feel we must also check the embedded user object, so I've re-added it.

appservice/src/main/java/nu/yona/server/messaging/rest/MessageController.java
Bert-R: In the past, we were wrapping the DTOs in resources and adding the links there. Now we add the links to the DTOs. In my prototype, I reused the DTOs and then added the links multiple times. To prevent that, I clear the links firs, so we are sure it only has what we add here.

core/src/main/java/nu/yona/server/rest/JsonRootRelProvider.java
Bert-R: Extended the JavaDoc of this class to explain what it does.

Bert-R:
Pushed an update that should address your comments.

Bert-R:
It was a lot of fun to go through all these tests :)

PR #121, opened 2016-02-24
YD-191 Use standard date/time instead of epoch

Bert-R:
Extended the tests too

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
@@ -67,4 +67,18 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
 		int num = sequenceNumber++
 		return "$baseTimestamp$num"
 	}
+
+	def assertDateTimeInLimits(dateTimeString, int rangeSeconds = 10, Date comparisonDateTime = null)
+	{
+		// Example date string: 2016-02-23T21:28:58.556+0000
+		assert dateTimeString ==~ /[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\.[0-9]{3}\+0000/
+		Date dateTime = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ").parse(dateTimeString)
+		comparisonDateTime = (comparisonDateTime) ?: new Date()
+		int rangeMilliseconds = rangeSeconds * 1000
+
+		assert dateTime > new Date(comparisonDateTime.getTime() - rangeMilliseconds)
+		assert dateTime < new Date(comparisonDateTime.getTime() + rangeMilliseconds)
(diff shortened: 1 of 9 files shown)


appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
```diff
@@ -67,4 +67,18 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
 		int num = sequenceNumber++
 		return "$baseTimestamp$num"
 	}
-}
+
+	def assertDateTimeInLimits(dateTimeString, int rangeSeconds = 10, Date comparisonDateTime = null)
```
aristotelos: Possible alternative name is `assertDateEquals` with `rangeSeconds` as third parameter (default 0).

appservice/src/intTest/groovy/nu/yona/server/BasicBuddyTest.groovy
```diff
@@ -80,6 +80,7 @@ class BasicBuddyTest extends AbstractAppServiceIntegrationTest
 		def buddyConnectRequestMessages = response.responseData._embedded.messages.findAll{ it."@type" == "BuddyConnectRequestMessage"}
 		buddyConnectRequestMessages.size() == 1
 		buddyConnectRequestMessages[0].nickname == richard.nickname
+		assertDateTimeInLimits(buddyConnectRequestMessages[0].creationTime)
```
aristotelos: I think `assertDateEquals(buddyConnectRequestMessages[0].creationTime, new Date(), 10)` is more understandable

aristotelos:
Looks good, added a minor suggestion

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
Bert-R: I don't think that's a good idea:
* The ``equals`` part of the name conflicts with the range. If the range is non-zero, then it's a range or limits assertion.
* Making the default zero is a call for trouble for true client/server testing. Currently, the tests run against local host, but we must also be able to run them against a remote host. In that case, clock synchronization will be in place, but not up to the microsecond. That's why I made the range default 10 seconds.

Bert-R:
Replied your suggestions.

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
aristotelos: More frameworks have e.g. `assertEquals(double expected, double actual, double epsilon)`, so I do not agree with the first part. However, it is a good idea to not allow the comparison without the third parameter.

Bert-R: OK, I wasn't aware of that. Will make the change.

Bert-R:
Updated as requested.

PR #122, opened 2016-02-25
YD-172 Use URIs and curies for custom HATEOAS link relation types

Bert-R:


core/src/main/java/nu/yona/server/exceptions/InvalidDataException.java
```diff
@@ -44,7 +44,7 @@ public static InvalidDataException blankMobileNumber()
 
 	public static InvalidDataException invalidMobileNumber(String mobileNumber)
 	{
-		return new InvalidDataException("error.user.mobile.number.invalid");
+		return new InvalidDataException("error.user.mobile.number.invalid", mobileNumber);
```
aristotelos: Nice bugfix

appservice/src/main/java/nu/yona/server/subscriptions/rest/BuddyController.java
```diff
@@ -158,11 +166,11 @@ public BuddyResource(BuddyDTO buddy)
 			}
 
 			HashMap<String, Object> result = new HashMap<String, Object>();
-			result.put(BuddyDTO.USER_REL_NAME,
-					new UserController.UserResourceAssembler(false).toResource(getContent().getUser()));
+			result.put(curieProvider.getNamespacedRelFor(BuddyDTO.USER_REL_NAME),
```
aristotelos: Is this necessary? I read in the [documentation](http://docs.spring.io/spring-hateoas/docs/current/reference/html/#spis.curie-provider): "Note that now the prefix (...) automatically appears before all rels which are not registered with IANA, as in ex:orders."

aristotelos:
Added one question.

appservice/src/main/java/nu/yona/server/subscriptions/rest/BuddyController.java
Bert-R: Indeed, provided Spring HATEOAS it is a link. Here, we have a property with the name ``_embedded``, of type ``map``. There doesn't know it needs to do the replacement.

Bert-R:
Responded to it

PR #123, opened 2016-02-25
YD-194 Removed the type from the ResponseMessageDTO

Bert-R:
This response is only used for errors, so the type is pointless.


core/src/main/java/nu/yona/server/rest/ResponseMessageDTO.java
```diff
@@ -2,13 +2,9 @@
 
 /**
  * This DTO is used to communicate error messages or other type of messages to the client.
- * 
- * @author pgussow
  */
 public class ResponseMessageDTO
```
aristotelos: Consider to rename the class to `ErrorResponseDTO` or something like that.

aristotelos:
Added one suggestion.

core/src/main/java/nu/yona/server/rest/ResponseMessageDTO.java
Bert-R: Good one. Will do.

Bert-R:
Done.

PR #126, opened 2016-03-01
YD-174 More HATEOAS

Bert-R:
- Added more collection links (e.g. messages)
- Made the tests use the links instead of URL composition.


appservice/src/main/java/nu/yona/server/analysis/rest/AppActivityController.java
```diff
@@ -52,4 +58,28 @@ public void addAppActivity(@RequestHeader(value = PASSWORD_HEADER) Optional<Stri
...
 		});
 	}
+
+	@RequestMapping(value = "/", method = RequestMethod.GET)
+	@ResponseBody
+	@ResponseStatus(HttpStatus.OK)
+	public HttpEntity<AppActivityDTO[]> getAppActivities(@RequestHeader(value = PASSWORD_HEADER) Optional<String> password,
+			@PathVariable UUID userID)
+	{
+		// This method is just a shorthand to facilitate link building
```
aristotelos: I don't think this is necessary or good to do. The intention is to share a link to which the user can POST app activities. Therefore the link can use the `addAppActivity` method and the `getAppActivities` method is not necessary.

appservice/src/main/java/nu/yona/server/analysis/rest/AppActivityController.java
```diff
@@ -52,4 +58,28 @@ public void addAppActivity(@RequestHeader(value = PASSWORD_HEADER) Optional<Stri
...
+		return null;
+	}
+
+	public static Link getAppActivityLink(UUID userID)
+	{
+		try
+		{
+			ControllerLinkBuilder linkBuilder = linkTo(
+					methodOn(AppActivityController.class).getAppActivities(Optional.empty(), userID));
```
aristotelos: Please use `addAppActivity` method with third parameter `null`.

appservice/src/main/java/nu/yona/server/analysis/rest/AppActivityController.java
Bert-R: I tried that, but that doesn't work.

appservice/src/main/java/nu/yona/server/analysis/rest/AppActivityController.java
Bert-R: See previous comment.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
```diff
@@ -39,98 +30,6 @@ class YonaServer
 		formatter.format(new Date())
 	}
 
-	def deleteGoal(goalURL)
```
aristotelos: Nice cleanup :)

PR #127, opened 2016-03-02
YD-190 Make goals polymorphic

aristotelos:
Copying solution from Bert :)
Created common base class


aristotelos:
I just updated Swagger for this too

aristotelos:
Maybe it's an idea to have the swagger definition as a file in github too? Then the changes can be reviewed

Bert-R:
> Maybe it's an idea to have the swagger definition as a file in github too? Then the changes can be reviewed

Yes, that's the plan, see [YD-177](http://jira.yona.nu/browse/YD-177). I expect to do that this Saturday.

Bert-R:
> Maybe a more extensive explanation would be good here ;)
If time permits, I plan to write a [Gist`](https://help.github.com/articles/about-gists/) about our ``JsonRootRelProvider``. There I'll touch the polymorphic DTOs too, so that can act as a piece of documentation.

Bert-R:
Clean solution!

PR #129, opened 2016-03-03
YD-161 Added DayActivity and WeekActivity

aristotelos:
No controller yet
No aggregation properties/aggregation yet
Caching DayActivity per user


core/src/main/java/nu/yona/server/subscriptions/entities/BuddyAnonymized.java
```diff
@@ -13,7 +13,7 @@
 import nu.yona.server.entities.RepositoryProvider;
 
 @Entity
-@Table(name = "BUDDIES")
+@Table(name = "BUDDIES_ANONYMIZED")
```
Bert-R: Good catch.

core/src/main/java/nu/yona/server/subscriptions/entities/UserAnonymized.java
```diff
@@ -29,7 +29,7 @@ public static UserAnonymizedRepository getRepository()
 		return (UserAnonymizedRepository) RepositoryProvider.getRepository(UserAnonymized.class, UUID.class);
 	}
 
-	@ManyToOne
+	@OneToOne
```
Bert-R: This one too.

core/src/main/java/nu/yona/server/subscriptions/service/UserAnonymizedDTO.java
```diff
@@ -44,6 +44,11 @@ public UUID getID()
 		return goals;
 	}
 
+	public String getTimeZoneId()
```
Bert-R: What is this for?

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -0,0 +1,53 @@
...
+
+import javax.persistence.Entity;
+
+import nu.yona.server.entities.EntityWithID;
+import nu.yona.server.goals.entities.Goal;
+
+@Entity
+public abstract class IntervalActivity extends EntityWithID
+{
+	private UUID goalID;
```
Bert-R: Please make this a regular association.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -0,0 +1,53 @@
...
+	private UUID userAnonymizedID;
```
Bert-R: Same for this one.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -0,0 +1,53 @@
...
+	private ZonedDateTime zonedStartTime;
```
Bert-R: Just name it ``startTime``. Some day, we need to switch all date/times to "zoned".

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -0,0 +1,53 @@
...
+
+	public IntervalActivity(UUID id, UUID userAnonymizedID, UUID goalID, ZonedDateTime zonedStartTime)
```
Bert-R: Can be protected.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -0,0 +1,53 @@
...
+	{
+		return userAnonymizedID;
+	}
+
+	public UUID getGoalID()
+	{
+		return goalID;
+	}
+
+	public ZonedDateTime getZonedStartTime()
```
Bert-R: Rename in getStartTime

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -0,0 +1,53 @@
...
+	{
+		return zonedStartTime;
+	}
+
+	public Date getStartTime()
```
Bert-R: Why do you need this? Old consumers? Let's switch these to ``ZonedDateTime``.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -0,0 +1,53 @@
...
+	{
+		return Date.from(zonedStartTime.toInstant());
+	}
+
+	public abstract Date getEndTime();
```
Bert-R: Make this ``ZonedDateTime``.

core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
```diff
@@ -0,0 +1,66 @@
...
+	@OneToMany(cascade = CascadeType.ALL)
+	private List<Activity> activities;
+
+	// Default constructor is required for JPA
+	public DayActivity()
+	{
+		super(null, null, null, null);
+	}
+
+	public DayActivity(UUID id, UUID userAnonymizedID, UUID goalID, ZonedDateTime zonedStartOfDay, List<Activity> activities)
```
Bert-R: This one can be private, right? It's only used from the ``createInstance`` method.

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
```diff
@@ -0,0 +1,59 @@
...
+	@OneToMany(cascade = CascadeType.ALL)
+	private List<DayActivity> dayActivities;
+
+	// Default constructor is required for JPA
+	public WeekActivity()
+	{
+		super(null, null, null, null);
+	}
+
+	public WeekActivity(UUID id, UUID userAnonymizedID, UUID goalID, ZonedDateTime zonedStartOfWeek,
```
Bert-R: This one can be private, right? It's only used from the ``createInstance`` method.

core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
```diff
@@ -0,0 +1,66 @@
...
+		this.activities = activities;
+	}
+
+	@Override
+	public Date getEndTime()
+	{
+		return Date.from(getZonedStartTime().plusDays(1).toInstant());
+	}
+
+	public Activity getLatestActivity()
```
Bert-R: Please rename into ``getLastActivity``. That's more concise and common.

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
```diff
@@ -0,0 +1,59 @@
...
+		this.dayActivities = dayActivities;
+	}
+
+	@Override
+	public Date getEndTime()
+	{
+		return Date.from(getZonedStartTime().plusDays(7).toInstant());
+	}
+
+	public void addActivity(DayActivity dayActivity)
```
Bert-R: Please rename into ``addDayActivity``.

core/src/main/java/nu/yona/server/analysis/entities/DayActivityRepository.java
```diff
@@ -0,0 +1,18 @@
...
+import java.time.ZonedDateTime;
+import java.util.UUID;
+
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.data.repository.query.Param;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface DayActivityRepository extends CrudRepository<DayActivity, UUID>
```
Bert-R: Remove "zoned" everywhere. ``ZonedDateTime`` needs to become our new date class.

core/src/main/java/nu/yona/server/analysis/entities/WeekActivityRepository.java
```diff
@@ -0,0 +1,18 @@
...
+import java.time.ZonedDateTime;
+import java.util.UUID;
+
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.data.repository.query.Param;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface WeekActivityRepository extends CrudRepository<WeekActivity, UUID>
```
Bert-R: Remove "zoned" everywhere. ``ZonedDateTime`` needs to become our new (and thus common) date class.

core/src/main/java/nu/yona/server/analysis/service/ActivityDTO.java
```diff
@@ -0,0 +1,38 @@
...
+	private Date endTime;
+
+	@JsonCreator
+	public ActivityDTO(@JsonProperty("startTime") Date startTime, @JsonProperty("endTime") Date endTime)
+	{
+		this.startTime = startTime;
+		this.endTime = endTime;
+	}
+
+	public Date getStartTime()
```
Bert-R: Use ``ZonedDateTime``.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineCacheService.java
```diff
@@ -4,30 +4,38 @@
...
+	}
 
-		return results != null && !results.isEmpty() ? results.get(0) : null;
+	@CachePut(value = "dayActivities", key = "{#dayActivity.userAnonymizedID,#dayActivity.goalID,#dayActivity.zonedStartTime}")
+	public DayActivity updateDayActivityForUser(DayActivity dayActivity)
+	{
+		return DayActivity.getRepository().save(dayActivity);
+	}
+
+	public WeekActivity fetchWeekActivityForUser(UUID userAnonymizedID, UUID goalID, ZonedDateTime zonedStartOfWeek)
```
Bert-R: Why is week activity on the cache service but not cached?

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -72,16 +77,19 @@ private void analyze(ActivityPayload payload, UserAnonymizedDTO userAnonymized,
 	private void addOrUpdateActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal)
```
Bert-R: This method is growing too long. Please move the branches if the if/elsif into separate methods.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -89,21 +97,66 @@ private void addOrUpdateActivity(ActivityPayload payload, UserAnonymizedDTO user
...
 		{
-			updateLastActivity(payload, userAnonymized, matchingGoal, activity);
-			cacheService.updateLatestActivityForUser(activity);
+			updateLastActivity(dayActivity, payload, userAnonymized, matchingGoal, activity);
+			cacheService.updateDayActivityForUser(dayActivity);
 		}
 	}
 
-	private Activity createNewActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal)
+	private ZonedDateTime getZonedStartOfDay(Date startTime, UserAnonymizedDTO userAnonymized)
```
Bert-R: Remove "zoned" from the name.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -72,16 +77,19 @@ private void analyze(ActivityPayload payload, UserAnonymizedDTO userAnonymized,
 	private void addOrUpdateActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal)
 	{
 		Date minEndTime = new Date(payload.startTime.getTime() - yonaProperties.getAnalysisService().getConflictInterval());
-		Activity activity = cacheService.fetchLatestActivityForUser(userAnonymized.getID(), matchingGoal.getID(), minEndTime);
+		DayActivity dayActivity = cacheService.fetchDayActivityForUser(userAnonymized.getID(), matchingGoal.getID(),
+				getZonedStartOfDay(payload.startTime, userAnonymized));
+		Activity activity = dayActivity != null ? dayActivity.getLatestActivity() : null;
```
Bert-R: Swap the condition to prevent the negative test:
```
Activity activity = dayActivity == null ? null : dayActivity.getLatestActivity();
```

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -89,21 +97,66 @@ private void addOrUpdateActivity(ActivityPayload payload, UserAnonymizedDTO user
...
-	private Activity createNewActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal)
+	private ZonedDateTime getZonedStartOfDay(Date startTime, UserAnonymizedDTO userAnonymized)
 	{
-		return Activity.createInstance(userAnonymized.getID(), matchingGoal, payload.startTime, payload.endTime);
+		return startTime.toInstant().atZone(ZoneId.of(userAnonymized.getTimeZoneId())).truncatedTo(ChronoUnit.DAYS);
 	}
 
-	private void updateLastActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal,
-			Activity activity)
+	private ZonedDateTime getZonedStartOfWeek(Date startTime, UserAnonymizedDTO userAnonymized)
```
Bert-R: Same here.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -89,21 +97,66 @@ private void addOrUpdateActivity(ActivityPayload payload, UserAnonymizedDTO user
...
+		{
+			dayActivity = createNewDayActivity(payload, userAnonymized, matchingGoal);
+		}
+
+		Activity activity = Activity.createInstance(payload.startTime, payload.endTime);
+		dayActivity.addActivity(activity);
+		return dayActivity;
+	}
+
+	private DayActivity createNewDayActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal)
```
Bert-R: If two threads concurrently conclude that the day activity is missing, we end up with two entities for one day. We need to figure out how to handle that. I propose to create a separate ticket for this.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -124,11 +176,11 @@ private void updateLastActivity(ActivityPayload payload, UserAnonymizedDTO userA
 				.collect(Collectors.toSet());
 	}
 
-	private void sendConflictMessageToAllDestinationsOfUser(ActivityPayload payload, UserAnonymizedDTO userAnonymized,
-			Activity activity)
+	private void sendConflictMessageToAllDestinationsOfUser(DayActivity dayActivity, ActivityPayload payload,
```
Bert-R: Why is ``dayActivity`` passed? It's unused.

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
```diff
@@ -0,0 +1,55 @@
...
+import nu.yona.server.analysis.entities.IntervalActivity;
+import nu.yona.server.goals.service.GoalDTO;
+
+@JsonRootName("intervalActivity")
+public class IntervalActivityDTO
+{
+	private GoalDTO goal;
+	private Date startTime;
+	private Date endTime;
+	private Set<ActivityDTO> activities;
```
Bert-R: Shouldn't this be a list? It's ordered, right?

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
```diff
@@ -0,0 +1,55 @@
...
+
+	public IntervalActivityDTO(GoalDTO goal, Date startTime, Date endTime, Set<ActivityDTO> activities)
```
Bert-R: This one can be private

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
```diff
@@ -0,0 +1,55 @@
...
+import com.fasterxml.jackson.annotation.JsonRootName;
+
+import nu.yona.server.analysis.entities.IntervalActivity;
+import nu.yona.server.goals.service.GoalDTO;
+
+@JsonRootName("intervalActivity")
+public class IntervalActivityDTO
+{
+	private GoalDTO goal;
+	private Date startTime;
```
Bert-R: Please make this ``ZonedDateTime`` immediately.

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
Bert-R: What is this for? an ``IntervalActivity`` doesn't have activities inside, right?

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
```diff
@@ -0,0 +1,55 @@
...
+		return startTime;
+	}
+
+	public Date getEndTime()
+	{
+		return endTime;
+	}
+
+	@JsonIgnore
+	public GoalDTO getGoal()
```
Bert-R: This will become a link on the resource?

core/src/main/java/nu/yona/server/subscriptions/service/UserAnonymizedDTO.java
```diff
@@ -44,6 +44,11 @@ public UUID getID()
...
+	{
+		return "Europe/Paris";
```
Bert-R: This needs to become modifiable. Please create a ticket for this. As we start in NL, let's use "Europe/Amsterdam" as the initial fixed value.

Bert-R:
That's a considerable piece of code!
I've added some comments.

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
aristotelos: If we do not embed the goal, then I could make this a `UUID`.

Bert-R: Indeed. Embedding it implies that every day or week would bring it, but it's already in the user profile, so just a reference suffices.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
aristotelos: I am not sure if we want all date/times zoned. For a date on goal creation/message creation I think a UTC date is appropriate, it can be converted to the user zoned date when sending it to the client (or by the app itself). For the `DayActivity` and `WeekActivity` the time zone really does matter, it is impossible to save these as UTC, because around the globe the days (and weeks) start later or sooner. So I would say to keep all dates in UTC `Date` except for the `DayActivity` and `WeekActivity`.

Bert-R: With Java 8, [new date/time related classes](https://docs.oracle.com/javase/8/docs/api/index.html?java/time/LocalDateTime.html) are introduced as successors for the old ``java.util.Date``. There are now "local" date/time classes and there is "zoned" date/time. With "zoned", the time zone is embedded and thus can be converted to anything you want, including UTC. That way, the "old" class is obsolete. See also [this article](https://docs.oracle.com/javase/tutorial/datetime/iso/legacy.html).

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineCacheService.java
aristotelos: Because the code looks so much alike :). The reason that it is not cached is that I'm not sure the activities of a whole week for all users fits in memory.

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
aristotelos: This DTO is not used yet, but will be used as soon as the activity controller is created. The activity controller should return sub activities if they are not aggregated. Maybe it has to be reworked, so this is a stub.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
aristotelos: OK let's create a Task for that!

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
aristotelos: Agreed

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
Bert-R: That's fine. For the new code here, just use ``ZonedDateTime`` everywhere.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineCacheService.java
Bert-R: OK, fine. Let's go with this to start.

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
Bert-R: It doesn't need to return subactivities, but only the time spread..

core/src/main/java/nu/yona/server/analysis/service/ActivityDTO.java
Bert-R: Any particular reason why you're not using ``ZonedDateTime`` here?

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
Bert-R: Created [YD-199](http://jira.yona.nu/browse/YD-199) for this.

core/src/main/java/nu/yona/server/analysis/service/ActivityDTO.java
aristotelos: The `Activity` class already existed and was not updated, and contains `Date`s.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
aristotelos: Done

core/src/main/java/nu/yona/server/analysis/entities/WeekActivityRepository.java
aristotelos: Done

core/src/main/java/nu/yona/server/analysis/entities/DayActivityRepository.java
aristotelos: Done

core/src/main/java/nu/yona/server/analysis/service/ActivityDTO.java
Bert-R: OK, fine.

Bert-R:
Looks good. I'm going to merg it.

PR #130, opened 2016-03-04
YD-156 Added TimeZoneGoal

aristotelos:
Added a unit test to check that no conflict message gets sent, but activity gets recorded
Added a small integration test to add the goal


Bert-R:
Note that this PR has merge conflicts

core/src/main/java/nu/yona/server/goals/service/TimeZoneGoalDTO.java
```diff
@@ -0,0 +1,56 @@
...
+	private final String[] zones;
+
+	@JsonCreator
+	public TimeZoneGoalDTO(@JsonProperty("activityCategoryName") String activityCategoryName,
+			@JsonProperty("zones") String[] zones)
+	{
+		this(null, activityCategoryName, zones);
+	}
+
+	public TimeZoneGoalDTO(UUID id, String activityCategoryName, String[] zones)
```
Bert-R: Can be private

PR #131, opened 2016-03-05
YD-196 Messages now include a link when possible

Bert-R:
Till now, messages included an embedded User. This implies unnecessary overhead, has the user is available already as part of the buddy data.

The following messages now use a link to the user:

- BuddyConnectRequestMessage
- DisclosureRequestMessage
- DisclosureResponseMessage
- GoalChangeMessage

The following messages embed the user, as the user data might not be available in the buddy data (but now inside _embedded):

- BuddyConnectRequestMessage
- BuddyDisconnectMessage

Additional changes:

- The DisclosureRequestMessage does not embed the target anymore. There is a link, that's sufficient.
- The DisclosureResponseMessage now includes a "related" link for the target message

### DIFF
--- a/core/src/main/java/nu/yona/server/messaging/service/BuddyMessageEmbeddedUserDTO.java
+++ b/core/src/main/java/nu/yona/server/messaging/service/BuddyMessageEmbeddedUserDTO.java
@@ -1,0 +1,40 @@
+package nu.yona.server.messaging.service;
+
+import java.util.Collections;
+import java.util.Date;
+import java.util.Map;
+import java.util.UUID;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import nu.yona.server.subscriptions.service.UserDTO;
+
+public abstract class BuddyMessageEmbeddedUserDTO extends BuddyMessageDTO
+{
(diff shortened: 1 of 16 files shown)


core/src/main/java/nu/yona/server/messaging/service/BuddyMessageEmbeddedUserDTO.java
```diff
@@ -0,0 +1,40 @@
...
+		super(id, creationTime, user, nickname, message);
+	}
+
+	protected BuddyMessageEmbeddedUserDTO(UUID id, Date creationTime, UUID targetGoalConflictMessageOriginID, UserDTO user,
+			String nickname, String message)
+	{
+		super(id, creationTime, user, nickname, message);
+	}
+
+	@JsonProperty("_embedded")
```
aristotelos: This is a new pattern. Should we not keep this in the `ResourceAssembler` classes for consistency?

aristotelos:
Added one question. For the rest it looks great.

core/src/main/java/nu/yona/server/messaging/service/BuddyMessageEmbeddedUserDTO.java
Bert-R: This is not really new. On ``UserResource``, we do the same, but we cannot do that for polymorphic types. The resource assembler is responsible for filling the embedded map, but cannot define it.

PR #137, opened 2016-03-15
YD-166 activity collection controller

aristotelos:
Added pageable week and day views, with links to individual items (individual items can contain spread)
Added properties
To be done: validate if this is the right direction, then finish the spread computation, messages linking, individual items response detail, and extend the tests

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
@@ -22,14 +22,19 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		then:
 		response.status == 200
 		response.responseData._embedded
-		response.responseData._embedded."yona:dayActivities"
-		response.responseData._embedded."yona:dayActivities".size() == 1
-		//response.responseData._embedded."yona:dayActivities"[0].spread
-		response.responseData._embedded."yona:dayActivities"[0].totalActivityDurationMinutes == 1
-		response.responseData._embedded."yona:dayActivities"[0].goalAccomplished == false
-		response.responseData._embedded."yona:dayActivities"[0].totalMinutesBeyondGoal == 1
-		response.responseData._embedded."yona:dayActivities"[0].date
-		response.responseData._embedded."yona:dayActivities"[0].timeZoneId == "Europe/Amsterdam"
+		response.responseData._embedded."yona:dayActivityOverviews"
+		response.responseData._embedded."yona:dayActivityOverviews".size() == 1
+		def dayActivityOverview = response.responseData._embedded."yona:dayActivityOverviews"[0]
+		dayActivityOverview._embedded."yona:dayActivities"
+		dayActivityOverview._embedded."yona:dayActivities".size() == 1
+		def dayActivityForGoal = dayActivityOverview._embedded."yona:dayActivities"[0]
+		dayActivityForGoal.spread
+		dayActivityForGoal.spread.size() <= 96
+		dayActivityForGoal.totalActivityDurationMinutes == 1
+		dayActivityForGoal.goalAccomplished == false
+		dayActivityForGoal.totalMinutesBeyondGoal == 1
+		dayActivityForGoal.date
+		dayActivityForGoal.timeZoneId == "Europe/Amsterdam"
 	}
 
 	def 'Get week activity overviews'()
@@ -46,10 +51,14 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		response.responseData._embedded
 		response.responseData._embedded."yona:weekActivityOverviews"
 		response.responseData._embedded."yona:weekActivityOverviews".size() == 1
-		response.responseData._embedded."yona:weekActivityOverviews"[0]._embedded."yona:weekActivities"
-		//response.responseData._embedded."yona:weekActivityOverviews"[0]._embedded."yona:weekActivities"[0].spread
-		response.responseData._embedded."yona:weekActivityOverviews"[0]._embedded."yona:weekActivities"[0].totalActivityDurationMinutes == 1
-		response.responseData._embedded."yona:weekActivityOverviews"[0]._embedded."yona:weekActivities"[0].date
-		response.responseData._embedded."yona:weekActivityOverviews"[0]._embedded."yona:weekActivities"[0].timeZoneId == "Europe/Amsterdam"
+		def weekActivityOverview = response.responseData._embedded."yona:weekActivityOverviews"[0]
+		weekActivityOverview._embedded."yona:weekActivities"
+		weekActivityOverview._embedded."yona:weekActivities".size() == 1
+		def weekActivityForGoal = weekActivityOverview._embedded."yona:weekActivities"[0]
+		weekActivityForGoal.spread
+		weekActivityForGoal.spread.size() <= 96
+		weekActivityForGoal.totalActivityDurationMinutes == 1
+		weekActivityForGoal.date
+		weekActivityForGoal.timeZoneId == "Europe/Amsterdam"
 	}
 }
--- a/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
@@ -10,7 +10,7 @@ import groovy.json.*
 
 class ActivityTest extends AbstractAppServiceIntegrationTest
 {
-	def 'Get days activity'()
+	def 'Get day activity overviews'()
 	{
 		given:
 		def richard = addRichard()
@@ -32,7 +32,7 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		response.responseData._embedded."yona:dayActivities"[0].timeZoneId == "Europe/Amsterdam"
 	}
 
-	def 'Get weeks activity'()
+	def 'Get week activity overviews'()
 	{
 		given:
 		def richard = addRichard()
@@ -44,11 +44,12 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		then:
 		response.status == 200
 		response.responseData._embedded
-		response.responseData._embedded."yona:weekActivities"
-		response.responseData._embedded."yona:weekActivities".size() == 1
-		//response.responseData._embedded."yona:weekActivities"[0].spread
-		response.responseData._embedded."yona:weekActivities"[0].totalActivityDurationMinutes == 0
-		response.responseData._embedded."yona:weekActivities"[0].date
-		response.responseData._embedded."yona:weekActivities"[0].timeZoneId == "Europe/Amsterdam"
+		response.responseData._embedded."yona:weekActivityOverviews"
+		response.responseData._embedded."yona:weekActivityOverviews".size() == 1
+		response.responseData._embedded."yona:weekActivityOverviews"[0]._embedded."yona:weekActivities"
+		//response.responseData._embedded."yona:weekActivityOverviews"[0]._embedded."yona:weekActivities"[0].spread
+		response.responseData._embedded."yona:weekActivityOverviews"[0]._embedded."yona:weekActivities"[0].totalActivityDurationMinutes == 0
+		response.responseData._embedded."yona:weekActivityOverviews"[0]._embedded."yona:weekActivities"[0].date
+		response.responseData._embedded."yona:weekActivityOverviews"[0]._embedded."yona:weekActivities"[0].timeZoneId == "Europe/Amsterdam"
 	}
 }
--- a/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
@@ -1,0 +1,40 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Stichting Yona Foundation
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v.2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at https://mozilla.org/MPL/2.0/.
+ *******************************************************************************/
+package nu.yona.server
+
+import groovy.json.*
+
+class ActivityTest extends AbstractAppServiceIntegrationTest
+{
+	def 'Get days activity'()
+	{
+		given:
+		def richard = addRichard()
+		analysisService.postToAnalysisEngine(richard, ["Gambling"], "http://www.poker.com")
+
+		when:
+		def response = appService.getDaysActivity(richard)
+
+		then:
+		response.status == 200
+		response.responseData
+	}
+
+	def 'Get weeks activity'()
+	{
+		given:
+		def richard = addRichard()
+		analysisService.postToAnalysisEngine(richard, ["Gambling"], "http://www.poker.com")
+
+		when:
+		def response = appService.getWeeksActivity(richard)
+
+		then:
+		response.status == 200
+		response.responseData
+	}
+}
--- a/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
@@ -27,6 +27,7 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		//response.responseData._embedded."yona:dayActivities"[0].spread
 		response.responseData._embedded."yona:dayActivities"[0].totalActivityDurationMinutes == 0
 		response.responseData._embedded."yona:dayActivities"[0].goalAccomplished == false
+		response.responseData._embedded."yona:dayActivities"[0].totalMinutesBeyondGoal == 0
 		response.responseData._embedded."yona:dayActivities"[0].date
 		response.responseData._embedded."yona:dayActivities"[0].timeZoneId == "Europe/Amsterdam"
 	}
--- a/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
@@ -17,7 +17,7 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		analysisService.postToAnalysisEngine(richard, ["Gambling"], "http://www.poker.com")
 
 		when:
-		def response = appService.getDaysActivity(richard)
+		def response = appService.getDayActivityOverviews(richard)
 
 		then:
 		response.status == 200
@@ -38,7 +38,7 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		analysisService.postToAnalysisEngine(richard, ["Gambling"], "http://www.poker.com")
 
 		when:
-		def response = appService.getWeeksActivity(richard)
+		def response = appService.getWeekActivityOverviews(richard)
 
 		then:
 		response.status == 200
--- a/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
@@ -33,8 +33,39 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		dayActivityForGoal.totalActivityDurationMinutes == 1
 		dayActivityForGoal.goalAccomplished == false
 		dayActivityForGoal.totalMinutesBeyondGoal == 1
-		dayActivityForGoal.date
+		dayActivityForGoal.date =~ /\d{4}\-\d{2}\-\d{2}/
 		dayActivityForGoal.timeZoneId == "Europe/Amsterdam"
+		dayActivityForGoal._links."yona:goal"
+		dayActivityForGoal._links.self
+	}
+
+	def 'Get day activity detail'()
+	{
+		given:
+		def richard = addRichard()
+		analysisService.postToAnalysisEngine(richard, ["Gambling"], "http://www.poker.com")
+
+		when:
+		def overviewsResponse = appService.getDayActivityOverviews(richard)
+		overviewsResponse.responseData._embedded."yona:dayActivityOverviews"
+		overviewsResponse.responseData._embedded."yona:dayActivityOverviews".size() == 1
+		def dayActivityOverview = overviewsResponse.responseData._embedded."yona:dayActivityOverviews"[0]
+		dayActivityOverview._embedded."yona:dayActivities"
+		dayActivityOverview._embedded."yona:dayActivities".size() == 1
+		def dayActivityForGoal = dayActivityOverview._embedded."yona:dayActivities"[0]
+		dayActivityForGoal._links.self
+		def response = appService.getResourceWithPassword(dayActivityForGoal._links.self.href, richard.password)
+
+		then:
+		response.status == 200
+		response.responseData.spread
+		response.responseData.spread.size() <= 96
+		response.responseData.totalActivityDurationMinutes == 1
+		response.responseData.goalAccomplished == false
+		response.responseData.totalMinutesBeyondGoal == 1
+		response.responseData.date =~ /\d{4}\-\d{2}\-\d{2}/
+		response.responseData.timeZoneId == "Europe/Amsterdam"
+		response.responseData._links."yona:goal"
 	}
 
 	def 'Get week activity overviews'()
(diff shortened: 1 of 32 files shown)


core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -12,16 +17,44 @@
...
+	private LocalDate date;
+
+	/*
+	 * The start date and time with saved time zone.
+	 */
 	private ZonedDateTime startTime;
 
-	protected IntervalActivity(UUID id, Goal goal, ZonedDateTime startTime)
+	@ElementCollection
+	private List<Integer> spread;
```
aristotelos: Maybe an alternative is to use a `String` here and use a `@Transient` array that is joined in this string (e.g. comma-separated).

core/src/main/java/nu/yona/server/goals/entities/TimeZoneGoal.java
```diff
@@ -43,4 +45,11 @@ public boolean isNoGoGoal()
 	{
 		return false;
 	}
+
+	@Override
+	public boolean isGoalAccomplished(DayActivity dayActivity)
+	{
+		// TODO: zones should be parsed? maybe in spread format
+		return true;
```
aristotelos: Maybe it is easier to use 96 `boolean`s to model the allowed time zones instead of just `String`s. If `String` is still preferred, we need to establish a fixed format that can be parsed, e.g. "HH:mm-HH:mm".

appservice/src/main/java/nu/yona/server/analysis/rest/ActivityController.java
```diff
@@ -0,0 +1,212 @@
...
+@RequestMapping(value = "/users/{userID}/activity")
+public class ActivityController
+{
+	@Autowired
+	private AnalysisEngineService analysisEngineService;
+
+	@Autowired
+	private UserService userService;
+
+	@RequestMapping(value = "/week/", method = RequestMethod.GET)
```
Bert-R: This is a collection, so it should be plural (``/weeks/``).

appservice/src/main/java/nu/yona/server/analysis/rest/ActivityController.java
```diff
@@ -0,0 +1,212 @@
...
+	@ResponseBody
+	public HttpEntity<PagedResources<WeekActivityResource>> getWeeksActivity(
+			@RequestHeader(value = PASSWORD_HEADER) Optional<String> password, @PathVariable UUID userID, Pageable pageable,
+			PagedResourcesAssembler<WeekActivityDTO> pagedResourcesAssembler)
+	{
+		return CryptoSession.execute(password, () -> userService.canAccessPrivateData(userID),
+				() -> new ResponseEntity<>(
+						pagedResourcesAssembler.toResource(analysisEngineService.getWeekActivity(userID, pageable),
```
Bert-R: The analysis engine has no role here. It has done its work already and attached the result of its analysis to the user. I propose to create an ``ActivityService`` for this purpose.

appservice/src/main/java/nu/yona/server/analysis/rest/ActivityController.java
```diff
@@ -0,0 +1,212 @@
...
+{
+	@Autowired
+	private AnalysisEngineService analysisEngineService;
+
+	@Autowired
+	private UserService userService;
+
+	@RequestMapping(value = "/week/", method = RequestMethod.GET)
+	@ResponseBody
+	public HttpEntity<PagedResources<WeekActivityResource>> getWeeksActivity(
```
Bert-R: As said on the wiki, I don't think we can work without a date here. The starting point is a specific date, not a page number. I'm not sure whether we can do that with ``PagedResources`` or need to use custom pagination.

Bert-R: Propose to name this method ``getWeekActivityOverviews``.

appservice/src/main/java/nu/yona/server/analysis/rest/ActivityController.java
```diff
@@ -0,0 +1,212 @@
...
+		return CryptoSession.execute(password, () -> userService.canAccessPrivateData(userID),
+				() -> new ResponseEntity<>(
+						pagedResourcesAssembler.toResource(analysisEngineService.getWeekActivity(userID, pageable),
+								new WeekActivityResourceAssembler(userID)),
+						HttpStatus.OK));
+	}
+
+	@RequestMapping(value = "/day/", method = RequestMethod.GET)
+	@ResponseBody
+	public HttpEntity<PagedResources<DayActivityResource>> getDaysActivity(
```
Bert-R: Similar comments apply.

appservice/src/main/java/nu/yona/server/analysis/rest/ActivityController.java
```diff
@@ -0,0 +1,212 @@
...
+				() -> new ResponseEntity<>(
+						pagedResourcesAssembler.toResource(analysisEngineService.getDayActivity(userID, pageable),
+								new DayActivityResourceAssembler(userID)),
+						HttpStatus.OK));
+	}
+
+	@RequestMapping(value = "/week/{week}", method = RequestMethod.GET)
+	@ResponseBody
+	public HttpEntity<WeekActivityResource> getWeekActivity(@RequestHeader(value = PASSWORD_HEADER) Optional<String> password,
+			@PathVariable UUID userID, @PathVariable(value = "week") String weekStr, @RequestParam(value = "goal") UUID goalID)
```
Bert-R: Why do you use a query parameter for the goal? Can't that become a path variable too?

appservice/src/main/java/nu/yona/server/analysis/rest/ActivityController.java
```diff
@@ -0,0 +1,212 @@
...
+			super(ActivityController.class, WeekActivityResource.class);
+			this.userID = userID;
+		}
+
+		@Override
+		public WeekActivityResource toResource(WeekActivityDTO weekActivity)
+		{
+			WeekActivityResource weekActivityResource = instantiateResource(weekActivity);
+			addSelfLink(weekActivityResource);
+			addGoalLink(weekActivityResource);
```
Bert-R: Yep, this is the right thing to do.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -423,6 +426,18 @@ private static void addConfirmMobileNumberLink(Resource<UserDTO> userResource, S
 			userResource.add(UserController.getConfirmMobileLink(userResource.getContent().getID()));
 		}
 
+		private void addWeeksActivityLink(UserResource userResource)
```
Bert-R: Good to see the links here too.
Please apply the renames as I have proposed for the controller methods.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -438,6 +438,41 @@ paths:
           description: Unexpected error
           schema:
             $ref: '#/definitions/Error'
+  '/users/{userID}/activity/days/':
```
Bert-R: This is correctly plural. The controller has it as ``/day/``

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -438,6 +438,41 @@ paths:
...
+        type: string
+    get:
+      summary: Get day network and app activity
+      description: Get activity of the current day, pageable to yesterday and before 
+      parameters:
+        - name: size
+          in: query
+          description: The number of elements to be fetched
+          required: false
+          default: 10
```
Bert-R: Make this 3, in accordance with the [functional requirements](http://wiki.yona.nu/display/DEV/Me+-+Day)

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -850,12 +885,12 @@ definitions:
     allOf:
       - $ref: '#/definitions/PostPutGoal'
     properties:
-      maxDuration:
+      maxDurationMinutes:
         type: integer
         description: The budget (duration) for this goal
```
Bert-R: Please change into "The budget (duration) for this goal in minutes"

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1166,6 +1201,116 @@ definitions:
...
+            description: Link to the goal for which activity was recorded
+      spread:
+        type: array
+        items:
+          type: integer
+        description: Aggregates all activity on every quarter hour of the day (of which there are 96)
+      totalActivityDurationMinutes:
+        type: integer
+        description: Total minutes spent on the activity category during this interval
+      timeZoneId:
```
Bert-R: Fancy! So we can support traveling users.
Isn't this too fancy? The days can overlap in this way, right?

core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
```diff
@@ -70,8 +84,104 @@ public void addActivity(Activity activity)
 		this.activities.add(activity);
 	}
 
+	@Override
+	protected List<Integer> computeSpread()
```
Bert-R: Just a quick comment (I need more time to review in detail): this method is too long.

core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
```diff
@@ -70,8 +84,104 @@ public void addActivity(Activity activity)
...
+				}
+				i++;
+			}
+
+			addToSpread(result, activity.getStartTime(), activityBlockEndTime);
+		}
+		return result;
+	}
+
+	private void addToSpread(List<Integer> result, Date activityBlockStartTime, Date activityBlockEndTime)
```
Bert-R: Same here.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -29,10 +62,53 @@ public Goal getGoal()
...
+		}
+
+		return computeTotalActivityDurationMinutes();
+	}
+
+	protected abstract int computeTotalActivityDurationMinutes();
+
+	protected static List<Integer> getEmptySpread()
+	{
+		return Arrays.asList(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
Bert-R: For sake of readability, I'd rather write that this way:
```
return Collections.nCopies(96, 0);
```

core/src/main/java/nu/yona/server/goals/entities/TimeZoneGoal.java
Bert-R: That text format sounds good. We currently don't constrain the time zones to 15-minute boundaries. I'm writing a mail to Jan and Cornelis. I'll include that proposal. With that, we can internally make a translation to 96 booleans if we want. Note that we will need to store the original time spans too.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
Bert-R: I prefer not to do that, as math would become impossible.

Bert-R:
You've gone a long way already! I have provided some comments.
I propose we first settle on the API through [the wiki page](http://wiki.yona.nu/display/DEV/Design+change+to+capture+app+activity). Once that's done, we can write the swagger spec and do the implementation.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
aristotelos: That's why I wrote it in this way: I thought you would pick it out at review and give a better suggestion :)

Bert-R: :)
An alternative would be
```
return Arrays.asList(IntStream.iterate(0, i -> i).limit(96).toArray());
```

appservice/src/main/resources/static/swagger/swagger-spec.yaml
aristotelos: I think it is a good feature to support. Otherwise the Yona app becomes useless during holidays, while it makes sense to control your internet activity then. My idea to solve this:

- Always add activities to the current day using the day date. This means if you travel, the day activity can contain activities from more or less than 24 hours.
- If you travel backwards in time we have an issue: the current cached day is tomorrow. My proposal is to add the activities to tomorrow in this case.

So we can solve the problems in a 'practical way' that is not completely accurate when you are travelling. I think that is acceptible.

appservice/src/main/java/nu/yona/server/analysis/rest/ActivityController.java
aristotelos: I thought it would be more explicit that we filter on a goal here. Otherwise that is not visible from the URL.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
aristotelos: It will be stored using a mapping table now, which is less effective.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
aristotelos: Fixed

core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
aristotelos: Made it a bit shorter.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
Bert-R: I see. How does JPA do an array? Will that be a mapping table too?

appservice/src/main/resources/static/swagger/swagger-spec.yaml
Bert-R: Let's go with this. Traveling users will understand the complexity of the subject. If we don't do anything for it, it's certainly wrong.

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -0,0 +1,54 @@
...
+		def response = appService.getDayActivityOverviews(richard)
+
+		then:
+		response.status == 200
+		response.responseData._embedded
+		response.responseData._embedded."yona:dayActivities"
+		response.responseData._embedded."yona:dayActivities".size() == 1
+		//response.responseData._embedded."yona:dayActivities"[0].spread
+		response.responseData._embedded."yona:dayActivities"[0].totalActivityDurationMinutes == 0
+		response.responseData._embedded."yona:dayActivities"[0].goalAccomplished == false
```
Bert-R: This is weird. The duration beyond goal is zero, but the goal is still not accomplished. Technically, this is understandable, but functionally, it's wrong. I propose we fix this in the analysis engine. If we get the first web event, we set the end time one second past the start time. And when we round seconds to minutes, we ensure we don't round to zero.

With that, the story is functionally correct: the user did something outside the goal limits for a small period of time. Technically, we can now do away wit the ``goalAccomplished`` flag, as ``totalMinutesBeyondGoal`` already communicates the status.

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -0,0 +1,54 @@
...
+		response.responseData._embedded."yona:dayActivities"[0].totalMinutesBeyondGoal == 0
+		response.responseData._embedded."yona:dayActivities"[0].date
```
Bert-R: Please add a date assert

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -0,0 +1,54 @@
...
+		when:
+		def response = appService.getWeekActivityOverviews(richard)
+
+		then:
+		response.status == 200
+		response.responseData._embedded
+		response.responseData._embedded."yona:weekActivities"
+		response.responseData._embedded."yona:weekActivities".size() == 1
+		//response.responseData._embedded."yona:weekActivities"[0].spread
+		response.responseData._embedded."yona:weekActivities"[0].totalActivityDurationMinutes == 0
```
Bert-R: See comment above about ``goalAccomplished``. Here we should assert it's 1 minute.

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
```diff
@@ -0,0 +1,52 @@
...
+import nu.yona.server.analysis.entities.WeekActivity;
+import nu.yona.server.subscriptions.service.UserService;
+
+@Service
+public class ActivityService
+{
+	@Autowired
+	private UserService userService;
+
+	public Page<WeekActivityDTO> getWeekActivity(UUID userID, Pageable pageable)
```
Bert-R: Rename in accordance with the controller name: ``getWeekActivityOverviews``.

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
```diff
@@ -0,0 +1,52 @@
...
+	{
+		UUID userAnonymizedID = userService.getPrivateUser(userID).getPrivateData().getUserAnonymizedID();
+		Page<WeekActivity> weekActivityEntitiesPage = WeekActivity.getRepository().findAll(userAnonymizedID, pageable);
+		return new PageImpl<WeekActivityDTO>(weekActivityEntitiesPage.getContent().stream()
+				.map(a -> WeekActivityDTO.createInstance(a)).collect(Collectors.toList()), pageable,
+				weekActivityEntitiesPage.getTotalElements());
+	}
+
+	public Page<DayActivityDTO> getDayActivity(UUID userID, Pageable pageable)
```
Bert-R: Rename in accordance with the controller name: ``getDayActivityOverviews``.

core/src/main/java/nu/yona/server/analysis/service/WeekActivityDTO.java
```diff
@@ -0,0 +1,39 @@
...
+import java.time.format.DateTimeFormatter;
+import java.util.List;
+import java.util.UUID;
+
+import com.fasterxml.jackson.annotation.JsonRootName;
+
+import nu.yona.server.analysis.entities.WeekActivity;
+
+@JsonRootName("weekActivity")
+public class WeekActivityDTO extends IntervalActivityDTO
```
Bert-R: Rename into ``WeekActivityOverviewDTO`` (together with the JSON root name)

core/src/main/java/nu/yona/server/analysis/service/DayActivityDTO.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.UUID;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonRootName;
+
+import nu.yona.server.analysis.entities.DayActivity;
+import nu.yona.server.messaging.service.MessageDTO;
+
+@JsonRootName("dayActivity")
+public class DayActivityDTO extends IntervalActivityDTO
```
Bert-R: Rename into ``DayActivityOverviewDTO`` (together with the JSON root name)

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
aristotelos: I think there is no support for array in JPA?

core/src/main/java/nu/yona/server/analysis/service/WeekActivityDTO.java
aristotelos: Added another class with that name. This class will remain because it maps one to one with `WeekActivity` (which is the activity for a specific goal) and will be grouped into a `WeekActivityOverviewDTO`.

appservice/src/main/java/nu/yona/server/analysis/rest/ActivityController.java
```diff
@@ -0,0 +1,271 @@
...
+
+	@Autowired
+	private UserService userService;
+
+	@Autowired
+	private CurieProvider curieProvider;
+
+	@RequestMapping(value = "/weeks/", method = RequestMethod.GET)
+	@ResponseBody
+	public HttpEntity<PagedResources<WeekActivityOverviewResource>> getWeekActivityOverviews(
```
Bert-R: We need methods for individual week overviews too, right? For the self-link of the items in the collection. Something like this:
```
	@RequestMapping(value = "/weeks/{week}", method = RequestMethod.GET)
	@ResponseBody
	public HttpEntity<PagedResources<WeekActivityOverviewResource>> getWeekActivityOverview(
			@RequestHeader(value = PASSWORD_HEADER) Optional<String> password, @PathVariable UUID userID, @PathVariable String week)
```

aristotelos: Can be done, but why is it necessary?

appservice/src/main/java/nu/yona/server/analysis/rest/ActivityController.java
```diff
@@ -0,0 +1,271 @@
...
+
+	static class WeekActivityResource extends Resource<WeekActivityDTO>
+	{
+		public WeekActivityResource(WeekActivityDTO weekActivity)
+		{
+			super(weekActivity);
+		}
+	}
+
+	static class WeekActivityOverviewResource extends Resource<WeekActivityOverviewDTO>
```
Bert-R: This class corresponds with a collection of week activity overviews, right? I wonder why that is required. We don't have a class for this in the ``MessageControler``, right?

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -438,6 +438,41 @@ paths:
...
+          required: false
+          default: 0
+          type: number
+      tags:
+        - Activity
+      responses:
+        '200':
+          description: Activity of this user.
+          schema:
+            $ref: '#/definitions/PagedDayActivity'
```
Bert-R: Shouldn't this be ``PagedDayActivityOverview``, to stay in line with the rest of the terminology?

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1166,6 +1201,119 @@ definitions:
...
+          - $ref: '#/definitions/SelfLink'
+        properties:
+          next:
+            properties:
+              href:
+                type: string
+                description: Link to next page, if available
+            required:
+              - href
+          self:
```
Bert-R: This is already included through the ``allOf`` above.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1166,6 +1201,119 @@ definitions:
...
+                description: Link to previous page, if available
+            required:
+              - href
+        required:
+          - next
+          - self
+          - prev
+      _embedded:
+        properties:
+          "yona:dayActivities":
```
Bert-R: This should be ``dayActivityOverviews``.

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -0,0 +1,55 @@
...
+		def richard = addRichard()
+		analysisService.postToAnalysisEngine(richard, ["Gambling"], "http://www.poker.com")
+
+		when:
+		def response = appService.getDayActivityOverviews(richard)
+
+		then:
+		response.status == 200
+		response.responseData._embedded
+		response.responseData._embedded."yona:dayActivities"
```
Bert-R: Please rename into ``dayActivityOverviews``, in accordance with the ``weekActivityOverviews``

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1166,6 +1201,119 @@ definitions:
...
+      - page
+  DayActivity:
+    allOf:
+      - $ref: '#/definitions/IntervalActivity'
+    properties:
+      date:
+        type: string
+        example: 2016-03-16
+        description: The day date
+      goalAccomplished:
```
Bert-R: If we ensure the ``totalMinutesBeyondGoal`` is always > 0 when the goal is not accomplished, we don't need this boolean.

core/src/main/java/nu/yona/server/analysis/entities/Activity.java
```diff
@@ -48,6 +49,11 @@ public void setEndTime(Date endTime)
 		this.endTime = endTime;
 	}
 
+	public int getDurationMinutes()
+	{
+		return (int) TimeUnit.MILLISECONDS.toMinutes(this.getEndTime().getTime() - this.getStartTime().getTime());
```
Bert-R: This should return 1 if the time difference between start and end is > 0 (in any unit), to ensure we never round something to nothing.

core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
```diff
@@ -30,18 +34,23 @@ public static DayActivityRepository getRepository()
 	@OneToMany(cascade = CascadeType.ALL)
 	private List<Activity> activities;
 
+	private boolean goalAccomplished;
```
Bert-R: See earlier comment on this boolean.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
Bert-R: Let's keep it simple now. We can easily optimize this without breaking any API contract.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -29,10 +63,51 @@ public Goal getGoal()
...
 	}
 
+	public boolean areAggregatesComputed()
+	{
+		return aggregatesComputed;
+	}
+
 	public abstract ZonedDateTime getEndTime();
+
+	public abstract ChronoUnit getTimeUnit();
```
Bert-R: Where do you intend to use this for?

core/src/main/java/nu/yona/server/analysis/service/WeekActivityDTO.java
Bert-R: To be explicit, shouldn't we rename this one in ``WeekActivityDetailDTO``?

appservice/src/main/java/nu/yona/server/analysis/rest/ActivityController.java
Bert-R: Every element in the ``_embedded`` has a self-link, right? That needs to work, so the method needs to exist. Or do you have something else in mind?

PR #141, opened 2016-03-30
YD-213 Prevent information leakage through error codes

Bert-R:
It shouldn't be possible to see whether a mobile number exists, except during the short moment the new device request exists.

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
@@ -34,24 +34,35 @@ class AddDeviceTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(richard)
 	}
 
-	def 'Try get new device request with wrong user secret'()
+	def 'Try get new device request with wrong information'()
 	{
 		given:
 		def userSecret = "unknown secret"
 		User richard = addRichard()
+		User bob = addBob()
 		appService.setNewDeviceRequest(richard.mobileNumber, richard.password, userSecret)
 
 		when:
-		def response = appService.getNewDeviceRequest(richard.mobileNumber, "wrong secret")
+		def responseWrongSecret = appService.getNewDeviceRequest(richard.mobileNumber, "wrong secret")
+		def responseWrongMobileNumber = appService.getNewDeviceRequest("+31690609181", "wrong secret")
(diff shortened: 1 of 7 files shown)


appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
```diff
@@ -61,13 +72,17 @@ class AddDeviceTest extends AbstractAppServiceIntegrationTest
...
 
 		then:
-		response.status == 400
+		responseWrongPassword.status == 400
+		responseWrongPassword.responseData.code == "error.decrypting.data"
 		def getResponseAfter = appService.getNewDeviceRequest(richard.mobileNumber, userSecret)
 		getResponseAfter.status == 200
 		getResponseAfter.responseData.userPassword == richard.password
+		responseWrongMobileNumber.status == 400
+		responseWrongMobileNumber.responseData.code == "error.decrypting.data"
```
aristotelos: Would it not be more logical to have "error.wrong.mobile.number.or.password" in both cases?

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
```diff
@@ -61,13 +72,17 @@ class AddDeviceTest extends AbstractAppServiceIntegrationTest
...
 
 		when:
-		def response = appService.setNewDeviceRequest(richard.mobileNumber, "foo", "Some secret")
+		def responseWrongPassword = appService.setNewDeviceRequest(richard.mobileNumber, "foo", "Some secret")
+		def responseWrongMobileNumber = appService.setNewDeviceRequest("+31690609181", "foo", "Some secret")
 
 		then:
-		response.status == 400
+		responseWrongPassword.status == 400
+		responseWrongPassword.responseData.code == "error.decrypting.data"
```
aristotelos: Would it not be more logical to have "error.wrong.mobile.number.or.password" in both cases?

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
Bert-R: We're simply returning the decryption exception to the client. We could catch that and throw a different exception, but I wonder whether that's worth the effort.

aristotelos: Well, the code now is a bit strange with a passed exception supplier. It becomes easier with just catching an exception.

PR #144, opened 2016-04-02
YD-201 YD-202 activity level of detail

aristotelos:


appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -386,6 +386,38 @@ paths:
           description: Activity of this user.
           schema:
             $ref: '#/definitions/PagedDayActivityOverviews'
+  '/users/{userID}/activity/days/{date}/{goalID}':
```
Bert-R: You earlier said you're not happy with the two consecutive IDs. What about defining the URL this way?
``/users/{userID}/activity/days/{date}/details/{goalID}``.
The additional benefit is that we can define other actions and things on a date too.

aristotelos: Agreed, will update it to that

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -26,17 +29,23 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
...
+		dayActivityForBudgetGoal.date =~ /\d{4}\-\d{2}\-\d{2}/
+		dayActivityForBudgetGoal.timeZoneId == "Europe/Amsterdam"
+		dayActivityForBudgetGoal._links."yona:goal"
+		dayActivityForBudgetGoal._links.self
+
+		//time zone goal should have spread
+		def dayActivityForTimeZoneGoal = dayActivityOverview._embedded."yona:dayActivities".find{ it._links."yona:goal".href == timeZoneGoal.url}
+		dayActivityForTimeZoneGoal
+		dayActivityForTimeZoneGoal.spread
+		dayActivityForTimeZoneGoal.spread.size() <= 96
```
Bert-R: Shouldn't this be exactly 96? We are reporting all 15-minute blocks of the day, right?

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
```diff
@@ -58,13 +64,31 @@ public UUID getGoalID()
...
+	{
+		if (condition)
+		{
+			try
+			{
+				return calculator.call();
+			}
+			catch (Exception e)
+			{
+				throw new RuntimeException(e);
```
Bert-R: Change this into ``throw YonaException.unexpected(e);``.

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
```diff
@@ -58,13 +64,31 @@ public UUID getGoalID()
...
 	}
 
+	@JsonInclude(Include.NON_NULL)
 	public List<Integer> getSpread()
 	{
 		return spread;
 	}
 
-	public int getTotalActivityDurationMinutes()
+	@JsonInclude(Include.NON_NULL)
```
Bert-R: Change this into ``Include.NON_EMPTY``. A collection that is null is against good coding practices.

core/src/main/java/nu/yona/server/analysis/service/WeekActivityDTO.java
```diff
@@ -45,10 +45,14 @@ public static LocalDate parseDate(String iso8601)
...
-	static WeekActivityDTO createInstance(WeekActivity weekActivity)
+	static WeekActivityDTO createInstance(WeekActivity weekActivity, LevelOfDetail levelOfDetail)
 	{
-		return new WeekActivityDTO(weekActivity.getGoal().getID(), weekActivity.getStartTime(), weekActivity.getSpread(),
-				weekActivity.getTotalActivityDurationMinutes(), weekActivity.getDayActivities().stream()
-						.map(dayActivity -> DayActivityDTO.createInstance(dayActivity)).collect(Collectors.toList()));
+		boolean includeDetail = levelOfDetail == LevelOfDetail.WeekDetail;
+		return new WeekActivityDTO(weekActivity.getGoal().getID(), weekActivity.getStartTime(),
+				includeIf(() -> weekActivity.getSpread(), includeDetail),
+				includeIf(() -> weekActivity.getTotalActivityDurationMinutes(), includeDetail),
```
Bert-R: Same here: ``includeDetail ? weekActivity.getTotalActivitDurationMinutes() : Optional.empty()``

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
```diff
@@ -1,27 +1,33 @@
...
+	{
+		WeekOverview, WeekDetail, DayOverview, DayDetail
+	}
+
 	private UUID goalID;
 	private ZonedDateTime startTime;
 
 	private List<Integer> spread;
-	private int totalActivityDurationMinutes;
+	private Integer totalActivityDurationMinutes;
```
Bert-R: Instead of making this null, it's better to use ``Optional<Integer>``.

Bert-R:
Structure is good. Added some coding pattern comments.

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
aristotelos: That doesn't work on the DTO level, because `Optional<Integer>` is rendered differently by Jackson than an `Integer`.

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
aristotelos: Changed that on the `getSpread` method.

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
Bert-R: You don't have to do it there. For Jackson, you can still make it null.

PR #145, opened 2016-04-05
YD-207 Fixed cross-day app activity

aristotelos:
Assumption: activity never crosses 2 days

### DIFF
--- a/core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
+++ b/core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
@@ -1,6 +1,6 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2016 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License, v.
- * 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ * Copyright (c) 2015, 2016 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
  *******************************************************************************/
 package nu.yona.server.analysis.service;
 
@@ -27,7 +27,6 @@ import nu.yona.server.properties.YonaProperties;
 import nu.yona.server.subscriptions.entities.UserAnonymized;
 import nu.yona.server.subscriptions.service.UserAnonymizedDTO;
 import nu.yona.server.subscriptions.service.UserAnonymizedService;
-import nu.yona.server.subscriptions.service.UserService;
 
 @Service
 public class AnalysisEngineService
@@ -39,15 +38,12 @@ public class AnalysisEngineService
 	@Autowired
 	private AnalysisEngineCacheService cacheService;
 	@Autowired
-	private UserService userService;
-	@Autowired
 	private UserAnonymizedService userAnonymizedService;
 	@Autowired
 	private MessageService messageService;
 
-	public void analyze(UUID userID, AppActivityDTO[] appActivities)
+	public void analyze(UUID userAnonymizedID, AppActivityDTO[] appActivities)
 	{
-		UUID userAnonymizedID = userService.getPrivateUser(userID).getPrivateData().getUserAnonymizedID();
 		UserAnonymizedDTO userAnonymized = userAnonymizedService.getUserAnonymized(userAnonymizedID);
 		for (AppActivityDTO appActivity : appActivities)
 		{
@@ -77,15 +73,35 @@ public class AnalysisEngineService
 
 	private void addOrUpdateActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal)
 	{
-		Date minEndTime = new Date(payload.startTime.getTime() - yonaProperties.getAnalysisService().getConflictInterval());
-		DayActivity dayActivity = cacheService.fetchDayActivityForUser(userAnonymized.getID(), matchingGoal.getID());
-		if (dayActivity != null && dayActivity.getStartTime().isBefore(getStartOfDay(payload.startTime, userAnonymized)))
+		ActivityPayload truncatedPayload = payload;
+		ActivityPayload nextDayPayload = null;
+		if (isCrossDayActivity(payload, userAnonymized))
 		{
-			// Last day cached was not today
-			dayActivity = null;
-		}
-		Activity activity = dayActivity == null ? null : dayActivity.getLastActivity();
+			// crossing day border
+			// set current payload to end of day
+			// use new payload for remainder
+			// assumption: activity never crosses 2 days
+			truncatedPayload = new ActivityPayload(payload.url, payload.startTime,
+					Date.from(getEndOfDay(payload.startTime, userAnonymized).toInstant()), payload.application);
 
+			nextDayPayload = new ActivityPayload(payload.url,
+					Date.from(getStartOfDay(payload.endTime, userAnonymized).toInstant()), payload.endTime, payload.application);
+		}
+
+		addOrUpdateDayTruncatedActivity(truncatedPayload, userAnonymized, matchingGoal);
+
+		if (nextDayPayload != null)
+		{
+			// day should be created
+			addActivity(nextDayPayload, userAnonymized, matchingGoal, null);
+		}
+	}
+
+	private void addOrUpdateDayTruncatedActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal)
+	{
+		Date minEndTime = new Date(payload.startTime.getTime() - yonaProperties.getAnalysisService().getConflictInterval());
+		DayActivity dayActivity = getCachedDayActivity(payload, userAnonymized, matchingGoal);
+		Activity activity = dayActivity == null ? null : dayActivity.getLastActivity();
 		if (activity == null || activity.getEndTime().before(minEndTime))
 		{
 			addActivity(payload, userAnonymized, matchingGoal, dayActivity);
@@ -97,6 +113,22 @@ public class AnalysisEngineService
 		{
 			updateActivity(payload, userAnonymized, matchingGoal, dayActivity, activity);
 		}
+	}
+
+	private DayActivity getCachedDayActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal)
+	{
+		DayActivity dayActivity = cacheService.fetchDayActivityForUser(userAnonymized.getID(), matchingGoal.getID());
+		if (dayActivity != null && dayActivity.getStartTime().isBefore(getStartOfDay(payload.startTime, userAnonymized)))
+		{
+			// Last day cached was not today
+			dayActivity = null;
+		}
+		return dayActivity;
+	}
+
+	private boolean isCrossDayActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized)
+	{
+		return getStartOfDay(payload.startTime, userAnonymized).isBefore(getStartOfDay(payload.endTime, userAnonymized));
 	}
 
 	private void addActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal,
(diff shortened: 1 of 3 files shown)


core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -77,15 +73,35 @@ private void analyze(ActivityPayload payload, UserAnonymizedDTO userAnonymized,
...
+		ActivityPayload truncatedPayload = payload;
+		ActivityPayload nextDayPayload = null;
+		if (isCrossDayActivity(payload, userAnonymized))
 		{
-			// Last day cached was not today
-			dayActivity = null;
+			// crossing day border
+			// set current payload to end of day
+			// use new payload for remainder
+			// assumption: activity never crosses 2 days
```
Bert-R: Makes sense to me.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -77,15 +73,35 @@ private void analyze(ActivityPayload payload, UserAnonymizedDTO userAnonymized,
 
 	private void addOrUpdateActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal)
```
Bert-R: To me, this is a lot more readable (without the need for comments):
```
	private void addOrUpdateActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal)
	{
		if (isCrossDayActivity(payload, userAnonymized))
		{
			ActivityPayload truncatedPayload = new ActivityPayload(payload.url, payload.startTime,
					Date.from(getEndOfDay(payload.startTime, userAnonymized).toInstant()), payload.application);

			ActivityPayload nextDayPayload = new ActivityPayload(payload.url,
					Date.from(getStartOfDay(payload.endTime, userAnonymized).toInstant()), payload.endTime, payload.application);

			addOrUpdateDayTruncatedActivity(truncatedPayload, userAnonymized, matchingGoal);
			addOrUpdateDayTruncatedActivity(nextDayPayload, userAnonymized, matchingGoal);
		}
		else
		{
			addOrUpdateDayTruncatedActivity(payload, userAnonymized, matchingGoal);
		}
	}
```

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -77,15 +73,35 @@ private void analyze(ActivityPayload payload, UserAnonymizedDTO userAnonymized,
...
+			// day should be created
+			addActivity(nextDayPayload, userAnonymized, matchingGoal, null);
+		}
+	}
+
+	private void addOrUpdateDayTruncatedActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal)
+	{
+		Date minEndTime = new Date(payload.startTime.getTime() - yonaProperties.getAnalysisService().getConflictInterval());
+		DayActivity dayActivity = getCachedDayActivity(payload, userAnonymized, matchingGoal);
+		Activity activity = dayActivity == null ? null : dayActivity.getLastActivity();
```
Bert-R: Please rename ``activity`` into ``lastRegisteredActivity``.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -39,15 +38,12 @@
...
 	private UserAnonymizedService userAnonymizedService;
 	@Autowired
 	private MessageService messageService;
 
-	public void analyze(UUID userID, AppActivityDTO[] appActivities)
+	public void analyze(UUID userAnonymizedID, AppActivityDTO[] appActivities)
 	{
-		UUID userAnonymizedID = userService.getPrivateUser(userID).getPrivateData().getUserAnonymizedID();
 		UserAnonymizedDTO userAnonymized = userAnonymizedService.getUserAnonymized(userAnonymizedID);
 		for (AppActivityDTO appActivity : appActivities)
```
Bert-R: The logic below assumes that the activities are added in order, oldest first. Instead of relying on the client to send them in that order, I propose to sort them here.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -77,15 +73,35 @@ private void analyze(ActivityPayload payload, UserAnonymizedDTO userAnonymized,
...
+		addOrUpdateDayTruncatedActivity(truncatedPayload, userAnonymized, matchingGoal);
+
+		if (nextDayPayload != null)
+		{
+			// day should be created
+			addActivity(nextDayPayload, userAnonymized, matchingGoal, null);
+		}
+	}
+
+	private void addOrUpdateDayTruncatedActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal)
```
Bert-R: Propose to refactor this method to move the nitty gritty details to separate methods:
```
	private void addOrUpdateDayTruncatedActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal)
	{
		DayActivity dayActivity = getCachedDayActivity(payload, userAnonymized, matchingGoal);
		Activity lastRegisteredActivity = dayActivity == null ? null : dayActivity.getLastActivity();
		if (isInIntervalWithLastRegisteredActivity(lastRegisteredActivity, payload))
		{
			addActivity(payload, userAnonymized, matchingGoal, dayActivity);
		}
		else if (isBeyondSkipWindowAfterLastRegisteredActivity(lastRegisteredActivity, payload))
		{
			updateActivity(payload, userAnonymized, matchingGoal, dayActivity, lastRegisteredActivity);
		}
	}

	private boolean isBeyondSkipWindowAfterLastRegisteredActivity(Activity lastRegisteredActivity, ActivityPayload payload)
	{
		return payload.endTime.getTime() - lastRegisteredActivity.getEndTime().getTime() >= yonaProperties.getAnalysisService()
				.getUpdateSkipWindow() || payload.startTime.before(lastRegisteredActivity.getStartTime());
	}

	private boolean isInIntervalWithLastRegisteredActivity(Activity lastRegisteredActivity, ActivityPayload payload)
	{
		Date minEndTime = new Date(payload.startTime.getTime() - yonaProperties.getAnalysisService().getConflictInterval());
		return lastRegisteredActivity == null || lastRegisteredActivity.getEndTime().before(minEndTime);
	}
```

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
Bert-R: Even then, I'm afraid the current code doesn't accept app activity reports that are reported way after they happened, while network (Smoothwall) activities have bene reported in that same period. What do you think?

Bert-R:
Added a few comments.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
aristotelos: Yes, I agree. If the posted app activity is in a day before the current cached day, the day should be fetched from the database.

Bert-R: To me, it doesn't need to be fixed as part of this PR. Logging a ticket for it is fine too.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
aristotelos: Thanks for the refactoring! Looks much better indeed

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
aristotelos: Has good ideas but the code is not correct. I'll try to extract methods with similar names for the complex conditions, that indeed helps to understand (and fix!) the algorithm.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
aristotelos: Applied

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -73,57 +76,114 @@ private void analyze(ActivityPayload payload, UserAnonymizedDTO userAnonymized,
...
 		}
 	}
 
 	private void addOrUpdateDayTruncatedActivity(ActivityPayload payload, UserAnonymizedDTO userAnonymized, Goal matchingGoal)
 	{
-		Date minEndTime = new Date(payload.startTime.getTime() - yonaProperties.getAnalysisService().getConflictInterval());
-		DayActivity dayActivity = getCachedDayActivity(payload, userAnonymized, matchingGoal);
-		Activity activity = dayActivity == null ? null : dayActivity.getLastActivity();
-		if (activity == null || activity.getEndTime().before(minEndTime))
+		DayActivityCacheResult dayActivity = getRegisteredDayActivity(payload, userAnonymized, matchingGoal);
```
Bert-R: Rename ``dayActivity`` into ``lastRegisteredActivity``, also on line 107 below, to fix compilation errors.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -73,57 +76,114 @@ private void analyze(ActivityPayload payload, UserAnonymizedDTO userAnonymized,
...
 		}
-		// Update message only if it is within five seconds to avoid unnecessary cache flushes.
-		// Or if the start time is earlier than the existing start time (this can happen with app activity, see below).
-		else if (payload.endTime.getTime() - activity.getEndTime().getTime() >= yonaProperties.getAnalysisService()
-				.getUpdateSkipWindow() || payload.startTime.before(activity.getStartTime()))
+	}
+
+	private boolean isBeyondSkipWindowAfterLastRegisteredActivity(ActivityPayload payload, Activity lastRegisteredActivity)
+	{
+		return payload.endTime.getTime() - lastRegisteredActivity.getEndTime().getTime() >= yonaProperties.getAnalysisService()
```
Bert-R: Shouldn't you compare the start time of the new activity with the end time of the last registered one?

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -73,57 +76,114 @@ private void analyze(ActivityPayload payload, UserAnonymizedDTO userAnonymized,
...
+	private boolean precedesLastRegisteredActivity(ActivityPayload payload, Activity lastRegisteredActivity)
 	{
-		DayActivity dayActivity = cacheService.fetchDayActivityForUser(userAnonymized.getID(), matchingGoal.getID());
-		if (dayActivity != null && dayActivity.getStartTime().isBefore(getStartOfDay(payload.startTime, userAnonymized)))
+		return payload.startTime.before(lastRegisteredActivity.getStartTime());
+	}
+
+	private boolean isBeyondCombineIntervalWithLastRegisteredActivity(ActivityPayload payload, Activity lastRegisteredActivity)
+	{
+		Date intervalEndTime = new Date(
```
Bert-R: What about this?
```
ZonedDateTime intervalEndTime = lastRegisteredActivity.getEndTime().plus(yonaProperties.getAnalysisService().getConflictInterval(), ChronoUnit.MILLIS);
```

Bert-R:
Added a few comments, including one about a build failure.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
aristotelos: In my opinion doesn't really reduce complexity. Moreover, it has to be compared with `Date` anyway.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
aristotelos: No. If the end time of the new activity is 5 minutes later than the existing activity, the existing activity should be updated.

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
aristotelos: Fixed now, added unit test too

core/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
Bert-R: It prevents using the old ``Date`` class and it's more explicit about what happens.

aristotelos: The `Activity.getEndTime()` also returns a `Date` (so your line does not compile), and the `ActivityPayload.startTime` is a `Date`. I propose to create a refactoring task for it to move to `ZonedDateTime` altogether.

Bert-R: Fine with me.

PR #147, opened 2016-04-07
YD-217 Confirmation code enhancements

Bert-R:


core/src/main/java/nu/yona/server/exceptions/UserOverwriteConfirmationException.java
```diff
@@ -13,24 +10,46 @@
 public class UserOverwriteConfirmationException extends YonaException
```
aristotelos: Can't we make a base class for `UserOverwriteConfirmationException` and `MobileNumberConfirmationException `?

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -221,19 +209,16 @@ private void deleteExistingUserToOverwriteIt(String mobileNumber, String overwri
...
-		{
-			throw MobileNumberConfirmationException.confirmationCodeNotSet(userEntity.getMobileNumber());
-		}
-
-		if (!userEntity.getMobileNumberConfirmationCode().equals(code))
-		{
-			throw MobileNumberConfirmationException.confirmationCodeMismatch(userEntity.getMobileNumber(), code);
-		}
+		verifyConfirmationCode(userEntity, confirmationCode, userProvidedConfirmationCode,
+				() -> MobileNumberConfirmationException.confirmationCodeNotSet(userEntity.getMobileNumber()),
```
aristotelos: If we have one generic exception with a placeholder, that would simplify the code I guess. So just a `ConfirmationCodeException` which has a parameter `purposeOfUse` or something like that.

appservice/src/main/java/nu/yona/server/BruteForceAttemptService.java
```diff
@@ -1,136 +0,0 @@
-/*******************************************************************************
```
aristotelos: Hmm... Now I wonder what happened to the brute force checking on new device request? Is there any brute force checking done there still?

aristotelos:
Nice work, added some comments

appservice/src/main/java/nu/yona/server/BruteForceAttemptService.java
Bert-R: It was used at two places: overwrite account and confirm number. Both are now replaced by the ``ConfirmationCode`` entity.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
Bert-R: That I don't like, but a generic base class is a good idea.

Bert-R:
Added a common base class

appservice/src/main/java/nu/yona/server/BruteForceAttemptService.java
aristotelos: OK, it is not necessary at device requests because these use a user ID and should be short-living.

Bert-R: Indeed.

PR #148, opened 2016-04-08
YD-218 Enhancements to new device request

Bert-R:
* Renamed "user secret" into "new device request password"
* It's now passed as a header. Passwords should never go in the URL.
* Renamed "user password" into "yona password", in accordance with the header
* Removed duplicate header for PUT of new device request in Swagger spec.


aristotelos:
This adds some clarity indeed.

PR #149, opened 2016-04-12
YD-208 show zero activity

aristotelos:
Adding "no activity" records using goal creation time


core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
```diff
@@ -212,4 +209,9 @@ public static DayActivity createInstance(UserAnonymized userAnonymized, Goal goa
 		return new DayActivity(UUID.randomUUID(), userAnonymized, goal, startOfDay, new ArrayList<Activity>(),
 				new ArrayList<Integer>(IntervalActivity.SPREAD_COUNT), 0, true, false);
 	}
+
+	public static DayActivity createInstanceZeroActivity(UserAnonymized userAnonymized, Goal goal, ZonedDateTime startOfDay)
```
Bert-R: Isn't it better to create talk about inactivity instead of ``zeroActivity``? Inactivity is just a plain English word, understood by everyone.

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
```diff
@@ -30,48 +40,172 @@
...
+		}
+		ZonedDateTime dateAtStartOfDay = date.atStartOfDay(ZoneId.of(userAnonymized.getTimeZoneId()));
+		if (!goal.get().getCreationTime().isBefore(dateAtStartOfDay.plus(1, timeUnit)))
+		{
+			throw ActivityServiceException.activityDateGoalMismatch(userID, date, goalID);
+		}
+		return zeroActivityEntitySupplier.apply(goal.get(), dateAtStartOfDay);
+	}
+
+	class StartEndDate
```
Bert-R: Please rename in to ``Interval``, as that's really what it is. Joda Time has a similar type ([Interval](http://joda-time.sourceforge.net/apidocs/org/joda/time/Interval.html)). Java unfortunately doesn't have it.

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
```diff
@@ -30,48 +40,172 @@
...
+				return currentDayDate.minusDays(currentDayDate.getDayOfWeek().getValue());
+		}
+	}
+
+	private LocalDate getCurrentDayDate(UserAnonymizedDTO userAnonymized)
+	{
+		return LocalDate.now(ZoneId.of(userAnonymized.getTimeZoneId()));
+	}
+
+	private <T extends IntervalActivity> void addMissingZeroActivity(Map<LocalDate, Set<T>> activityEntitiesByDate,
```
Bert-R: Rename into ``addInactivity``.

Bert-R: Nice generic implementation!

core/src/main/resources/messages/messages.properties
```diff
@@ -57,6 +57,8 @@ error.buddy.only.twoway.buddies.allowed=Only Two-way buddy relationships are all
 error.buddy.requesting.user.buddy.is.null=The requesting user cannot be null
 error.buddy.accepting.user.is.null=The accepting user cannot be null
 
+error.activity.date.goal.mismatch=The activity date {0} and goal with ID ''{1}'' do not match
```
Bert-R: This message is unclear to me. Is this what you mean?

> At {1}, user {0} did not have a goal with ID {2}

Bert-R:
Generally looks good. Added a few comments.

core/src/main/resources/messages/messages.properties
aristotelos: Indeed

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
aristotelos: I don't think that type is the same. It forgets about start and end date and just remembers the amount of time in between.

aristotelos: Same with `Period` time in Java (https://docs.oracle.com/javase/tutorial/datetime/iso/period.html).

Bert-R: Yes, you're right.
But still... ``StartEndDate`` is a very confusing thing. If you have anything better than ``Interval``, then please suggest.

aristotelos: No problem, Interval sounds OK to me.

aristotelos:
Reworked after comments

aristotelos:
Synced with master and checked tests

PR #150, opened 2016-04-15
YD-216 Implemented pin reset request API

Bert-R:
The batch job to set the pin after a configurable delay will be added through a separate ticket (YD-220). This way, the app development can move on, with no dependency on the batch job.


appservice/src/intTest/groovy/nu/yona/server/PinResetRequestTest.groovy
```diff
@@ -0,0 +1,153 @@
...
+		richard = appService.getUser(appService.&assertUserGetResponseDetailsPublicDataAndVpnProfile, richard.url, true, richard.password)
+
+		when:
+		def response = appService.yonaServer.postJson(richard.clearPinResetUrl, [:], ["Yona-Password" : richard.password])
+
+		then:
+		response.status == 200
+		User  richardAfterGet = appService.getUser(appService.&assertUserGetResponseDetailsPublicDataAndVpnProfile, richard.url, true, richard.password)
+		richardAfterGet.pinResetRequestUrl
+		!richardAfterGet.verifyPinResetUrl
```
aristotelos: Please add a line to test that the original `richard.verifyPinResetUrl` returns 400 now.

appservice/src/intTest/groovy/nu/yona/server/PinResetRequestTest.groovy
```diff
@@ -0,0 +1,153 @@
...
+		appService.yonaServer.postJson(richard.pinResetRequestUrl, [:], ["Yona-Password" : richard.password])
+		richard = appService.getUser(appService.&assertUserGetResponseDetailsPublicDataAndVpnProfile, richard.url, true, richard.password)
+
+		when:
+		def response = appService.yonaServer.postJson(richard.verifyPinResetUrl, """{"code" : "1234"}""", ["Yona-Password" : richard.password])
+
+		then:
+		response.status == 200
+		User  richardAfterGet = appService.getUser(appService.&assertUserGetResponseDetailsPublicDataAndVpnProfile, richard.url, true, richard.password)
+		!richardAfterGet.pinResetRequestUrl
```
aristotelos: Please add a test that a post to the original `richard.pinResetRequestUrl` gives an error that a PIN has already been requested for this or another device.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -372,6 +378,7 @@ public UserResource toResource(UserDTO user)
 					addWeekActivityOverviewsLink(userResource);
 					addNewDeviceRequestLink(userResource);
 					addAppActivityLink(userResource);
+					pinResetRequestController.addLinks(userResource);
```
aristotelos: This way does not follow the pattern. Other controllers are just exposing internal/public methods to get the link, so for consistency, should we not follow the same pattern here?

core/src/main/java/nu/yona/server/subscriptions/service/PinResetRequestService.java
```diff
@@ -0,0 +1,81 @@
...
+	@Autowired
+	private UserService userService;
+
+	@Autowired
+	private YonaProperties yonaProperties;
+
+	@Transactional
+	public void requestPinReset(UUID userID)
+	{
+		ConfirmationCode confirmationCode = createConfirmationCode();
```
aristotelos: This should have a check that there does not exist a pin reset request yet (see remark about test). Otherwise it is stil possible to bruteforce the code the other way around.

appservice/src/main/java/nu/yona/server/subscriptions/rest/PinResetRequestController.java
```diff
@@ -0,0 +1,142 @@
...
+			return new ResponseEntity<ErrorResponseDTO>(responseMessage, e.getStatusCode());
+		}
+		return globalExceptionMapping.handleYonaException(e);
+	}
+
+	public void addLinks(UserResource userResource)
+	{
+		ConfirmationCode confirmationCode = userService.getUserByID(userResource.getContent().getID())
+				.getPinResetConfirmationCode();
+		if (confirmationCode == null || pinResetRequestService.isExpired(confirmationCode))
```
aristotelos: Will the batch script also remove expired confirmation codes?

aristotelos:
Looks great, added a few comments

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
Bert-R: The reason for doing it this way is that otherwise the knowledge about which link type are appropriate in what situation would be spread across the user controller and the pin reset request controller. For the other links, the condition (mobile number confirmed) is know to the user controller.

core/src/main/java/nu/yona/server/subscriptions/service/PinResetRequestService.java
Bert-R: That brute force approach doesn't work. You'll always get a new pin, so you need to start all over again.

appservice/src/main/java/nu/yona/server/subscriptions/rest/PinResetRequestController.java
Bert-R: Yes.

appservice/src/intTest/groovy/nu/yona/server/PinResetRequestTest.groovy
Bert-R: Done.

appservice/src/intTest/groovy/nu/yona/server/PinResetRequestTest.groovy
Bert-R: See comment below.

Bert-R:
Pushed an update that adds the requested extra test and brings a few more improvements.

core/src/main/java/nu/yona/server/subscriptions/service/PinResetRequestService.java
aristotelos: Hmm... If you can request a random number infinite times, I guess the chance that it equals 00000 is 1.
Because you anyway need the user password this is not the biggest thing to worry about though.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
aristotelos: OK, indeed this seems necessary. But there is maybe a way to avoid having a `null` PinResetRequestController` in the constructor? Maybe an overload for the include private data case and an overload for the public case?

Bert-R: You're right. That I should have done immediately.
It's done now.

PR #153, opened 2016-04-16
YD-221 Made activity category names translatable

Bert-R:
Along with that:
* Added GoalChangeMessage to the Swagger spec
* Resolved TODO in Swagger spec: activity category link for goal conflict messageAggregation
* Relationship with activity category is now by ID/URL


appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1022,13 +1030,12 @@ definitions:
         - BudgetGoal
         - TimeZoneGoal
         description: The type of goal
-      activityCategoryName:
+      activityCategoryUrl:
```
aristotelos: Why not put this in the `_links`?

core/src/main/java/nu/yona/server/goals/service/GoalServiceException.java
```diff
@@ -25,9 +22,9 @@ public static GoalServiceException goalNotFoundById(UUID userID, UUID goalID)
 		return new GoalServiceException("error.goal.not.found.id", userID, goalID);
 	}
 
-	public static GoalServiceException cannotAddSecondGoalOnActivityCategory(String activityCategoryName)
+	public static GoalServiceException cannotAddSecondGoalOnActivityCategory(UUID activityCategoryID)
 	{
-		return new GoalServiceException("error.goal.cannot.add.second.on.activity.category", activityCategoryName);
+		return new GoalServiceException("error.goal.cannot.add.second.on.activity.category", activityCategoryID);
```
aristotelos: Please change the corresponding messages.

core/src/main/java/nu/yona/server/goals/service/BudgetGoalDTO.java
```diff
@@ -22,15 +19,15 @@
 	private final int maxDurationMinutes;
 
 	@JsonCreator
-	public BudgetGoalDTO(@JsonProperty("activityCategoryName") String activityCategoryName,
-			@JsonProperty("maxDurationMinutes") int maxDurationMinutes)
+	public BudgetGoalDTO(@JsonProperty(required = true, value = "activityCategoryUrl") String activityCategoryUrl,
+			@JsonProperty(required = true, value = "maxDurationMinutes") int maxDurationMinutes)
 	{
-		this(null, activityCategoryName, maxDurationMinutes, false /* ignored */);
+		this(null, determineActivityCategoryID(activityCategoryUrl), maxDurationMinutes, false /* ignored */);
```
aristotelos: Maybe we can move this call to `determineActivityCategoryID` to a `GoalDTO` constructor such that it does not have to be repeated for every goal DTO.

core/src/main/java/nu/yona/server/goals/service/GoalDTO.java
```diff
@@ -56,6 +54,11 @@ public boolean isMandatory()
 
 	public abstract Goal createGoalEntity();
 
+	protected static UUID determineActivityCategoryID(String activityCategoryUrl)
+	{
+		return UUID.fromString(activityCategoryUrl.substring(activityCategoryUrl.lastIndexOf('/') + 1));
```
aristotelos: Officially, we should ask this to the controller... Or rather the controller should already substitute the URL with an ID in the goal DTO when executing the POST.

aristotelos:
Pfew, quite a lot of changes. Added a few comments.
By the way, is the app going to translate the activity category? How long is it allowed to cache those activity category translations?

aristotelos:
I'm sorry, I read that the translation is done on the server. So then when the user changes its language, the cached activity categories should be refreshed.

core/src/main/java/nu/yona/server/goals/service/GoalServiceException.java
Bert-R: The message was still holding good. However, it's pointless to send an activity category ID back to the user, so I've changed it back to the activity category name.

Bert-R:
> I'm sorry, I read that the translation is done on the server. So then when the user changes its language, the cached activity categories should be refreshed.

Indeed. Fetching these when opening the app sounds like the best approach to me.

core/src/main/java/nu/yona/server/goals/service/GoalDTO.java
Bert-R: That's not possible, as the DTOs are deserialized before the controller is called.

core/src/main/java/nu/yona/server/goals/service/BudgetGoalDTO.java
Bert-R: OK

aristotelos:
Maybe add that to the Swagger spec, that the name is a translated name and should be refreshed when switching the user language or with every startup?

core/src/main/java/nu/yona/server/goals/service/GoalDTO.java
aristotelos: But after deserialization, you can also parse the `activityCategoryUrl` at the controller and after that set an `activityCategoryID` on the same DTO.

Bert-R: Sure, but then the DTO needs two properties, one for the URL and one for the ID. This way, it's parsed at the time it's received, so you don't need to keep it.

PR #154, opened 2016-04-18
YD-222 The activity category URL is now in _links

Bert-R:
This by nature creates the code structure as requested in the previous review.


Bert-R:
As I want to give the new build to Mobiquity today, I've merged this PR myself. Please still do the review. If changes are necessary, these will be done through another PR.

PR #156, opened 2016-04-20
YD-209 date at activity overview

aristotelos:
- Showing date at overview level when possible
- Using day of week names


aristotelos:
Hmm, ik kan de integratietesten niet draaien, geven exception: "Caused by: groovy.lang.GroovyRuntimeException: Conflicting module versions. Module [groovy-all is loaded in version 2.4.5 and you are trying to load version 2.4.4"

Bert-R:
Dat is vreemd. In Eclipse of vanuit Gradle? In Eclipse heb ik zulke dingen
vaker gezien, in Gradle niet. Als er iets in Gradle mis gaat zou je een
keer gradlew dependencies kunnen draaien, dan krijg je alle
afhankelijkheden en de triggers om ze te upgraden naar hogere versies.

Ik heb de branch hier uitgecheckt en doe nu een test run. Ik denk niet dat
hij op tijd klaar is voor ik naar kantoor moet, maar ik probeer het in elk
geval even. Lukt het nu niet, dan laat ik het aan het begin van de avond
even weten.

Groetjes,
Bert

aristotelos:
Nog iets anders: ik kreeg unit test fouten omdat mijn laptop geïnstalleerd
is met en-GB ipv en-US en omdat dan de activity category getName() een
assertionerror geeft. Krijg nu nog steeds een integratietest fout op
hetzelfde...

2016-04-21 7:47 GMT+02:00 Bert Roos <notifications@github.com>:

Bert-R:
Hoi Aris,

Dat moet dan worden aangepast, want we moeten niet afhankelijk zijn van de
locale van de machine.
Bij mij draaien de tests (in Gradle). Hij is bij 77.
Eventueel kan je Gradle alle dependencies laten refreshen
(--refresh-dependencies). Verder de dependencies doorkijken met gradlew
dependencies.
Ik vertrek nu.

Groetjes,
Bert

2016-04-21 7:49 GMT+02:00 aristotelos <notifications@github.com>:

aristotelos:
In deze PR zit in ieder geval een fix voor een gedeelte (de unit tests), de rest kun je misschien zelf fixen als vervolg van YD-221?

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1545,25 +1545,41 @@ definitions:
         allOf:
           - $ref: '#/definitions/Curies'
           - $ref: '#/definitions/SelfLink'
+      date:
```
Bert-R: Why do you have a date at this level? Ever week has it's date already, right?

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1631,20 +1647,26 @@ definitions:
...
```
Bert-R: Same here: every day has a date, so why do you add a date for the set of days?

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1545,25 +1545,41 @@ definitions:
...
+      - date
   WeekActivity:
     allOf:
       - $ref: '#/definitions/IntervalActivity'
     properties:
       date:
         type: string
         example: 2016-W2
-        description: The week date in ISO 8601 format (YYYY-'W'w)
+        description: The week date in ISO 8601 format (YYYY-'W'w). Only included in week detail
```
Bert-R: Why is this? Now the app needs to do math to find out that the "batch" starts in week 5 and this is the 3 item in the batch, so this is week 8.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1631,20 +1647,26 @@ definitions:
...
+      - date
   DayActivity:
     allOf:
       - $ref: '#/definitions/IntervalActivity'
     properties:
       date:
         type: string
         example: 2016-03-16
-        description: The day date in ISO 8601 format (yyyy-MM-dd)
+        description: The day date in ISO 8601 format (yyyy-MM-dd). Only included in day detail
```
Bert-R: The WeekActivity question applies here too.

Bert-R:
Met "de rest" bedoel je het probleem net en-GB?

Groetjes,
Bert

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
```diff
@@ -102,7 +103,31 @@ public static WeekActivity createInstance(UserAnonymized userAnonymized, Goal go
...
+			{
+				result.addDayActivity(DayActivity.createInstanceInactivity(userAnonymized, goal, startOfDay));
+			}
+		}
 		return result;
 	}
+
+	public static boolean wasGoalActiveAtDay(Goal goal, ZonedDateTime startOfDay)
+	{
+		return goal.getCreationTime().isBefore(startOfDay.plusDays(1));
```
Bert-R: This now handles goals that are defined later in time. Shouldn't this also deal with goals that are deleted some point in time?

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
```diff
@@ -102,7 +103,31 @@ public static WeekActivity createInstance(UserAnonymized userAnonymized, Goal go
...
+			}
+			if (wasGoalActiveAtDay(goal, startOfDay))
+			{
+				result.addDayActivity(DayActivity.createInstanceInactivity(userAnonymized, goal, startOfDay));
+			}
+		}
 		return result;
 	}
+
+	public static boolean wasGoalActiveAtDay(Goal goal, ZonedDateTime startOfDay)
```
Bert-R: If you implement ``wasActive(Goal goal, ZonedDateTime startOfDay)`` on ``Goal``, then you can also reuse it from ``ActivityService``, see later comment.

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
```diff
@@ -129,25 +131,30 @@ private LocalDate getCurrentDayDate(UserAnonymizedDTO userAnonymized)
...
 		Set<Goal> activeGoals = userAnonymized.getGoals().stream()
-				.filter(g -> g.getCreationTime().isBefore(dateAtStartOfDay.plus(1, timeUnit))).collect(Collectors.toSet());
+				.filter(g -> wasGoalActiveAtInterval(g, dateAtStartOfInterval, timeUnit)).collect(Collectors.toSet());
 		return activeGoals;
 	}
 
-	private <T extends IntervalActivity> void addMissingInactivity(Goal activeGoal, ZonedDateTime dateAtStartOfDay,
+	private boolean wasGoalActiveAtInterval(Goal goal, ZonedDateTime dateAtStartOfInterval, ChronoUnit timeUnit)
+	{
+		return goal.getCreationTime().isBefore(dateAtStartOfInterval.plus(1, timeUnit));
```
Bert-R: This method can call the ``Goal.wasActive`` that I proposed above.

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
```diff
@@ -129,25 +131,30 @@ private LocalDate getCurrentDayDate(UserAnonymizedDTO userAnonymized)
...
+	private Set<Goal> getActiveGoals(UserAnonymizedDTO userAnonymized, ZonedDateTime dateAtStartOfInterval, ChronoUnit timeUnit)
 	{
 		Set<Goal> activeGoals = userAnonymized.getGoals().stream()
-				.filter(g -> g.getCreationTime().isBefore(dateAtStartOfDay.plus(1, timeUnit))).collect(Collectors.toSet());
+				.filter(g -> wasGoalActiveAtInterval(g, dateAtStartOfInterval, timeUnit)).collect(Collectors.toSet());
 		return activeGoals;
 	}
 
-	private <T extends IntervalActivity> void addMissingInactivity(Goal activeGoal, ZonedDateTime dateAtStartOfDay,
+	private boolean wasGoalActiveAtInterval(Goal goal, ZonedDateTime dateAtStartOfInterval, ChronoUnit timeUnit)
```
Bert-R: Can you make clear from the code why you are adding one `timeunit` to the give date?

Bert-R: Please rename into ``wasGoalActiveAtStartOfInterval`` (unless my earlier comment leads to a name like ``wasGoalActiveAtEndOfInterval`` :) )

Bert-R:
I've added a few comments and pushed a different fix for the locale issue. Don't forget to do a pull.

The main comments are about the way you expose the date in the resources. I don't understand why you do that. I only see added complexity, client-side and server-side.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
aristotelos: The reason for my changes is:
- In the design [http://wiki.yona.nu/display/DEV/Design+change+to+capture+app+activity] you mentioned the date a level higher, so day X [day for goal G1, day for goal G2] and week Z [week for goal G1, week for goal G2]. The date was mentioned at X and Z and not a level deeper.
- A reason is just efficiency: If you mention the date at a higher level, you do not have to mention it _n_ times at a lower level.
- Another reason is that the structure is more logical, the app builders would expect a date at the highest level instead of the same date being repeated at the children.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
aristotelos: No, every  week activity for a specific goal (`WeekActivity`) included in a week overview (`WeekActivityOverview`) has exactly the same date.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
aristotelos: Same answer: 
- In day overview, every `DayActivity` has the same date as its parent `DayActivityOverview`, so it is not necessary to include here. 
- In week activity/week activity overview, I could add the date if you like. However in the original design the date was not displayed, only the day of the week. There is a self link already to get the date if necessary, so I would propose to not include it.

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
aristotelos: Currently the implementation of goal delete also deletes all activity, so this is not necessary. Moreover, the batch script to add inactivity will have done its task every day.

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
aristotelos: OK, will do that.

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
aristotelos: Indeed the name should be that the goal was active before the end of the interval. If you add a goal halfway the day, you would like to see inactivity for that day. Same for week.

aristotelos:
Added comments to explain. Maybe we should have a call?

aristotelos:
Inderdaad bedoelde ik met "de rest" de overblijvende problemen met en-GB
integratietesten.

2016-04-23 9:37 GMT+02:00 Bert Roos <notifications@github.com>:

appservice/src/main/resources/static/swagger/swagger-spec.yaml
Bert-R: It was a mistake on my side. I misinterpreted the structure.

Bert-R:
It's clear to me. We don't need a call. I just misunderstood it.

aristotelos:
Rework has been done

PR #160, opened 2016-04-26
YD-159 Add creationTime to goal DTO

aristotelos:
The property was already added earlier for inactivity support.
One question: Should we also refactor Message.creationTime to ZonedDateTime? That would be more consistent. Or the other way around, expose a Date here?

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
@@ -51,10 +51,12 @@ class EditGoalsTest extends AbstractAppServiceIntegrationTest
 		gamblingGoals.size() == 1
 		gamblingGoals[0]."@type" == "BudgetGoal"
 		!gamblingGoals[0]._links.edit //mandatory goal
+		gamblingGoals[0].creationTime =~ /\d{4}\-\d{2}\-\d{2}T\d{2}\:\d{2}\:\d{2}\.\d{3}\+\d{2}\:\d{2}/
 		def newsGoals = filterGoals(response, NEWS_ACT_CAT_URL)
 		newsGoals.size() == 1
 		newsGoals[0]."@type" == "BudgetGoal"
 		newsGoals[0]._links.edit.href
+		newsGoals[0].creationTime =~ /\d{4}\-\d{2}\-\d{2}T\d{2}\:\d{2}\:\d{2}\.\d{3}\+\d{2}\:\d{2}/
 	}
 
 	def 'Add budget goal'()
--- a/appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
@@ -10,6 +10,8 @@ import groovy.json.*
 import nu.yona.server.test.BudgetGoal
 import nu.yona.server.test.Goal
 import nu.yona.server.test.TimeZoneGoal
+import java.time.ZonedDateTime
+import java.time.format.DateTimeFormatter
 
 class EditGoalsTest extends AbstractAppServiceIntegrationTest
 {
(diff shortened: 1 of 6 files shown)


appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
```diff
@@ -51,10 +51,12 @@ class EditGoalsTest extends AbstractAppServiceIntegrationTest
 		gamblingGoals.size() == 1
 		gamblingGoals[0]."@type" == "BudgetGoal"
 		!gamblingGoals[0]._links.edit //mandatory goal
+		gamblingGoals[0].creationTime =~ /\d{4}\-\d{2}\-\d{2}T\d{2}\:\d{2}\:\d{2}\.\d{3}\+\d{2}\:\d{2}/
```
Bert-R: Can't you add an assert that the creation time is equal to the time the user was created?

appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
```diff
@@ -51,10 +51,12 @@ class EditGoalsTest extends AbstractAppServiceIntegrationTest
...
 		def newsGoals = filterGoals(response, NEWS_ACT_CAT_URL)
 		newsGoals.size() == 1
 		newsGoals[0]."@type" == "BudgetGoal"
 		newsGoals[0]._links.edit.href
+		newsGoals[0].creationTime =~ /\d{4}\-\d{2}\-\d{2}T\d{2}\:\d{2}\:\d{2}\.\d{3}\+\d{2}\:\d{2}/
```
Bert-R: Similar here, but for the time the goal was created.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1201,9 +1201,15 @@ definitions:
           - $ref: '#/definitions/SelfLink'
           - $ref: '#/definitions/EditLink'
           - $ref: '#/definitions/ActivityCategoryLink'
+      creationTime:
+        type: string
+        format: dateTime
+        description: The creation time of this goal
+        example: 2016-04-26T20:08:55.365+02:00
```
Bert-R: Other examples use this time stamp sample format: ``2016-02-31T21:28:02.290+0000`` (without the colon in the middle). Pleas use that here too.

Bert-R: I just now see in the test that the serialization format also differs. Please update that too.

core/src/main/java/nu/yona/server/goals/service/GoalDTO.java
```diff
@@ -45,6 +48,11 @@ public UUID getID()
 		return id;
 	}
 
+	public ZonedDateTime getCreationTime()
```
Bert-R: To change the serialization format, you will need to add an annotation. In ``AppActivityDTO``, I used the following for deserialization:
```
@JsonFormat(pattern = **"yyyy-MM-dd'T'HH:mm:ss.SSSZ")
```

appservice/src/main/resources/static/swagger/swagger-spec.yaml
aristotelos: How should I do that? Is it a Jackson setting? The type is different: the other places serialize `Date`, this place serializes `ZonedDateTime`...

Bert-R:
> One question: Should we also refactor Message.creationTime to ZonedDateTime? That would be more consistent. Or the other way around, expose a Date here?

The direction we evolve in is to use ``ZonedDateTime``, so using that here is appropriate. We need to get rid of all occurrences of ``Date`` some day, not necessarily now. For as long as the external API remains te same, we can stick with ``Date`` till we find time.

Bert-R:
I have added a few comments related to the date format and checking.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
Bert-R: See comment on ``GoalDTO``.

PR #164, opened 2016-04-27
YD-232 Correct mail templates

Bert-R:
Replaced @ with $


aristotelos:
Good fix, no comments

PR #165, opened 2016-04-27
YD-233 Standard temp password in test config

Bert-R:
* Added the tempPassword query parameter to the Swagger spec for GET and PUT /users/{id}
* Removed property userCreatedInviteURL from BuddyDTO
* Now generates password abcd in case e-mail is disabled
* The test now composes the invitation URL from the self-link of the buddy user and the abcd password.


aristotelos:
Reviewed, no comments

PR #167, opened 2016-05-03
YD-210 Extended activity tests, fixed two bugs

aristotelos:


core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
```diff
@@ -104,22 +104,33 @@ public static WeekActivity createInstance(UserAnonymized userAnonymized, Goal go
...
-		// so for consistency it is important that the batch script adding inactivity does so
-		for (int i = 0; i < 7; i++)
+		result.addInactivityDays();
+		return result;
+	}
+
+	public void addInactivityDays() 
+	{
+		// if the batch job has already run, skip
+		if(dayActivities.size() < 7)
```
Bert-R: Can you pull this condition out of this method, to the caller? That way, ``addInactivityDays`` truly adds inactivity days. Otherwise rename this method into something like ``ensureFullWeek``. If you opt for the latter, switch the condition and make it return immediately when the week is already full. That way, you don't need to nest the body so deeply.

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
```diff
@@ -104,22 +104,33 @@ public static WeekActivity createInstance(UserAnonymized userAnonymized, Goal go
...
-			{
-				result.addDayActivity(DayActivity.createInstanceInactivity(userAnonymized, goal, startOfDay));
+				ZonedDateTime startOfDay = getStartTime().plusDays(i);
+				if (isInFuture(startOfDay, getStartTime().getZone()))
+				{
+					break;
+				}
+				if (getGoal().wasActiveAtInterval(startOfDay, ChronoUnit.DAYS))
+				{
+					if(!dayActivities.stream().anyMatch(dayActivity -> dayActivity.getDate().getDayOfWeek().equals(startOfDay.getDayOfWeek())))
```
Bert-R: You're now here testing whether the day is already available in the list, right? You're maintaining the day activities in a list, so wouldn't it be a lot simpler of position 0 of the list contains Sunday, position 1 contains Monday, etc.? The you could put a marker on ``WeekActivity`` to indicate the week is completely filled, as an alternative to the ``if(dayActivities.size() < 7)`` test, and for the rest simply use the week day as index in the list.

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
Bert-R: It looks like you switched off the standard code formatter: we normally have a space right after the ``if``. I see that a few lines below as well.

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
```diff
@@ -145,17 +149,22 @@ private LocalDate getCurrentDayDate(UserAnonymizedDTO userAnonymized)
...
 	private <T extends IntervalActivity> void addMissingInactivity(Goal activeGoal, ZonedDateTime dateAtStartOfInterval,
 			Set<T> activityEntitiesAtDate, UserAnonymizedDTO userAnonymized,
-			BiFunction<Goal, ZonedDateTime, T> inactivityEntitySupplier)
+			BiFunction<Goal, ZonedDateTime, T> inactivityEntitySupplier, Consumer<T> existingEntityInactivityCompletor)
 	{
-		if (!containsActivityForGoal(activityEntitiesAtDate, activeGoal))
+		Optional<T> activityForGoal = getActivityForGoal(activityEntitiesAtDate, activeGoal);
+		if (activityForGoal.isPresent())
+		{
+			existingEntityInactivityCompletor.accept(activityForGoal.get());
```
Bert-R: I don't understand what's going on here. The method is named "add missing inactivity", but you're here adding "activity" to an "inactivity". Can you explain?

core/src/test/java/nu/yona/server/analysis/service/ActivityServiceTests.java
```diff
@@ -118,19 +131,135 @@ public void setUp()
...
 	}
+	
+	@Test
+	public void dayActivityOverview()
+	{
+		ZonedDateTime today = getDayStartTime(ZonedDateTime.now(userAnonZone));
+		ZonedDateTime yesterday = today.minusDays(1);
+		
+		// gambling goal was created 2 weeks ago, see above
+		// mock some activity in previous week on yesterday 20:58-21:00
```
Bert-R: You say "in previous week", but yesterday isn't necessarily in a previous week, right?

core/src/test/java/nu/yona/server/analysis/service/ActivityServiceTests.java
```diff
@@ -118,19 +131,135 @@ public void setUp()
...
+						.thenReturn(new HashSet<DayActivity>(
+								Arrays.asList(yesterdayRecordedActivity)));
+		
+		Page<DayActivityOverviewDTO> dayOverviews = service.getDayActivityOverviews(userID, new PageRequest(0, 3));
+		
+		// assert that the right retrieve from database was done
+		verify(mockDayActivityRepository, times(1)).findAll(userAnonID, today.minusDays(2).toLocalDate(), today.toLocalDate());
+		
+		// because the gambling goal was added with creation date two weeks ago, there are multiple days
+		assertThat(dayOverviews.getNumberOfElements(), equalTo(3));
```
Bert-R: Doesn't this vary with the moment in the week you run the test? I.e. on Tuesday, this is right, but on Friday, it's 6?

Bert-R:
I've added some comments and questions. Regarding the tests, I'm a bit concerned. Given that the tests can be run on any day, it's hard to build scenarios that make assumptions about the current weekday.

Have you considered [mocking a clock](http://stackoverflow.com/questions/24491260/mocking-time-in-java-8s-java-time-api)? That way, you can make it look like you always run the test on Friday April 1st 2016. It would imply that our code always determines "now" through a clock that can be mocked for unit tests.

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
aristotelos: Oops, will enable the code formatter again. I reinstalled some software because of a Groovy version error.

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
aristotelos: I don't think it will make it much simpler. If there was no activity on Sunday, Monday will be the first item in the list -- unless we add Sunday on the fly at that moment. If you think that last option should be chosen, I have nothing against it but it sounds as making it more complex.

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
aristotelos: In case of week activities, it may be that there was activity recorded at some days in the week, Wednesday and Thursday for example. When retrieving the week, inactivity days should be added, in this case Sunday, Monday, Tuesday, Friday, Saturday.

core/src/test/java/nu/yona/server/analysis/service/ActivityServiceTests.java
aristotelos: Hmm, copy-paste error of a comment... Will change it

core/src/test/java/nu/yona/server/analysis/service/ActivityServiceTests.java
aristotelos: No, it is always equal to the number of elements retrieved. Updated comment to explain

aristotelos:
I did not know of the option to mock a clock, that sounds like a good possibility indeed for those tests. I made the current tests work for whatever weekday is today (this could be achieved by just making the creation of the goal 2 weeks back and mocking some activity 1 week back), but mocking a clock will make the tests easier to read and extend.

aristotelos:
To 'mock a clock', you need to have a `Clock` bean available which you can inject in the bean you test. So it would mean that all `LocalDate.now`, `ZonedDateTime.now` etc. calls get an extra argument `clock` and that `CoreConfiguration` gets a `Clock` configuration.

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
Bert-R: We anyway need to return an item for that inactive Sunday, right? So why does it make it more complex?

aristotelos:
It seems the `Clock` approach is not going to work. There is no `LocalDate.now(clock, userZone)`, only a `LocalDate.now(clock)` and `LocalDate.now(userZone)` which hard-coded uses the system clock. Same for `ZonedDateTime`. So it is not possible to have a single `Clock` property for the tested `ActivityService` because the zone may depend on the user.

aristotelos:
Processed/replied to your comments

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
aristotelos: Because you're suddenly adding inactivity via two ways: via the batch job (at the end of the week) and on-the-fly (at the start of the week).

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
Bert-R: That's what the method name suggests and what happens on line 161 below, but above here you concluded that there is activity for this goal, so you're adding activity to the existing entity, right? That doesn't sound logical in a method named like this.

core/src/test/java/nu/yona/server/analysis/service/ActivityServiceTests.java
Bert-R: Clear now.

Bert-R:
Responded with two more questions.

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
Bert-R: But now you're making the DTO on the fly for inactive days, so that code could go away, right?

Bert-R:
Regarding the clock: it would be a matter of injecting a clock factory rather than a clock. That way, you can get a clock for a timezone.

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
aristotelos: No. If there was no activity at the end of the week, this needs to be added
using the algorithm above. If there was no activity at all, the same.

Bert-R: Let's stick with this for now and discuss.

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
aristotelos: Added a comment to explain. Even if activity was already recorded (e.g. a `WeekActivity`), it might be that this is not for the complete period (e.g. only for one day in that week), so make the interval activity complete (e.g. add `DayActivity` inactivity) with a consumer.

Bert-R:
Merged it. We'll discuss the remaining topics later and apply changes if we conclude we need to.

PR #168, opened 2016-05-04
YD-220 Implemented pin reset batch job

Bert-R:


adminservice/src/main/java/nu/yona/server/AdminServiceApplication.java
```diff
@@ -1,21 +1,43 @@
...
 import org.springframework.cache.annotation.EnableCaching;
 import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+
+import nu.yona.server.admin.batch.SpringBatchAdmin;
+import nu.yona.server.admin.batch.config.MainConfiguration;
 
 @ComponentScan("nu.yona.server")
-@SpringBootApplication
```
aristotelos: Why not leave this and put the `exclude` here? And also merge the line above with `scanBasePackages`. This way it would look more consistent with the other two applications.

adminservice/src/main/java/nu/yona/server/admin/batch/SpringBatchAdmin.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.context.annotation.Import;
+
+import nu.yona.server.admin.batch.config.MainConfiguration;
+
+/**
+ * Main class to start the Spring Boot application. There are two ways to start it. First: Just run the main method of
+ * this class. Second: Create a war file with the spring-boot-maven-plugin and deploy it to tomcat or another servlet
+ * container.
+ *
+ * @author Thomas Bosch
```
aristotelos: This file is all commented out. Should it not be removed?

adminservice/src/main/java/nu/yona/server/admin/batch/config/ServletConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package nu.yona.server.admin.batch.config;
+
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.ImportResource;
+
+/**
+ * Just the import of the servlet-config.xml of spring-batch-admin-manager.
+ *
+ * @author Thomas Bosch
```
aristotelos: Remove the author?

adminservice/src/main/java/nu/yona/server/admin/batch/config/WebappConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package nu.yona.server.admin.batch.config;
+
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.ImportResource;
+
+/**
+ * Just the import of the webapp-config.xml of spring-batch-admin-manager.
+ *
+ * @author Thomas Bosch
```
aristotelos: Remove the author?

adminservice/src/main/java/nu/yona/server/admin/batch/config/ServletConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+@Configuration
+@ImportResource("classpath:/org/springframework/batch/admin/web/resources/servlet-config.xml")
```
aristotelos: Why not configure it with methods instead? That would be more consistent with all other configuration classes.

adminservice/src/main/java/nu/yona/server/admin/batch/SpringBatchAdmin.java
Bert-R: Yes, it should.

adminservice/src/main/java/nu/yona/server/AdminServiceApplication.java
Bert-R: Because I didn't see that option on ``SpringBootApplication`` :). I'll use that.

adminservice/src/main/java/nu/yona/server/admin/batch/config/ServletConfiguration.java
Bert-R: No, let's give him his credits.

adminservice/src/main/java/nu/yona/server/admin/batch/config/ServletConfiguration.java
Bert-R: This is all take from the sample provided by Thomas. I'd rather not touch it.

adminservice/src/main/java/nu/yona/server/admin/batch/config/WebappConfiguration.java
Bert-R: See earlier comment.

adminservice/src/main/resources/batch-hsql.properties
```diff
@@ -0,0 +1,21 @@
+# Placeholders batch.*
```
aristotelos: Is this used? I see no reference...

aristotelos: OK, I think it is a hard-coded reference in Spring Batch.

Bert-R: That's what I was about to check. I think you're right on that.

adminservice/src/main/java/nu/yona/server/admin/batch/config/ServletConfiguration.java
aristotelos: Agreed! I understand the context of those files now :).

appservice/src/intTest/groovy/nu/yona/server/PinResetRequestTest.groovy
```diff
@@ -72,14 +81,33 @@ class PinResetRequestTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(richard)
 	}
 
-	def 'Clear pin reset request before verification'()
+	def 'Try to verify pin reset confirmation code before end of delay period'()
```
aristotelos: This is a weird scenario, a "Hacking attempt" maybe? You could add that prefix to the test case name.

Bert-R: If the app is built correctly, then it indeed cannot happen, as the link is unavailable. So hacking attempt is indeed a better name. I'll edit that in-place.

appservice/src/intTest/groovy/nu/yona/server/PinResetRequestTest.groovy
```diff
@@ -72,14 +81,33 @@ class PinResetRequestTest extends AbstractAppServiceIntegrationTest
...
 	{
 		given:
 		User richard = addRichard()
 		appService.yonaServer.postJson(richard.pinResetRequestUrl, [:], ["Yona-Password" : richard.password])
 		richard = appService.getUser(appService.&assertUserGetResponseDetailsPublicDataAndVpnProfile, richard.url, true, richard.password)
```
aristotelos: Is this line needed? Only the URL/password are used and those are the same isn't it?

Bert-R: I think you're right on this one. If you don't mind, I remove that line along with something else, to reduce the overhead.

batchservice/build.gradle
```diff
@@ -0,0 +1,135 @@
...
+}
+
+task intTest(type:Test){
+	description = "Run integration tests (located in src/intTest/...)."
+	testClassesDir = project.sourceSets.intTest.output.classesDir
+	classpath = project.sourceSets.intTest.runtimeClasspath
+}
+
+task setupIntTest(type: Test) {
+	include '**/InitialGoalCreationTest.*'
```
aristotelos: Copy-paste I suppose...

batchservice/src/main/java/nu/yona/server/BatchServiceApplication.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.context.web.SpringBootServletInitializer;
+import org.springframework.cache.annotation.EnableCaching;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.scheduling.TaskScheduler;
+import org.springframework.scheduling.annotation.EnableScheduling;
+import org.springframework.scheduling.concurrent.ConcurrentTaskScheduler;
+
+@ComponentScan("nu.yona.server")
```
aristotelos: Can be included in `@SpringBootApplication` below.

batchservice/src/main/java/nu/yona/server/BatchServiceServletInitializer.java
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.boot.builder.SpringApplicationBuilder;
+import org.springframework.boot.context.web.SpringBootServletInitializer;
+
+public class BatchServiceServletInitializer extends SpringBootServletInitializer
+{
+
+	@Override
+	protected SpringApplicationBuilder configure(SpringApplicationBuilder application)
+	{
+		return application.sources(BatchServiceApplication.class);
```
aristotelos: What is this necessary for?

batchservice/src/main/java/nu/yona/server/batch/sample/PinResetConfirmationCodeSenderBatchJob.java
```diff
@@ -0,0 +1,186 @@
+package nu.yona.server.batch.sample;
```
aristotelos: Could we leave the 'sample' out? :)

batchservice/src/main/java/nu/yona/server/batch/sample/PinResetConfirmationCodeSenderBatchJob.java
```diff
@@ -0,0 +1,186 @@
...
+	public Job pinResetConfirmationCodeSenderJob()
+	{
+		return jobBuilderFactory.get("pinResetConfirmationCodeSenderJob").incrementer(new RunIdIncrementer()).flow(step1()).end()
+				.build();
+	}
+
+	@EventListener({ ContextRefreshedEvent.class })
+	void onContextStarted(ContextRefreshedEvent event)
+	{
+		scheduler.scheduleWithFixedDelay(this::runJob,
```
aristotelos: Maybe add a comment here that you do this by code as the annotation did not allow expressions.

batchservice/src/main/java/nu/yona/server/batch/sample/PinResetConfirmationCodeSenderBatchJob.java
```diff
@@ -0,0 +1,186 @@
...
+				yonaProperties.getBatch().getPinResetRequestConfirmationCodeInterval().toMillis());
+	}
+
+	// TODO SPR-13625 @Scheduled(fixedRateString =
+	// "#{T(java.time.Duration).parse('${yona.batch.pinResetRequestConfirmationCodeInterval}').toMillis()}")
+	public void runJob()
```
aristotelos: Does this need to be public?

adminservice/src/main/java/nu/yona/server/admin/batch/config/ServletConfiguration.java
aristotelos: Can you maybe add a link to the original source and a comment that these files were copied somewhere?

aristotelos:
Review done, looks good to me... Quite advanced code in some places, I can understand why it took more time to find out how these newly included libraries work and need to be configured. The result looks great, maybe a few more comments will help out the less experienced reader.

Bert-R:
I believe I addressed all review comments.

PR #170, opened 2016-05-07
YD-235 Fixed locale handling for admin service

Bert-R:
* Simplified the localization filter by removing the locale-specific resource loading, as the app developers anyway cannot use that
* Moved the localization filter to core, so it's also applied to the admin service
* Added a small localization test to the admin service


aristotelos:
Reviewed, no comments, thanks for fixing!

PR #171, opened 2016-05-07
YD-224 Completed goal updateGoal

Bert-R:
* The original goal is now copied as a history item
* EditGoalsTest is extended to verify update works with existing activities
* If the goal is identical, the update is ignored


core/src/main/java/nu/yona/server/goals/service/GoalService.java
```diff
@@ -85,23 +87,43 @@ public GoalDTO addGoal(UUID userID, GoalDTO goal, Optional<String> message)
...
 
 		broadcastGoalChangeMessage(userEntity, existingGoal, GoalChangeMessage.Change.GOAL_CHANGED, message);
+	}
 
-		return GoalDTO.createInstance(existingGoal);
+	private void cloneExistingGoalAsHistoryItem(UserAnonymized userAnonymizedEntity, Goal existingGoal)
+	{
+		Goal historyGoal = existingGoal.cloneAsHistoryItem();
+		existingGoal.setPreviousGoal(historyGoal);
+		WeekActivity.getRepository().findByGoal(existingGoal).stream().forEach(a -> a.setGoal(historyGoal));
```
aristotelos: I hope this will perform. Should we not write a specific update query in the repositories to make it more efficient (`update DayActivity set goal = :historyGoal where goal = :goal`)?

aristotelos:
Review done. Some final remarks:
- Could you please create integration tests that the existing activity is still retrievable after update of a goal, that it has the right goal set (with right end time)? For new goals you may add the assert that they do not have an end time.
- Do you think we should use the same mechanism (cloning as history item) for deleted goals later? In that case there is no new goal to refer to the deleted goal with a relation `previousGoal`, so history becomes a little harder to maintain because the history goal should also be refered from a collection maintained at the user.

core/src/main/java/nu/yona/server/goals/entities/Goal.java
```diff
@@ -29,6 +32,10 @@ public static GoalRepository getRepository()
 	private ActivityCategory activityCategory;
 
 	private ZonedDateTime creationTime;
+	private ZonedDateTime endTime;
+
+	@OneToOne(cascade = CascadeType.ALL, fetch = FetchType.LAZY)
+	private Goal previousGoal;
```
aristotelos: Consider to rename to `previousVersion`. This property is only used when the goal had a previous version, not when you delete a goal and create a goal in the same category.

core/src/main/java/nu/yona/server/goals/entities/Goal.java
```diff
@@ -58,6 +78,23 @@ public ZonedDateTime getCreationTime()
 		return creationTime;
 	}
 
+	public ZonedDateTime getEndTime()
```
aristotelos: Please modify the 'add inactivity' algorithm to use the end time, or create a ticket for that.

aristotelos: While writing this I realize that the 'add inactivity' algorithm will not work yet for historical goals. Please create a ticket that it should be updated to also process the historical goals (it could use the `previousGoal` relation to discover those now, or we could maintain a list of historical goals at the user).

core/src/main/java/nu/yona/server/goals/service/GoalService.java
Bert-R: A hand-written query won't make that any faster. It's more the opposite: JPA now knows that we do a find-by-goal, so it _might_ create an index.

core/src/main/java/nu/yona/server/goals/entities/Goal.java
Bert-R: Added [YD-239](http://jira.yona.nu/browse/YD-239)

core/src/main/java/nu/yona/server/goals/entities/Goal.java
Bert-R: Done

Bert-R:
> Could you please create integration tests that the existing activity is still retrievable after update of a goal, that it has the right goal set (with right end time)? For new goals you may add the assert that they do not have an end time.

We'll do that as a separate action. That's the subject I want to discuss in today's call.

Bert-R:
> Do you think we should use the same mechanism (cloning as history item) for deleted goals later?

No. I expect that for a deleted goal, we'll set the end time and leave it with that.

core/src/main/java/nu/yona/server/goals/service/GoalService.java
aristotelos: My comment was not about the `findByGoal` but about the `stream().forEach()` part. If there are many (thousands of) activities, this will take long. Therefore I believe it is more efficient and necessary to use a specific update query. We can manually specify that the goal column needs an index.

Bert-R: Such optimizations we should only make when there is evidence we have to. Till then, we should use the object abstraction.

aristotelos: After discussion: because the number of `DayActivity` entities per user will not be incredibly large (e.g. with 3 goals and a year of activity it will be about 1000), this construction is acceptible.

aristotelos:
Agreed that we do not need it now, and whether we need it later depends on
the functional decision if and how a user is able to clear history
(currently deleting a goals is the only way to do that).

PR #172, opened 2016-05-07
YD-148 Improve error reporting

Bert-R:
Several supportability improvements:
* Custom (basic) error page
* All HTTP errors are now logged as warnings
* It's now possible to do a GET on for instance /activitycategories/, to see whether the service is up.

### DIFF
--- a/core/src/main/java/nu/yona/server/rest/ErrorLoggingFilter.java
+++ b/core/src/main/java/nu/yona/server/rest/ErrorLoggingFilter.java
@@ -1,0 +1,63 @@
+package nu.yona.server.rest;
+
+import java.io.IOException;
+
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.HttpStatus.Series;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ErrorLoggingFilter implements Filter
+{
+	private static final Logger logger = LoggerFactory.getLogger(ErrorLoggingFilter.class);
+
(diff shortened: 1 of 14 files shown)


core/src/main/resources/templates/error.vm
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+<body>
```
aristotelos: Please add `<head><title>Request handling failed</title></head>` to make the HTML valid.

core/src/main/resources/templates/error.vm
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
aristotelos: Where is this file used exactly? I cannot find a reference...

core/src/main/java/nu/yona/server/rest/ErrorLoggingFilter.java
```diff
@@ -0,0 +1,63 @@
...
+	@Override
+	public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain chain)
+			throws IOException, ServletException
+	{
+		HttpServletRequest request = (HttpServletRequest) servletRequest;
+		HttpServletResponse response = (HttpServletResponse) servletResponse;
+
+		chain.doFilter(request, response);
+
+		if (HttpStatus.Series.valueOf(response.getStatus()) == Series.SUCCESSFUL)
```
aristotelos: Are 500 errors not logged already because they are related to exceptions? So you could also filter for the 400 series (invalid request). Although logging errors two times is not very problematic.

aristotelos: Which is http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/http/HttpStatus.Series.html#CLIENT_ERROR

Bert-R: I'd rather see an error logged twice than missing one. We are logging certain 400 and 500 errors, but not all.

core/src/main/resources/templates/error.vm
Bert-R: This is the default error page. A standard Spring Boot feature when Apache Velocity is available.

Bert-R:
Pushed update with <head>

PR #173, opened 2016-05-11
YD-205 Date/time handling update

Bert-R:
* Replaced ``java.util.Date`` with ``java.time.ZonedDateTime``, in the product and in all tests.
* Instead of adding a minute to all activity durations, now it sets a minimum activity duration of 1 minute
* All properties that correspond to a time span are now of type ``Duration`` and have values like PT1M to indicate 1 minute.


core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
```diff
@@ -144,22 +139,22 @@ private int getDurationInSpreadItem(ZonedDateTime startTime, ZonedDateTime endTi
...
 		{
-			// partial span
-			return (int) startTime.until(endTime, ChronoUnit.MINUTES) + 1;
+			// partial span of spread item
+			return (int) startTime.until(endTime, ChronoUnit.MINUTES);
 		}
 		else if (spreadItemIndex == spreadStartIndex)
 		{
-			// start part
+			// start spread item
```
aristotelos: Here I meant that the start of the activity occurs somewhere in the spread item. So the spanning is here from somewhere in the spread item to the end. A better comment is `// activity starts somewhere in spread item and continues till the end`.

core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
```diff
@@ -144,22 +139,22 @@ private int getDurationInSpreadItem(ZonedDateTime startTime, ZonedDateTime endTi
...
 			return 15 - (startTime.getMinute() % 15);
 		}
 		else if (spreadItemIndex == spreadEndIndex)
 		{
-			// end part
-			return 1 + (endTime.getMinute() % 15);
+			// end spread item
```
aristotelos: `// spans begin of spread item till activity end point (which occurs inside spread item)`

core/src/main/java/nu/yona/server/analysis/service/ActivityDTO.java
```diff
@@ -18,22 +16,25 @@
...
 public class ActivityDTO
 {
-	private Date startTime;
-	private Date endTime;
+	private ZonedDateTime startTime;
+	private ZonedDateTime endTime;
 
 	@JsonCreator
-	public ActivityDTO(@JsonProperty("startTime") Date startTime, @JsonProperty("endTime") Date endTime)
+	public ActivityDTO(@JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ss.SSSZ") @JsonProperty("startTime") ZonedDateTime startTime,
```
aristotelos: We should consider to remove duplicate code for the pattern.

aristotelos:
Great work! I made a few comments about comments that I think were changed to the worse. Another point of discussion is how fair it is to just make activity always 1 minute instead of properly rounding to minutes. If a user has activity of 5 seconds it becomes 1 minute, if a user has activity of 1 minute 59 seconds it will also become 1 minute. Should we not just "round up" (ceiling)?

About the change to `ZonedDateTime` I see only one possible drawback: storage size of `ZonedDateTime` is bigger, isn't it? Might this be relevant for the `Activity` class which will have the most instances in the DB? We may consider to optimise that later if required.

core/src/main/java/nu/yona/server/analysis/service/ActivityDTO.java
Bert-R: Absolutely. I remembered an issue in using constants in annotations, so I didn't try it. Triggered by you, I tried it and it simply works.

aristotelos: You may create a new ticket for that to make merging this PR easier.

Bert-R:
I have improved the comments in ``DayActivity`` and centralized the date pattern.
Regarding the other two comments:

* The objective of setting a minimum duration of 1 minute was to prevent from having activities that actually don't have a duration. It was not to solve any mathematical problem. I don't see what the practical benefit would be of ceiling all durations.
* The storage might indeed be bigger, as a timezone is included. Whether that has practical implications will vary by database, as some might always store a timezone in their date columns. Let's look into this at the time we move to MariaDB.

aristotelos:
OK, let's leave the details about rounding activities to minutes for now.
It is not that important.

core/src/main/java/nu/yona/server/analysis/service/ActivityDTO.java
Bert-R: I take it that you added this comment before seeing that I already centralized the pattern, right?

aristotelos: Indeed

PR #174, opened 2016-05-13
YD-242 Update to existing mobile number returns correct error code

Bert-R:


aristotelos:
Reviewed, no comments

PR #175, opened 2016-05-14
YD-243 YD-197 Send messages on user overwrite

Bert-R:
* Behavior change: remove user and overwrite user now seem immediately applicable to buddies
* Buddies are now informed about account overwrite. In case of a pending buddy request, a reject is received. In case of an existing buddy relationship, a disconnect is received.
* Added test: Overwrite user created on buddy request
* Updated OverwriteUserTest for the changed behavior, including:
  * Updated "Overwrite the existing user" to become "Overwrite Richard while being a buddy of Bob"
  * Added "Overwrite Richard with pending buddy invitation to Bob"
  * Added "Overwrite Richard with pending buddy invitation from Bob"
  * Removed "Bob removes overwritten user Richard as buddy" as the buddy is now immediately removed
  * Added "Goal conflict for Bob after Richard overwrote his account is not reported to Richard"
* Updated RemoveUserTest for the changed behavior, including:
  * Updated "Remove Richard and verify that Bob receives a remove buddy message" to become "Remove Richard while being a buddy of Bob and verify that Bob receives a buddy disconnect message"
  * Removed "Remove Richard and verify that buddy data is removed after Bob processed the remove buddy message", as that's not relevant now that removal is immediately active
  * Added "Remove Richard with pending buddy request from Bob and verify that Bob receives a reject message"
* Several name corrections:
  * login ID --> userAnonymizedID
  * relatedAnonymousMessageID --> relatedMessageID
  * All messages now talk about sender... where applicable and use a consistent parameter sequence
* Updated the Swagger spec to indicate that the embedded user in BuddyMessageWithEmbeddedUser is optional.
* It's now possible to have a BuddyMessage without a sender user
* It's now possible to have a getRelatedUserAnonymizedID without an anonymized user ID
* BuddyService.removeBuddy now really removes the buddy object (earlier it only removed it from the user but left it in the database).

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
@@ -377,10 +377,117 @@ class CreateUserOnBuddyRequestTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(richard)
 	}
 
+	def 'Overwrite user created on buddy request'()
+	{
+		given:
+		def richard = addRichard()
+		def mobileNumberBob = "+$timestamp"
+		def inviteURL = buildInviteUrl(sendBuddyRequestForBob(richard, mobileNumberBob))
+
+		when:
+		def response = appService.requestOverwriteUser(mobileNumberBob)
+
+		then:
+		response.status == 200
+
+		User bob = appService.addUser(this.&assertUserOverwriteResponseDetails, "B o b", "Bob Changed",
+				"Dunn Changed", "BD Changed", mobileNumberBob, ["overwriteUserConfirmationCode": "1234"])
+		bob
+		bob.firstName == "Bob Changed"
+		bob.lastName == "Dunn Changed"
+		bob.nickname == "BD Changed"
+		bob.mobileNumber == mobileNumberBob
+		bob.goals.size() == 1 //mandatory goal
+		bob.goals[0].activityCategoryUrl == GAMBLING_ACT_CAT_URL
+
+		def buddiesRichard = appService.getBuddies(richard)
+		buddiesRichard.size() == 0
+
+		def buddiesBob = appService.getBuddies(bob)
+		buddiesBob.size() == 0
+
+		def getMessagesResponse = appService.getMessages(richard)
+		getMessagesResponse.status == 200
+		getMessagesResponse.responseData._embedded
+		getMessagesResponse.responseData._embedded."yona:messages".size() == 1
+
+		def buddyConnectResponseMessages = getMessagesResponse.responseData._embedded."yona:messages".findAll{ it."@type" == "BuddyConnectResponseMessage"}
+		def buddyConnectResponseMessage = buddyConnectResponseMessages[0]
+		buddyConnectResponseMessage.message == "User account was deleted"
+		buddyConnectResponseMessage.nickname == "Bob Dunn"
+		buddyConnectResponseMessage._links.self.href.startsWith(richard.messagesUrl)
+		buddyConnectResponseMessage._links?."yona:process"?.href?.startsWith(getMessagesResponse.responseData._embedded."yona:messages".findAll{ it."@type" == "BuddyConnectResponseMessage"}[0]._links.self.href)
+
+		def getResponse = appService.getMessages(richard)
+		def disconnectMessage = getResponse.responseData._embedded?."yona:messages".findAll{ it."@type" == "BuddyConnectResponseMessage"}[0]
+		def processURL = disconnectMessage?._links?."yona:process"?.href
+		processURL
+		def processDisconnectResponse = appService.postMessageActionWithPassword(processURL, [:], richard.password)
+		processDisconnectResponse.status == 200
+
+		User richardAfterBobOverwrite = appService.getUser(appService.&assertUserGetResponseDetailsPublicDataAndVpnProfile, richard.url, true, richard.password)
+		richardAfterBobOverwrite.buddies.size() == 0
+
+		cleanup:
+		appService.deleteUser(richard)
+		appService.deleteUser(bob)
+	}
+
+	def 'Overwrite user created on buddy request and connect again'()
+	{
+		given:
+		def richard = addRichard()
+		def mobileNumberBob = "+$timestamp"
+		def inviteURL = buildInviteUrl(sendBuddyRequestForBob(richard, mobileNumberBob))
+		appService.requestOverwriteUser(mobileNumberBob)
+		User bob = appService.addUser(this.&assertUserOverwriteResponseDetails, "B o b", "Bob Changed",
+				"Dunn Changed", "BD Changed", mobileNumberBob, ["overwriteUserConfirmationCode": "1234"])
+
+		when:
+		appService.makeBuddies(richard, bob)
+		analysisService.postToAnalysisEngine(richard, ["news/media"], "http://www.refdag.nl")
+
+		then:
+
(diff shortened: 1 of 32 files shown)


appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -377,12 +377,119 @@ class CreateUserOnBuddyRequestTest extends AbstractAppServiceIntegrationTest
...
+		getMessagesResponse.responseData._embedded."yona:messages".size() == 1
+
+		def buddyConnectResponseMessages = getMessagesResponse.responseData._embedded."yona:messages".findAll{ it."@type" == "BuddyConnectResponseMessage"}
+		def buddyConnectResponseMessage = buddyConnectResponseMessages[0]
+		buddyConnectResponseMessage.message == "User account was deleted"
+		buddyConnectResponseMessage.nickname == "Bob Dunn"
+		buddyConnectResponseMessage._links.self.href.startsWith(richard.messagesUrl)
+		buddyConnectResponseMessage._links?."yona:process"?.href?.startsWith(getMessagesResponse.responseData._embedded."yona:messages".findAll{ it."@type" == "BuddyConnectResponseMessage"}[0]._links.self.href)
+
+		def getResponse = appService.getMessages(richard)
```
aristotelos: This is a bit duplicate. The `buddyConnectResponseMessage` and `disconnectMessage` are equal, isn't it?

appservice/src/intTest/groovy/nu/yona/server/OverwriteUserTest.groovy
```diff
@@ -73,47 +73,170 @@ class OverwriteUserTest extends AbstractAppServiceIntegrationTest
...
+		def buddies = appService.getBuddies(bob)
+		buddies.size() == 0
+
+		cleanup:
+		appService.deleteUser(richardChanged)
+		appService.deleteUser(bob)
+	}
+
+
+	def 'Goal conflict for Bob after Richard overwrote his account is not reported to Richard'()
```
aristotelos: You mean "is not reported to the detached to be deleted account"

appservice/src/intTest/groovy/nu/yona/server/RemoveUserTest.groovy
```diff
@@ -51,61 +51,144 @@ class RemoveUserTest extends AbstractAppServiceIntegrationTest
...
+
+		def getMessagesResponse = appService.getMessages(bob)
+		getMessagesResponse.status == 200
+		getMessagesResponse.responseData._embedded
+		getMessagesResponse.responseData._embedded."yona:messages".size() == 1
+
+		def buddyConnectResponseMessages = getMessagesResponse.responseData._embedded."yona:messages".findAll{ it."@type" == "BuddyConnectResponseMessage"}
+		buddyConnectResponseMessages.size() == 1
+		def buddyConnectResponseMessage = buddyConnectResponseMessages[0]
+		buddyConnectResponseMessage.message == "User account was deleted"
```
aristotelos: We could consider to add a `reason` here, such that the app can apply special style if required and because a user may fill the same message. But maybe that is not necessary.

appservice/src/intTest/groovy/nu/yona/server/OverwriteUserTest.groovy
Bert-R: Yes, this name is confusing. Now named it "Overwrite Richard while being a buddy of Bob and verify Richard does not receive Bob\'s goal conflicts anymore", in line with the other tests.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
Bert-R: Correct. Removed the duplication.

appservice/src/intTest/groovy/nu/yona/server/RemoveUserTest.groovy
Bert-R: Given that such automated responses will be rare, I think it's sufficient to have a (localized) text.

Bert-R:
Pushed an update with the improved tests.

core/src/main/resources/messages/messages.properties
```diff
@@ -8,6 +8,8 @@
 message.user.account.deleted=User account was deleted
 message.user.removed.buddy=User removed you as a buddy
 
+message.temp.nickname={0} {1}
```
aristotelos: Can we add a comment here that {0} {1} is first name / last name?

Bert-R: OK

core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
```diff
@@ -162,4 +163,9 @@ public UUID getUserAnonymizedID()
 	{
 		return userAnonymizedID;
 	}
+
+	public Set<Buddy> getBuddiesOfRemovedUsers()
```
aristotelos: Rename to `getBuddiesHavingUserRemoved`?

Bert-R: How about ``getBuddiesRelatedToRemovedUsers``?

core/src/main/java/nu/yona/server/subscriptions/service/BuddyDisconnectMessageDTO.java
```diff
@@ -120,7 +120,7 @@ public MessageActionDTO handleAction(UserDTO actingUser, Message messageEntity,
 		private MessageActionDTO handleAction_Process(UserDTO actingUser, BuddyDisconnectMessage messageEntity,
 				MessageActionDTO requestPayload)
 		{
-			buddyService.removeBuddyAfterBuddyRemovedConnection(actingUser.getID(), messageEntity.getUserID());
+			buddyService.removeBuddyAfterBuddyRemovedConnection(actingUser.getID(), messageEntity.getSenderUserID());
```
aristotelos: This process action is no longer needed if `messageEntity.getReason() == DropBuddyReason.USER_ACCOUNT_DELETED`, is it? I think we could remove the action in this case?

core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
aristotelos: Then rather `getBuddiesRelatingToRemovedUsers`...

aristotelos:
Review done! It took some time to read... Nice improvements and tests

core/src/main/java/nu/yona/server/subscriptions/service/BuddyDisconnectMessageDTO.java
Bert-R: I can give it a try.

core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
Bert-R: I don't think present continuous is necessary here. Common English is "related to".

aristotelos: OK, agreed.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyDisconnectMessageDTO.java
aristotelos: Hmm, on further thought this is maybe not the best. The user may not be retrieved but his messages may be...

PR #176, opened 2016-05-16
YD-168 Buddy activity

aristotelos:


core/src/test/java/nu/yona/server/exceptions/ResourceBasedExceptionTest.java
```diff
@@ -43,7 +40,7 @@ protected TestException(String messageId, Object... parameters)
 
 	}
 
-	@Test
+	// @Test
```
Bert-R: Why? Same for next test.

aristotelos: The tests are failing on my laptop that has nl-NL installed because they give Dutch instead of English translation. Will enable them again and create a separate ticket.

aristotelos:
To do: I need to update the Swagger spec still.

core/src/test/java/nu/yona/server/exceptions/ResourceBasedExceptionTest.java
Bert-R: I just pushed an update to your branch that I expect to fix the issue. Please check.

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -12,168 +12,257 @@ import java.time.ZoneId
...
+	}
+
+	def 'Get day activity overviews from buddy'()
+	{
+		given:
+		setupTestScenario()
 
 		when:
-		def response = appService.getDayActivityOverviews(richard)
+		def response = appService.getResourceWithPassword(bobDailyActivityReportsUrlForRichard, richard.password)
```
Bert-R: Please move this into a method on ``AppService`` with this signature:
```
def getBuddyDayActivityOverviews(User user, int buddyIndex =  0, parameters = [:])
```
That allows doing this:
```
def response = appService.getBuddyDayActivityOverviews(richard)
```

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -12,168 +12,257 @@ import java.time.ZoneId
...
+		def bobGoalsForRichard = richardWithBuddy.buddies[0].goals
+		bobBudgetGoalGamblingUrlForRichard = bobGoalsForRichard.find{it.activityCategoryUrl == GAMBLING_ACT_CAT_URL}.url
+		bobBudgetGoalNewsUrlForRichard = bobGoalsForRichard.find{it.activityCategoryUrl == NEWS_ACT_CAT_URL}.url
+		bobTimeZoneGoalUrlForRichard = bobGoalsForRichard.find{it.activityCategoryUrl == SOCIAL_ACT_CAT_URL}.url
+		bobDailyActivityReportsUrlForRichard = richardWithBuddy.buddies[0].dailyActivityReportsUrl
+		assert bobDailyActivityReportsUrlForRichard
+		bobWeeklyActivityReportsUrlForRichard = richardWithBuddy.buddies[0].weeklyActivityReportsUrl
+		assert bobWeeklyActivityReportsUrlForRichard
+
+		addSomeTestActivity(bob)
```
Bert-R: In all of our tests, Richard Quin (RQ) is making the ReQuests and and Bob Dunn (BD) is the watching BuDdy. I know it's quite a bit of work, so I hesitate to ask, but still I do: can you swap the roles of Bob and Richard here and add the activity for Richard?

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -12,168 +12,257 @@ import java.time.ZoneId
...
 		def dayActivityForGoal = dayActivityOverview._embedded."yona:dayActivities".find{ it._links."yona:goal".href == budgetGoalGamblingUrl}
-		dayActivityForGoal?._links?.self?.href
-		def response = appService.getResourceWithPassword(dayActivityForGoal._links.self.href, richard.password)
+		assert dayActivityForGoal?._links?.self?.href
+		def response = appService.getResourceWithPassword(dayActivityForGoal._links.self.href, fromUser.password)
+		assert response.status == 200
+		assert response.responseData.spread
+		assert response.responseData.spread.size() == 96
+		assert response.responseData.totalActivityDurationMinutes == 1
+		assert response.responseData.goalAccomplished == false
```
Bert-R: Given that line 29 defines the timezone as ``["11:00-12:00"]``, this test fails when run between 11:00 and 12:00, right? We need to improve on this with a bigger and better set of integration test cases, but not now.

For now, please add a comment, so we know what goes wrong if we ever run the test in that interval.

appservice/src/main/java/nu/yona/server/analysis/rest/ActivityControllerBase.java
```diff
@@ -40,136 +32,115 @@
...
+	protected final String WEEK_ACTIVITY_OVERVIEWS_URI_FRAGMENT = "/weeks/";
+	protected final String DAY_OVERVIEWS_URI_FRAGMENT = "/days/";
+	protected final String WEEK_ACTIVITY_DETAIL_URI_FRAGMENT = "/weeks/{date}/details/{goalID}";
+	protected final String DAY_ACTIVITY_DETAIL_URI_FRAGMENT = "/days/{date}/details/{goalID}";
+	protected final String GOAL_PATH_VARIABLE = "goalID";
+	protected final String DATE_PATH_VARIABLE = "date";
+
+	protected HttpEntity<PagedResources<WeekActivityOverviewResource>> getWeekActivityOverviews(Optional<String> password,
+			UUID userID, Pageable pageable, PagedResourcesAssembler<WeekActivityOverviewDTO> pagedResourcesAssembler,
+			Supplier<Page<WeekActivityOverviewDTO>> activityFunction, LinkProvider linkProvider)
```
Bert-R: Rename ``activityFunction`` into ``activitySupplier``, as the type as well as the functionality indicate that.

Bert-R:
I've added a few comments and am done with the review. Glad to see the genericity of the approach.

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
aristotelos: No, this test does not assert something about the time zone goal. It only asserts for `budgetGoalGamblingUrl`.

PR #178, opened 2016-05-19
YD-246 Added resend pin reset confirmation code link

Bert-R:


aristotelos:
Review done. Would it be possible for a hacker to first request a confirmation code and then request a resend to get a confirmation code immediately? Maybe we should add an integration test for this.
Is there already an integration test for the scenario that the user requests a confirmation code twice? Is the second request ignored because there already is one?

PR #180, opened 2016-05-20
YD-247 inconsistencies in swagger spec

Bert-R:
* PUT on /users/{id} actually returns a UserWithPrivateData, not a PostPutUser as used to be in the spec.
* DELETE on /users/{id} takes a message query parameter
* We were not consistent in the way we specify IDs. E.g. we use /users/{id}} and /users/{userID}/messages/.


aristotelos:
Review done, no comments.

PR #181, opened 2016-05-21
YD-249 Remove user failed

Bert-R:
* Extended the test to reproduce the scenario
* Changed the order of the equals operation in MessageDestination.removeMessagesFromUser


appservice/src/intTest/groovy/nu/yona/server/RemoveUserTest.groovy
```diff
@@ -118,6 +118,10 @@ class RemoveUserTest extends AbstractAppServiceIntegrationTest
 		response.responseData._embedded."yona:affectedMessages"[0]._links.self.href == buddyConnectResponseMessage._links.self.href
 		response.responseData._embedded."yona:affectedMessages"[0]._links."yona:process" == null
 
+		def newRichard = addRichard()
```
aristotelos: Consider adding a comment that this tests an advanced scenario.

aristotelos:
Reviewed, added one comment.

PR #183, opened 2016-05-25
YD-248 Scenario API for activity integration testing

Bert-R:


Bert-R:
This change set is not ready yet to for merging, but I'm already creating this PR, so you can start the review and see where I'm heading.

The current change set has a failing assert:

```
Condition not satisfied:

response.responseData.spread == expectedSpread
|        |            |      |  |
|        |            |      |  [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 15, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 15, 10, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
|        |            |      false
|        |            [0, 0, 0, 0, 0, 15, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 15, 10, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
|        [_embedded:[MONDAY:[_links:[self:[href:http://localhost:8082/users/0d5347e4-64ef-4a07-900e-6bfa45d3400a/activity/days/2016-05-16/details/95853636-55ea-492c-ba0a-3afc15bb8e8d]], goalAccomplished:false, timeZoneId:Europe/Amsterdam, totalActivityDurationMinutes:20, totalMinutesBeyondGoal:20], TUESDAY:[_links:[self:[href:http://localhost:8082/users/0d5347e4-64ef-4a07-900e-6bfa45d3400a/activity/days/2016-05-17/details/95853636-55ea-492c-ba0a-3afc15bb8e8d]], goalAccomplished:false, timeZoneId:Europe/Amsterdam, totalActivityDurationMinutes:25, totalMinutesBeyondGoal:25]], _links:[curies:[[href:http://dev.yona.nu/rels/{rel}, name:yona, templated:true]], self:[href:http://localhost:8082/users/0d5347e4-64ef-4a07-900e-6bfa45d3400a/activity/weeks/2016-W21/details/95853636-55ea-492c-ba0a-3afc15bb8e8d], yona:goal:[href:http://localhost:8082/users/0d5347e4-64ef-4a07-900e-6bfa45d3400a/goals/95853636-55ea-492c-ba0a-3afc15bb8e8d]], date:2016-W21, spread:[0, 0, 0, 0, 0, 15, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 15, 10, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], timeZoneId:Europe/Amsterdam, totalActivityDurationMinutes:45]
groovyx.net.http.HttpResponseDecorator@27c96b84

	at nu.yona.server.ActivityTest.assertWeekDetailForGoal(ActivityTest.groovy:100)
	at nu.yona.server.ActivityTest.New style test(ActivityTest.groovy:62)
```

This seems to be caused by a time zone issue. It looks like the server uses UTC to calculate the spread index.

Bert-R:
Please review it as it stands now. This is the set of APIs I would like to take forward.

For the issues found, I've added different expectations in the test, marked with TODOs and JIRA issue keys.

Bert-R:
I've pushed an updated Swagger spec, see the commit message. To play with the API, you can also find it at https://swaggerhub.com/api/yonadev/YonaAppService/1.0.0.

My next step is to align the product and the tests with the updated spec.

Bert-R:
This PR is ready for review. I do not plan to add more changes to it.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
```diff
@@ -172,4 +182,64 @@ class YonaServer
...
+		LocalDateTime now = LocalDateTime.now()
+		switch (fields.size())
+		{
+			case 3:
+				assert fields[0].startsWith("W")
+				weekOffset = Integer.parseInt(fields[0].substring(1))
+				parsedFields++
+			// Fall through
+			case 2:
+				int weekDay = DateTimeFormatter.ofPattern("eee").parse(fields[parsedFields]).get(ChronoField.DAY_OF_WEEK)
```
aristotelos: For Sunday, `DAY_OF_WEEK` gives 7 while we regard it as the first day of the week in the Yona app. So when you give relative date time string `"W-1 Sun"` I think you will first be corrected to next week Sunday (`dayOffset` will be positive) and then to this week Sunday (`weekOffset` is -1). However, the intention was to get to past week Sunday. So please correct the `weekDay` to 0 if the day is Sunday.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
```diff
@@ -172,4 +182,64 @@ class YonaServer
...
+				dayOffset = weekDay - now.dayOfWeek.value
```
aristotelos: This is not correct I think: because `now` is a `LocalDateTime` and you use the value to create a `ZonedDateTime` in zone Europe/Amsterdam it may end up in a different weekday than intended, isn't it?

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
```diff
@@ -172,4 +182,64 @@ class YonaServer
...
+	}
+
+	public static int getDayOfWeek(ZonedDateTime dateTime)
+	{
+		getDayOfWeek(dateTime.dayOfWeek.value)
+	}
+
+	public static int getDayOfWeek(int javaDayOfWeek)
+	{
+		(javaDayOfWeek == 7) ? 0 : javaDayOfWeek
```
aristotelos: Please add a comment to explain that this uses 0 for Sunday because in the Yona app the week starts with Sunday.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
```diff
@@ -172,4 +182,64 @@ class YonaServer
...
+				new DateTimeFormatterBuilder().appendPattern("HH:mm[:ss][.SSS]")
+				.parseDefaulting(ChronoField.EPOCH_DAY, epochDay)
+				.toFormatter()
+				.withZone(ZoneId.of("Europe/Amsterdam"))
+		ZonedDateTime.parse(timeString, formatter)
+	}
+
+	static def relativeDateStringToDaysOffset(weeksBack, shortDay)
+	{
+		int weekDay = getDayOfWeek(DateTimeFormatter.ofPattern("eee").parse(shortDay).get(ChronoField.DAY_OF_WEEK))
```
aristotelos: Time zone is not specified here but should be Europe/Amsterdam isn't it?

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
```diff
@@ -172,4 +182,64 @@ class YonaServer
...
+		return weeksBack * 7 - weekDay + LocalDateTime.now().dayOfWeek.value
```
aristotelos: Same here, should use a zoned date time in time zone Europe/Amsterdam, and should use `getDayOfWeek` method instead, isn't it?

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
```diff
@@ -172,4 +182,64 @@ class YonaServer
...
+	{
+		DateTimeFormatter formatter =
+				new DateTimeFormatterBuilder().appendPattern("HH:mm[:ss][.SSS]")
+				.parseDefaulting(ChronoField.EPOCH_DAY, epochDay)
+				.toFormatter()
+				.withZone(ZoneId.of("Europe/Amsterdam"))
+		ZonedDateTime.parse(timeString, formatter)
+	}
+
+	static def relativeDateStringToDaysOffset(weeksBack, shortDay)
```
aristotelos: Please add a comment explaining what this method does

core/src/test/java/nu/yona/server/analysis/service/DayActivityTests.java
```diff
@@ -29,7 +29,7 @@ public void setUp()
 		testZone = ZoneId.of("Europe/Amsterdam");
 		ActivityCategory activityCategory = ActivityCategory.createInstance(UUID.randomUUID(),
 				Collections.singletonMap(Locale.US, "being bored"), false, Collections.emptySet(), Collections.emptySet());
-		goal = BudgetGoal.createInstance(activityCategory, 60);
+		goal = BudgetGoal.createInstance(ZonedDateTime.now(), activityCategory, 60);
```
aristotelos: Is this necessary?

core/src/test/java/nu/yona/server/analysis/service/AnalysisEngineServiceTests.java
```diff
@@ -399,7 +403,7 @@ public void noMessagesCreatedOnNonZeroBudgetGoal()
 	{
 		// Execute the analysis engine service.
 		Set<String> conflictCategories = new HashSet<String>(Arrays.asList("social"));
-		service.analyze(new NetworkActivityDTO(userAnonID, conflictCategories, "http://localhost/test"));
+		service.analyze(new NetworkActivityDTO(userAnonID, conflictCategories, "http://localhost/test", Optional.empty()));
```
aristotelos: Can't we leave this argument out?

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -169,8 +170,8 @@ private void addMandatoryGoals(User userEntity)
 
 	private void addGoal(User userEntity, ActivityCategoryDTO category)
```
aristotelos: I guess this method should have been renamed already longer ago to `addMandatoryNoGoGoal`...

core/src/main/java/nu/yona/server/goals/service/GoalService.java
```diff
@@ -90,16 +91,30 @@ public GoalDTO updateGoal(UUID userID, UUID goalID, GoalDTO newGoalDTO, Optional
 		User userEntity = userService.getUserByID(userID);
 		Goal existingGoal = getGoalEntity(userEntity, goalID);
 
-		verifyGoalUpdate(existingGoal, newGoalDTO);
-
-		if (newGoalDTO.isGoalChanged(existingGoal))
+		if (newGoalDTO.getCreationTime().isPresent())
```
aristotelos: This method has become more complex. Can't we just use the previous version and check on change in creation time?
Otherwise add a comment here explaining why this is an exceptional case.

core/src/main/java/nu/yona/server/goals/service/GoalDTO.java
```diff
@@ -28,20 +30,21 @@
...
 	private final UUID id;
 	private UUID activityCategoryID;
-	private final ZonedDateTime creationTime;
+	private final Optional<ZonedDateTime> creationTime;
 	private final boolean mandatory;
 
-	protected GoalDTO(UUID id, UUID activityCategoryID, ZonedDateTime creationTime, boolean mandatory)
+	protected GoalDTO(UUID id, Optional<ZonedDateTime> creationTime, UUID activityCategoryID, boolean mandatory)
 	{
+		Objects.requireNonNull(creationTime);
```
aristotelos: Can we also verify here that setting the creation time is only allowed in test environments, not on the production environment? I suppose an end user/the app is not allowed to do this?

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityDTO.java
```diff
@@ -74,11 +74,16 @@ public String getDate()
...
 
-	/*
-	 * The time zone in which the interval was recorded.
+	/**
+	 * The time zone in which the interval was recorded, or {null} if the date should not be serialized.
 	 */
+	@JsonInclude(Include.NON_EMPTY)
 	public String getTimeZoneId()
 	{
+		if (!shouldSerializeDate)
```
aristotelos: Nice correction!

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
```diff
@@ -70,19 +70,36 @@
...
+				weekActivityEntitiesByLocalDate);
 		return new PageImpl<WeekActivityOverviewDTO>(
-				weekActivityEntitiesByDate.entrySet().stream().sorted((e1, e2) -> e2.getKey().compareTo(e1.getKey()))
+				weekActivityEntitiesByZonedDate.entrySet().stream().sorted((e1, e2) -> e2.getKey().compareTo(e1.getKey()))
 						.map(e -> WeekActivityOverviewDTO.createInstance(e.getKey(), e.getValue())).collect(Collectors.toList()),
 				pageable, yonaProperties.getAnalysisService().getWeeksActivityMemory());
 	}
 
+	private <T extends IntervalActivity> Map<ZonedDateTime, Set<T>> mapToZonedDateTime(
+			Map<LocalDate, Set<T>> weekActivityEntitiesByLocalDate)
```
aristotelos: Please rename the parameter and the local variable to `intervalActivityEntitiesByLocalDate`/`intervalActivityEntitiesByZonedDate`.

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
```diff
@@ -70,19 +70,36 @@
...
+	{
+		Map<ZonedDateTime, Set<T>> weekActivityEntitiesByZonedDate = weekActivityEntitiesByLocalDate.entrySet().stream()
+				.collect(Collectors.toMap(e -> e.getValue().iterator().next().getStartTime(), e -> e.getValue()));
+		return weekActivityEntitiesByZonedDate;
+	}
+
+	private Map<LocalDate, Set<WeekActivity>> getWeekActivitiesGroupedByDate(UUID userAnonymizedID, Interval interval)
```
aristotelos: Please apply this extract method to day activities as well if you think it makes the code more readable.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1925,213 +2008,204 @@ definitions:
...
       - date
       - timeZoneId
       - totalActivityDurationMinutes
       - goalAccomplished
       - totalMinutesBeyondGoal
+      - _links
   PagedMessages:
-    type: object
+    allOf:
+      - $ref: '#/definitions/Page'
```
aristotelos: Nice abstraction, I was already thinking we should make this.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1925,213 +2008,204 @@ definitions:
...
+      _links:
+        allOf:
+          - $ref: '#/definitions/Curies'
+          - $ref: '#/definitions/SelfLink'
+        description: The self link is currently not returned from the server. See YD-203
+    required:
+      - date
+      - dayActivities
+      - _links
+  DayGoalActivityBase:
```
aristotelos: I'm not sure if I like the changes in the Swagger file to expand all day types and create a base class for it. You removed only 3 comments that specify that an attribute is not always there, and I think it maybe adds more complexity to the Swagger result to have a base class than to have 3 comments.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1818,105 +1911,95 @@ definitions:
...
     properties:
-      _links:
-        allOf:
-          - $ref: '#/definitions/Curies'
-          - $ref: '#/definitions/SelfLink'
       date:
         type: string
         example: 2016-W2
         description: The week date in ISO 8601 format (YYYY-'W'w)
-      _embedded:
```
aristotelos: You removed the embbedded resources with self link to retrieve more details because it returns different content I guess. That is an improvement, and I think maybe even required to have a consistent API.

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -21,6 +21,84 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
...
+
+		when:
+		def responseWeekOverviews = appService.getWeekActivityOverviews(richard)
+		def responseDayOverviews = appService.getDayActivityOverviews(richard)
+
+		then:
+		assertWeekOverviewBasics(responseWeekOverviews, [3, 2])
+		def weekOverviewLastWeek = responseWeekOverviews.responseData._embedded."yona:weekActivityOverviews"[1]
+		assertNumberOfReportedDaysForGoalInWeekOverview(weekOverviewLastWeek, budgetGoalNewsUrl, 6)
+		assertDayInWeekOverviewForGoal(weekOverviewLastWeek, budgetGoalNewsUrl, expectedValuesLastWeek, "Mon")
```
aristotelos: It would be nicer if we can just assert all those lines at once by passing the `expectedValuesLastWeek` set, isn't it?

aristotelos:
Wow, this is the biggest PR ever I guess... I finally reviewed it and put some comments and questions. Overall I think this is a good direction, although I am not sure I like all the details.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
Bert-R: That's not relevant here, as it's transformed into a date. So today, W-1 Sun results in May 29, 2016.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
Bert-R: No. ``parseTimeForDay`` adds timezone Europe/Amsterdam to it.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
Bert-R: OK

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
Bert-R: > Same here, should use a zoned date time in time zone Europe/Amsterdam

Correct.

> and should use getDayOfWeek method instead, isn't it?

In Groovy, you can use the attribute syntax.

> For the rest this expression is hard to read and understand

I've chosen your last alternative.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
Bert-R: OK

core/src/test/java/nu/yona/server/analysis/service/AnalysisEngineServiceTests.java
Bert-R: Yes, but since it's just for the test, I don't think we should change the production code.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
Bert-R: Renamed into ``addNoGoGoal``. Left the mandatoryness out of the name, as that's not part of the method behavior but part of the activity category that's passed to it.

core/src/main/java/nu/yona/server/goals/service/GoalService.java
Bert-R: Added a comment. I want to keep the test-specific behavior as special.

core/src/main/java/nu/yona/server/goals/service/GoalDTO.java
Bert-R: That's correct, but currently we don't have a setting like that. The check would need to happen in the ``JsonCreator`` constructors of ``TimeZoneGoal`` and ``BudgetGoal``. Given that nothing really goes wrong if someone would squeeze the creation time in, I don't think it's worth the effort.

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
Bert-R: OK

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
Bert-R: OK

appservice/src/main/resources/static/swagger/swagger-spec.yaml
Bert-R: From our perspective, that's indeed true, but from the consumer's perspective, things improved a lot, as it's now explicit what we return when. That's why I did this.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
Bert-R: The main reason to remove it is that an embedded resource should be a true resource, retrievable separately or embedded in another object. In our case, these inner objects aren't retrievable separately. That's why I removed ``_embedded``.

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
Bert-R: That's rather easy to do, it makes it hard to interpret what fails. Now you can base yourself on the line number, but then the assert needs to tell everything. Let's see what we favor when we start to add more tests.

Bert-R:
Pushed an update that addresses your comments.
This PR is indeed way too large. Given that all successive changes built on top of the first one, I had to combine them.

I'm even not done yet. I regret that I added ``goalSpreadCells`` and ``goalBudget``. These repeat for  many times, resulting in a lot of data. Instead, I'm going to move ``spreadCells`` to ``TimeZoneGoalDTO`` and embed the inactive/history goals in the user, so the app can take the goal data from there. That reduces the amount of data and makes the server more extensible when we ever add new goal types.

core/src/test/java/nu/yona/server/analysis/service/AnalysisEngineServiceTests.java
aristotelos: OK accepted

core/src/main/java/nu/yona/server/goals/service/GoalDTO.java
aristotelos: OK accepted, it is not the most important scenario for hacking :)

appservice/src/main/resources/static/swagger/swagger-spec.yaml
aristotelos: OK accepted, I think it is necessary after all to make resource distinction between the items in overview resources and the detail resources.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
aristotelos: OK. Actually they were retrievable but returned the details resource instead which could contain more attributes. Making it a different resource makes sense.

Notice that the items of the paged overview resources are not yet separately retrievable and do not have a self link yet.

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
aristotelos: Does the assert not show a line number when failing?

appservice/src/main/resources/static/swagger/swagger-spec.yaml
Bert-R: Yes, I noticed that and added a comment for it. See line 1956 below.

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
Bert-R: Yes, but when you loop through a set of days, you only see it's handling the set, not what day it's handling.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
aristotelos: Suppose it is Tuesday in the test server timezone and Monday in Europe/Amsterdam timezone. Then if you give Monday as target day the day offset will be 1. However, the intention was to get to Monday in Europe/Amsterdam so the day offset should be 0.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
aristotelos: So that is incorrect. W-1 Sun should result in 22 May 2016. 29 May is contained in the current week in the Yona app.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
Bert-R: Sorry, yes, you're right. I thought your comment applied to the line above. I'll change this.

aristotelos:
Indeed that sounds as a further improvement. You could combine the first part (moving the `goalSpreadCells` and removing the `goalBudget`) with this PR and save the other part for a new PR.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
Bert-R: I made a mistake reading my calendar. It returns May 22.

If anything goes wrong, it will be when running the tests on a Sunday, but I'm not entirely sure of that.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
aristotelos: Aha, I didn't know the Groovy attribute syntax. I'm not sure I like it
though, it does not look standard and for a normal C#/java programmer it
may be hard to guess that the custom getDayOfWeek is called here, isn't it?

Bert-R:
Possible, but it does not matter that much. The entire set needs to move into the master today, as it is to be delivered to Mobiquity today.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
Bert-R: Groovy is Java related but different. I prefer to leverage the simplifications that Groovy offers over Java. We could do many things Java-style that we now do Groovy-style (think of loops for instance).

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
aristotelos: Have you already tested what is the result of `"W-1 Sun"`? I mean with executing java code, not by looking at your calendar ;)

Bert-R: I have tested quite a few of them last week, but honestly, I don't know this one.

If you're not convinced, please create a ticket to verify it later.

aristotelos:
OK, merged because of the time pressure and because we can correct small things later.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
aristotelos: For me, this 'simplification' makes it harder to understand.

core/src/testUtils/groovy/nu/yona/server/YonaServer.groovy
aristotelos: Done, http://jira.yona.nu/browse/YD-259

PR #187, opened 2016-06-07
YD-260 YD-267 Correct swagger spec

Bert-R:


aristotelos:
Reviewed, no comments

PR #189, opened 2016-06-08
YD-266 Fixed time zone length bug

Bert-R:
And enabled 24:00 as end time.


aristotelos:
Reviewed, no questions or suggestions. Nice tests added!

PR #191, opened 2016-06-11
YD-270 extend activity categories

Bert-R:
On request of Kinnar and Jan


Bert-R:
I've merged this one myself, so I can prepare the next build for Mobiquity

PR #194, opened 2016-06-14
Yd 225 add timestamp to user

rcprins:
Added creation time to user


appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
```diff
@@ -272,6 +275,8 @@ class UserTest extends AbstractAppServiceIntegrationTest
 		assert user.firstName == "John"
 		assert user.lastName == "Doe"
 		assert user.mobileNumber == "+${timestamp}"
+		assertEquals(user.creationTime, ZonedDateTime.now())
```
Bert-R: Please use ``YonaServer.now`` instead of ``ZonedDateTime.now()``. This is a relatively new API to centralize the creation of time stamps in the Europe/Amsterdam time zone.

core/src/testUtils/groovy/nu/yona/server/test/User.groovy
```diff
@@ -14,6 +14,7 @@ class User
 	final String firstName
 	final String lastName
 	final String mobileNumber
+	final String creationTime
```
Bert-R: Please declare this as ``ZonedDateTime creationTime``. See ``Goal.groovy`` for similar code.

Bert-R:
Added comments to the test code. Besides this: the branch has conflicts and needs to be synced with the master.

PR #195, opened 2016-06-14
YD-263 Cache goal DTOs inside UserAnonymizedDTO

Bert-R:
instead of goal entities

Along with this, the following is done:
* Add a test that failed before the fix (failed to lazily initialize a collection)
* Renamed GoalService.getGoal into getGoalForUserID and added getGoalForUserAnonymizedID
* Renamed UserService.getUserByID into getUserEntityByID, in alignment with other get...Entity methods
* Extended GoalService with operations to deal with user anonymized

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/service/UserAnonymizedDTO.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/UserAnonymizedDTO.java
@@ -36,7 +36,7 @@ public class UserAnonymizedDTO
 
 	public static UserAnonymizedDTO createInstance(UserAnonymized entity)
 	{
-		return new UserAnonymizedDTO(entity.getID(), getGetGoalsIncludingHistoryItems(entity),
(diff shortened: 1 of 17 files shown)


core/src/main/java/nu/yona/server/subscriptions/service/UserAnonymizedDTO.java
```diff
@@ -76,18 +77,35 @@ public BuddyAnonymized getBuddyAnonymized(UUID fromUserAnonymizedID)
...
-		{
-			// no earlier version
-			return goal;
-		}
+		return goal;
+	}
 
-		// recursive on previous version
-		return getOldestVersionOfGoal(goal.getPreviousVersionOfThisGoal().get());
+	static Set<GoalDTO> getGetGoalsIncludingHistoryItems(UserAnonymized userAnonymizedEntity)
```
rcprins: I don't understand why you use "getGet..."?

PR #206, opened 2016-07-05
YD-300 Corrected the week number

Bert-R:


appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -544,6 +551,20 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
...
+		boolean isSunday = YonaServer.now.getDayOfWeek() == DayOfWeek.SUNDAY
+		DateTimeFormatter weekFormatter = new DateTimeFormatterBuilder()
+				.parseCaseInsensitive()
+				.appendValue(IsoFields.WEEK_BASED_YEAR, 4, 10, SignStyle.EXCEEDS_PAD)
+				.appendLiteral("-W")
+				.appendValue(IsoFields.WEEK_OF_WEEK_BASED_YEAR, 2)
+				.toFormatter(YonaServer.EN_US_LOCALE)
+
+		String currentWeek = weekFormatter.format(YonaServer.now)
+		assert responseWeekOverviews.responseData._embedded."yona:weekActivityOverviews"[0].date == currentWeek || isSunday
```
aristotelos: Maybe add a comment why it doesn't work on Sunday.

core/src/main/java/nu/yona/server/analysis/service/IntervalActivityOverviewDTO.java
```diff
@@ -34,9 +32,11 @@ public String getTimeZoneId()
...
-	 * The ISO-8601 date formatter.
+	/**
+	 * Format the date in compliance with ISO-8601
+	 * 
+	 * @param localDate The date to be formatted
+	 * @return The formatted date
 	 */
-	@JsonIgnore
-	protected abstract DateTimeFormatter getDateFormatter();
+	protected abstract String formatDateAsISO(LocalDate localDate);
```
aristotelos: Should the `@JsonIgnore` not be retained here?

aristotelos: Sorry, my mistake, it is not a getter any more.

aristotelos:
Looks great, added one comment.

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
Bert-R: Done.

PR #207, opened 2016-07-06
YD-292 Added prev and next links to day and week activity details

Bert-R:


aristotelos:
Ai... ik was ook net aan het committen...

PR #210, opened 2016-07-14
YD 303 VPN profile, SSL certificate on server

aristotelos:
Moved OVPN profile to static/vpn/profile.ovpn
Added SSL root CA certificate to static/vpn/sslrootcert.cer
Added MIME types
Added placeholders/Swagger for Apple .mobileconfig support and for user-specific VPN authentication certificate, this is subject to change


appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -118,6 +118,22 @@
 		return createOKResponse(userService.getPublicUser(id), false);
 	}
 
+	@RequestMapping(value = "/{id}/vpnAuthCertificate.crt", produces = { "application/x-x509-user-cert" })
```
aristotelos: Dit zijn placeholders voor iOS support. Wijzigt waarschijnlijk nog. Ik kan het ook eruit halen als je dat beter vindt...

Bert-R: Just keep it. As far as I can see, the basics are right.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -118,6 +118,22 @@
...
+	public @ResponseBody byte[] getVpnAuthCertificate(@RequestHeader(value = Constants.PASSWORD_HEADER) Optional<String> password,
+			@PathVariable UUID id)
+	{
+		return CryptoSession.execute(password, () -> userService.canAccessPrivateData(id),
+				() -> userService.getPrivateUser(id).getPrivateData().getVpnProfile().getVpnAuthCertificateByteArray());
+	}
+
+	@RequestMapping(value = "/{id}/apple.mobileconfig", produces = { "application/x-apple-aspen-config" })
+	public @ResponseBody byte[] getVpnAppleMobileConfig(
```
Bert-R: My understanding is that this is a text file, so ``String`` would be a more appropriate return type. We can do this at the time we finalize the iOS-related code.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -189,6 +189,58 @@ paths:
...
+        description: The ID of the user
+        required: true
+        type: string
+      - name: Yona-Password
+        in: header
+        description: The device password
+        required: true
+        type: string
+    get:
+      summary: Get the user Apple mobile config
```
Bert-R: Please rephrase: Get the Apple mobile config for this user

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -189,6 +189,58 @@ paths:
...
+      description: Fetches the user-specific .mobileconfig file needed for iOS devices. Notice this contains VPN username and password.
+      tags:
+        - User
+      responses:
+        '200':
+          description: The config
```
Bert-R: The config **file**

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -542,7 +594,7 @@ paths:
         description: The ID of the user
         required: true
         type: string
-      - name: YYYY-'W'ww
+      - name: week
```
Bert-R: Good catch!

appservice/src/main/resources/static/vpn/sslrootcert.cer
```diff
@@ -0,0 +1,19 @@
+-----BEGIN CERTIFICATE-----
```
Bert-R: Why is this called a "root" certificate? Isn't this simply the SSL certificate for VPN? If so, the name ``ssl.cer`` would be less confusing.

core/src/main/java/nu/yona/server/subscriptions/service/VPNProfileDTO.java
```diff
@@ -60,21 +57,18 @@ public void setVpnPassword(String vpnPassword)
...
-		catch (IOException e)
-		{
-			throw YonaException.unexpected(e);
-		}
+		return new VPNProfileDTO(user.getAnonymized().getID(), user.getVPNLoginID(), user.getVPNPassword(),
+				user.getVPNAuthCertificateByteArray());
 	}
 
-	public static VPNProfileDTO createInstance(User user)
+	/*
```
Bert-R: Please move this method before the static method(s).

appservice/src/main/java/nu/yona/server/MimeTypeConfigurer.java
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+public class MimeTypeConfigurer implements EmbeddedServletContainerCustomizer
+{
+	@Override
+	public void customize(ConfigurableEmbeddedServletContainer container)
+	{
+		MimeMappings mappings = new MimeMappings(MimeMappings.DEFAULT);
+		mappings.add("ovpn", "application/x-openvpn-profile"); // see vpn/profile.ovpn
```
Bert-R: This file is missing.

Bert-R:
Made some minor comments.

There is a bigger one left: On the ticket [YD-303](http://jira.yona.nu/browse/YD-303), I mentioned 4 additional tasks. The third one is about changing ``VPNProfileDTO`` from a simple DTO into a resource and add links to it. This is not implemented. As a result, we are extending the set of URLs that's known to the app, which is not a good thing.

At the other hand: I am not sure whether the app can be installed without access to the OVPN profile and the related SSL certificate. If they need it at the time of installation rather than user sign-up, then it's pointless making links, as the links would become available too late in the process. I lack the knowledge to judge that, so I leave that to you.

If we continue without links, it's even more important to get the name of the SSL certificate right. If the certificate is indeed the SSL certificate for the VPN, then it should be a "regular" certificate, not a root one. If it's the MITM interception certificate, then it's indeed a root certificate, but not related to VPN. then It should move to a different folder (e.g. ``ssl/rootcert.cer``).

appservice/src/main/resources/static/swagger/swagger-spec.yaml
aristotelos: Swagger editor catched it for me :)

aristotelos:
I guess it is indeed better to include the links in the VPN profile, or rather one level higher in the user DTO. We can include the .mobileconfig, the .ovpn, the SSL .cer. I will see if I can do this this evening.

appservice/src/main/resources/static/vpn/sslrootcert.cer
aristotelos: Renamed to `ssl/rootcert.cer` because it is the SSL root certificate.

aristotelos:
Added the links now. Added to private user data after mobile number is confirmed. I thought it was better to not embed them in VPN profile DTO because SSL root certificate is not related to the VPN profile (nor the .mobileconfig link for iOS).

One thing I noted and that may be a concern is that the URLs are relative and not absolute. Maybe that needs to be changed, but I am not sure what is the best way to do so.

PR #211, opened 2016-07-15
YD-309 Added app activity validations

Bert-R:
* Start time must be before end type
* Start time and end time cannot be in the future

Added tests. Made explicit that we accept a 10 second deviation between the app and the server time.


appservice/src/intTest/groovy/nu/yona/server/AppActivityTest.groovy
```diff
@@ -209,5 +220,99 @@ class AppActivityTest extends AbstractAppServiceIntegrationTest
...
+
+		then:
+		response.status == 400
+		response.responseData.code == "error.analysis.invalid.app.activity.data.starts.in.future"
+
+		cleanup:
+		appService.deleteUser(richard)
+	}
+
+	def 'Send app activity within device time inacuracy window'()
```
aristotelos: Spelling error: "accuracy"

aristotelos:
Reviewed, looks great.

PR #212, opened 2016-07-16
YD-313 fixed inconsistency in totals

Bert-R:
For a time zone goal, the time beyond the goal is calculated by summing the conflicting spread cells. Due to rounding, that sum might be more than the total time spent. This is confusing, so it's corrected now.


aristotelos:
Looks good, apart from a missing test example ;)

Bert-R:
Agreed, we need to add a test. However, the test is currently disabled, see [YD-287](http://jira.yona.nu/browse/YD-287) and [YD-312](http://jira.yona.nu/browse/YD-312). It would be great if you would be able to revive that test

PR #217, opened 2016-07-23
YD-276 Fixed activity reporting issues related to goal updates

Bert-R:


appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
```diff
@@ -248,6 +274,12 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
 		assert dayActivityForGoal._links.self == null  // This is not a top level or embedded resource
 	}
 
+	private Goal findWeekGoal(expectedValues, activityCategoryUrl)
+	{
+		def expectedValuesEndOfWeek = (expectedValues.size() == 1) ? expectedValues[getCurrentShortDay()] : expectedValues["Sat"]
```
aristotelos: Why end of week? I thought the `WeekActivity` was constructed when the first `DayActivity` was constructed for that week?

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -98,7 +99,7 @@ public UserAnonymized getUserAnonymized()
 
 	public Goal getGoal()
 	{
-		return goal;
+		return EntityUtil.enforceLoading(goal);
```
aristotelos: Maybe add a comment "this is necessary because instances of this class are cached outside the hibernate session".

I guess this has to be done for all entity object references that somehow end up in the cache. Maybe we can introduce an interface for all cached objects with a method `forceLoadReferedEntityObjects` that is called just after computing them (or alternatively do some complicated things with reflection). This will become quite complicated however...

core/src/main/java/nu/yona/server/goals/service/GoalService.java
```diff
@@ -108,19 +114,15 @@ public GoalDTO updateGoal(UUID userID, UUID goalID, GoalDTO newGoalDTO, Optional
 		User userEntity = userService.getUserEntityByID(userID);
 		Goal existingGoal = getGoalEntity(userEntity, goalID);
 
-		if (newGoalDTO.getCreationTime().isPresent())
+		verifyGoalUpdate(existingGoal, newGoalDTO);
+		if (newGoalDTO.getCreationTime().isPresent() && !newGoalDTO.isGoalChanged(existingGoal))
```
aristotelos: I see that this is necessary for the test, however, I think that it is good to do a check that the passed creation time is later than the creation time of the existing goal to avoid strange errors at Mobiquity...

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
```diff
@@ -137,28 +138,50 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
...
 	}
 
-	void addTimeZoneGoal(User user, activityCategoryURL, zones, ZonedDateTime creationTime = YonaServer.now)
+	TimeZoneGoal addTimeZoneGoal(User user, activityCategoryURL, zones, ZonedDateTime creationTime = YonaServer.now)
 	{
 		appService.addGoal(appService.&assertResponseStatusCreated, user, TimeZoneGoal.createInstance(creationTime, activityCategoryURL, zones.toArray()))
 	}
 
-	void addBudgetGoal(User user, activityCategoryURL, int maxDurationMinutes, relativeCreationDateTimeString)
+	void updateTimeZoneGoal(User user, TimeZoneGoal originalGoal, zones, relativeCreationDateTimeString)
```
aristotelos: You could change this void in `TimeZoneGoal`. Same for the four following methods.

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
```diff
@@ -137,28 +138,50 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
...
 		addBudgetGoal(user, activityCategoryURL, maxDurationMinutes, creationTime)
 	}
 
-	void addBudgetGoal(User user, activityCategoryURL, int maxDurationMinutes, ZonedDateTime creationTime = YonaServer.now)
+	BudgetGoal addBudgetGoal(User user, activityCategoryURL, int maxDurationMinutes, ZonedDateTime creationTime = YonaServer.now)
 	{
 		appService.addGoal(appService.&assertResponseStatusCreated, user, BudgetGoal.createInstance(creationTime, activityCategoryURL, maxDurationMinutes))
 	}
 
+	void updateBudgetGoal(User user, BudgetGoal originalGoal, int maxDurationMinutes, relativeCreationDateTimeString)
```
aristotelos: Void can be replaced by `BudgetGoal`.

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
```diff
@@ -137,28 +138,50 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
...
+	{
+		ZonedDateTime updateTime = YonaServer.relativeDateTimeStringToZonedDateTime(relativeCreationDateTimeString)
+		appService.updateGoal(appService.&assertResponseStatusSuccess, user, originalGoal.editURL, BudgetGoal.createInstance(originalGoal, updateTime, maxDurationMinutes))
+	}
+
+	void updateBudgetGoal(User user, BudgetGoal originalGoal, int maxDurationMinutes)
```
aristotelos: Maybe good to change the parameter name `originalGoal` to `modifiedOriginalGoal`?

aristotelos:
Looks good, I hope that all parts of the code that assumed the goal ID did not change have been updated now... I made some minor comments/suggestions.

The check for `wasActiveAtInterval` was also fixed, I hope that there are no other parts of the code using the goal start time only?

aristotelos:
I see you also placed YD-239 under review. This has however not yet been completely fixed: in `ActivityService` the method `getActiveGoals` should be updated to use `getGoalsIncludingHistoryItems` (unless we think history goal inactivity should not be added dynamically but only by the script?).

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
Bert-R: A goal update always applies to the day or week at which it was updated. That implies that in case of multiple updates in that day or week, the last update applies. That's why this code finds the goal applicable to the last day of the week.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
Bert-R: It's not the caching but the JPA lazy loading.

Honestly, I don't know why I was facing the proxy issue here. The goal is not lazily loaded, so I was expecting it to be available along with the activity, but it wasn't always.

core/src/main/java/nu/yona/server/goals/service/GoalService.java
Bert-R: The Mobiquity app is not setting the creation time, so that's not an issue here. I'll add that check, also for our own safety when writing tests.

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
Bert-R: Good point, will do.

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
Bert-R: Good point, will do.

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
Bert-R: Updated goal would be better. Will change it.

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
aristotelos: OK

PR #220, opened 2016-08-02
YD-275 Add more activity tests

Bert-R:


appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -24,7 +24,7 @@ import nu.yona.server.test.User
 
 class ActivityTest extends AbstractAppServiceIntegrationTest
 {
-	def 'Fetch activity reports without activity'()
+	def 'Retrieve activity reports without activity'()
```
aristotelos: Please make the name more speaking to show all tested variations

PR #222, opened 2016-08-03
YD-327 Move onlyUnreadMessage query parameter to the correct URL

Bert-R:


appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -531,13 +531,6 @@ paths:
     get:
       summary: Get messages
       description: Fetches all messages related to this activity
-      parameters:
```
aristotelos: Line should be retained (I'll fix this this evening)

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1175,6 +1168,13 @@ paths:
       summary: Get messages
       description: Fetches all messages of the user with the given ID
       parameters:
+      parameters:
```
aristotelos: Duplicate `parameters:` tag

aristotelos:
This parameter is missing on week details messages as well

Bert-R:
I moved the ``parameters`` tag back.

The parameter ``onlyUnreadMessages`` only applies to the user messages collection, not to activity messages. I accidentally put the parameter on the wrong URL (day details).

PR #224, opened 2016-08-05
YD-311 Link by default without parameter

aristotelos:


Bert-R:
You tested that this indeed fixes it?

Bert-R:
To ensure this behavior, you could change the below line in ``UserTest.groovy``
```
john.messagesUrl.startsWith(john.url + "/messages/")
```
into this:
```
john.messagesUrl == john.url + "/messages/"```

aristotelos:
Done

PR #235, opened 2016-09-06
YD-339 Messages thread order

aristotelos:


Bert-R:
About as good as it gets. I'm not happy with the move of the replied message ID and the thread ID to the base ``Message`` class, but we'll leave it there for now.

Strictly speaking, it's unnecessary. The activity ID isn't there either and it still features in the where clause of ``MessageRepository.findByActivityID``. That works because all messages are mapped into the same table. We could have done that for ``threadHeadMessageID`` as well. That would have prevented from making the base class aware of things it doesn't need to know, but it would not have been nice either.

The whole message story is high on my list of to be refactored code.

BTW: I updated the tests for the changes done through other PRs.

aristotelos:
OK thanks for updating the tests! I was planning to do that too :)

PR #236, opened 2016-09-06
Fix apple mobileconfig

aristotelos:
Fixed \r\n to \n
Fixed \n in username\npassword
Other changes by Anders


core/src/main/resources/templates/apple.mobileconfig.vm
```diff
@@ -191,7 +207,7 @@
 				<key>auth</key>
 				<string>SHA1</string>
 				<key>auth-user-pass</key>
-				<string>${ldapUsername}\\${ldapPassword}</string>
+				<string>${ldapUsername}\\n${ldapPassword}</string>
```
aristotelos: Wait a sec. This should be \n and not \\n.

PR #239, opened 2016-09-09
YD-353 Fixing exception, no reproduce yet

aristotelos:
I would like to have a reproduce that can be added to the integration
test suite. I have asked Enosh if he can send us a full Charles log.


Bert-R:
Looks good. Will merge this after the current build is done

PR #241, opened 2016-09-09
YD-356 Added deep link landing page

aristotelos:
Added page as provided by Kinnar. Not sure how it detects that the app
is installed.

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
@@ -47,6 +47,25 @@ class CreateUserOnBuddyRequestTest extends AbstractAppServiceIntegrationTest
 		// TODO: How to delete the invited user?
 	}
 
+	def 'Bob opens the link and goes to the buddy invitation deep link landing page'()
+	{
+		given:
+		def richard = addRichard()
+		def mobileNumberBob = "+$timestamp"
+		def deepLinkLandingPageURL = buildInviteDeepLinkLandingPageUrl(sendBuddyRequestForBob(richard, mobileNumberBob))
+
+		when:
+		def response = appService.yonaServer.getNonJsonResource(deepLinkLandingPageURL)
+
+		then:
+		response.status == 200
+		response.contentType == "text/html"
+
+		cleanup:
+		appService.deleteUser(richard)
+		// Cannot delete Bob, as he did not set up a Yona password yet.
+	}
+
 	def 'Bob downloads the app and uses it to open the link sent in the e-mail and load the prefilled data provided by Richard'()
 	{
 		given:
@@ -507,5 +526,10 @@ class CreateUserOnBuddyRequestTest extends AbstractAppServiceIntegrationTest
 	String buildInviteUrl(def response)
 	{
 		response.responseData._embedded."yona:user"._links.self.href + "?tempPassword=abcd"
+	}
+
+	String buildInviteDeepLinkLandingPageUrl(def response)
+	{
+		response.responseData._embedded."yona:user"._links.self.href + "/buddy-invitation-deep-link-landing-page.html?tempPassword=abcd"
 	}
 }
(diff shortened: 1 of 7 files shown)


core/src/main/resources/templates/buddy-invitation-deep-link-landing-page.vm
```diff
@@ -0,0 +1,24 @@
...
+        2. if user doesn't have app installed, then their browser wouldn't recognize the URL scheme
+        and app wouldn't open since it's not installed. In 1 second (1000 milliseconds) user is redirected
+        to download app from app store.
+     -->
+    <script>
+    window.onload = function() {
+    <!-- Deep link URL for existing users with app already installed on their device -->
+        window.location = "${esc.javascript($deepLinkURL)}";
+    <!-- Download URL (MAT link) for new users to download the app -->
+        setTimeout("window.location = 'https://rink.hockeyapp.net/apps/e1e6d73a1d244306ba07e8a43a8565f4/app_versions/1';", 500);
```
Bert-R: This URL is not correct. Before, we used the Hockey app to distribute the app, but now it's in the Google Play Store. We need to find out what the URL there is.

Bert-R: On my request to Kinnar how we know whether to load Google Play Store or iTunes, he responded this:

> That can be handle from Java script.
> Here is sample link: http://app.makemytrip.com/i2ozz-d2/5 from another application, if you click from Android device, it will open play store and if you open on iPhone, it will open iTunes. We need to implement similar way in email invitation link.

That we also need to add.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -47,6 +47,25 @@ class CreateUserOnBuddyRequestTest extends AbstractAppServiceIntegrationTest
 		// TODO: How to delete the invited user?
 	}
 
+	def 'Bob opens the link and goes to the buddy invitation deep link landing page'()
+	{
+		given:
+		def richard = addRichard()
+		def mobileNumberBob = "+$timestamp"
+		def deepLinkLandingPageURL = buildInviteDeepLinkLandingPageUrl(sendBuddyRequestForBob(richard, mobileNumberBob))
```
Bert-R: This is a bit a strange test. The actor here is Richard, but he's masquerading Bob. Please add an explanatory comment.

Bert-R:
One more thing: Bastiaan isn't happy with the approach yet. He feels it's strange to ask the user to tap the same link twice. That makes a lot of sense to me too.

The deep link approach is very useful when the user receives mails with such links before and after installing the app. In our case however, we are only sending that mail before the user installed the app (and registered in it). After that, we're not sending the mail anymore but use in-app notifications. So why is it important to have "transparent links" for us?

Bert-R:
Kinnar just provided more information:

> there should be only one link in email for inviting user, when user clicks on that link:
>
> - server webpage identify platform either Android or iPhone
>   - If Android, it will check:
>      - whether application is installed - if Yes, it will open application directly from there with detail page  for friend request: accept/reject 
>      - If Application not installed on device, it will take user in play store and user need to press download button, as soon as user download and click on open, it will keep the clicked link by user in memory and redirect user on friend accept/reject page.
>
> Same logic apply for iOS, if user open email/sms on iOS device, it will do same steps.
>
> Also, if user open same link on webpage, it will show appropriate message on webpage and show 2 buttons there for Android and iOS stores to install application on device.

With this, the feature seems useful. The only issue is the app ID. The app is in the Yona Play Store now, so I guess the link is this: https://play.google.com/store/apps/details?id=nu.yona.app

Asked Kinnar for confirmation.

Bert-R:
Kinnar confirmed the URL

aristotelos:
Kinnar said "it will keep the clicked link by user in memory and redirect user on friend accept/reject page". How will this magic happen? Will Kinnar provide the javascript or do you have an idea?

PR #243, opened 2016-09-13
YD-308 Buddy info change message

aristotelos:


core/src/main/java/nu/yona/server/subscriptions/entities/User.java
```diff
@@ -277,7 +278,8 @@ public void setIsCreatedOnBuddyRequest()
 
 	public Buddy getBuddyByUserAnonymizedID(UUID relatedUserAnonymizedID)
 	{
-		return getBuddies().stream().filter(buddy -> buddy.getUserAnonymizedID().equals(relatedUserAnonymizedID)).findAny().get();
+		return getBuddies().stream().filter(buddy -> buddy.getUserAnonymizedID().isPresent()
```
aristotelos: This method was previously never used, corrected it now

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyInfoChangeMessage.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.UUID;
+
+import javax.persistence.Entity;
+
+import nu.yona.server.messaging.entities.BuddyMessage;
+
+@Entity
+public class BuddyInfoChangeMessage extends BuddyMessage
+{
+	private String newNickname;
```
Bert-R: This needs to be encrypted.

core/src/main/java/nu/yona/server/subscriptions/entities/User.java
Bert-R: Sorry, I missed that one.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyInfoChangeMessageDTO.java
```diff
@@ -0,0 +1,139 @@
...
+					return super.handleAction(actingUser, messageEntity, action, requestPayload);
+			}
+		}
+
+		private MessageActionDTO handleAction_Process(UserDTO actingUser, BuddyInfoChangeMessage messageEntity,
+				MessageActionDTO requestPayload)
+		{
+			if (messageEntity.getRelatedUserAnonymizedID().isPresent())
+			{
+				buddyService.updateBuddyInfo(actingUser.getID(), messageEntity.getRelatedUserAnonymizedID().get(),
```
Bert-R: Just call this ``updateNickname``, as that's all it does.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -2372,6 +2373,26 @@ definitions:
     required:
       - dropBuddyReason
       - _links
+  BuddyInfoChangeMessage:
```
Bert-R: Going through the code a bit further, I understand the intention of this name: the message indicates a change in the buddy object, which only carries the nickname, a status and a reference to the user.

Still, I stick with my proposal: let's have a single message for changes in user information. For now, it only comes for a nickname change. Later, it'll also come for changes in the name or mobile number. When we add the photo, we'll also use this, but that'll again be a change in the buddy object.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -525,4 +526,45 @@ private User getBuddyUser(BuddyDTO buddy)
 		}
 		return Optional.empty();
 	}
+
+	@Transactional
+	void broadcastBuddyInfoChangedMessageIfNeeded(User updatedUserEntity, UserDTO originalUser)
```
Bert-R: Please rename into ``broadcastUserInfoChangeToBuddies``. Why not always broadcast the event? That way, we immediately have the full feature of informing buddies about changes in the personal info of their friends.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -525,4 +526,45 @@ private User getBuddyUser(BuddyDTO buddy)
...
+		{
+			broadcastBuddyInfoChangeMessage(updatedUserEntity, originalUser);
+		}
+	}
+
+	private String getBuddyInfoChangeMessage(User updatedUserEntity, UserDTO originalUser)
+	{
+		if (!updatedUserEntity.getNickname().equals(originalUser.getPrivateData().getNickname()))
+		{
+			return translator.getLocalizedMessage("message.buddy.changed.nickname");
```
Bert-R: If we can agree to make the info message broader, I propose to go with a global message: personal information change. The user can change multiple things at once.

appservice/src/intTest/groovy/nu/yona/server/UpdateBuddyTest.groovy
```diff
@@ -0,0 +1,55 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v.2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at https://mozilla.org/MPL/2.0/.
+ *******************************************************************************/
+package nu.yona.server
+
+import groovy.json.*
+import nu.yona.server.test.User
+
+class UpdateBuddyTest extends AbstractAppServiceIntegrationTest
```
aristotelos: OK

appservice/src/main/resources/static/swagger/swagger-spec.yaml
aristotelos: OK, will remove `newNickname` from the DTO.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyInfoChangeMessageDTO.java
aristotelos: The intent of this method is broader: to update any information on the `Buddy` entity. Currently it is only nickname, later on it will include photo and maybe other things.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
aristotelos: Hmm, sounds as a good idea.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyInfoChangeMessageDTO.java
Bert-R: OK. One of the things that confused me is the term "buddy info", as that's new. What about "buddy entity"?

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
aristotelos: Agreed

aristotelos:
What about `updateBuddyUserInfo`?

appservice/src/intTest/groovy/nu/yona/server/UpdateBuddyTest.groovy
aristotelos: I renamed the test name to `UpdateBuddyUserInfoTest`. I don't like to add it to `BasicBuddyTest` because it is already so long (and hard to find it there later).

appservice/src/main/resources/static/swagger/swagger-spec.yaml
aristotelos: Done

aristotelos:
Reworked the review comments

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -302,12 +302,12 @@ private User addUserCreatedOnBuddyRequestInSubTransaction(UserDTO buddyUserResou
 	public UserDTO updateUser(UUID id, UserDTO user)
 	{
 		User originalUserEntity = getUserEntityByID(id);
```
Bert-R: Renamed this back into ``originalUserEntity``, to align with the DTO name.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -320,25 +320,26 @@ public UserDTO updateUser(UUID id, UserDTO user)
...
 		}
 		logger.info("Updated user with mobile number '{}' and ID '{}'", userDTO.getMobileNumber(), userDTO.getID());
+		buddyService.broadcastUserInfoChangeToBuddies(savedUserEntity, originalUser);
 		return userDTO;
 	}
 
-	private void validateUpdateRequest(UserDTO user, User originalUserEntity)
+	private void validateUpdateRequest(UserDTO user, UserDTO originalUser, User originalUserEntity)
 	{
 		if (originalUserEntity.isCreatedOnBuddyRequest())
```
Bert-R: Same here

PR #244, opened 2016-09-14
YD-363 Added buddy link to GoalChangeMessage

Bert-R:
And removed the DailyActivityReportsLink, as that is of no use.


aristotelos:
I don't agree that the DailyActivityReportsLink is of not use. See the [design ](http://wiki.yona.nu/display/DEV/Me+-+Notifications) where it says: "Tap --> go to Friend Day View".

Bert-R:
They implement that by going to the buddy and from there take the two links (for day and week).

aristotelos:
OK, if you have communicated it with Kinnar it is OK.

PR #246, opened 2016-09-20
YD-295 Using shared Hazelcast cache

aristotelos:
Used for caching user anonymized DTOs and activity category DTOs

DayActivity entity does not implement Serializable yet so the tests do not run yet!!


aristotelos:
Using local cache for cached DayActivity entities so that it will not serialize.

Bert-R:
Looks good!

I extended the activity category update tests to verify that the distributed cache is indeed updated, and that works nicely.

PR #248, opened 2016-09-27
YD-349 Added logging to trap the actual cause

aristotelos:


appservice/src/main/java/nu/yona/server/messaging/rest/MessageController.java
```diff
@@ -390,6 +394,19 @@ private void addUserLinkIfAvailable(BuddyMessageLinkedUserDTO buddyMessage)
 		private void addActivityCommentMessageLinks(ActivityCommentMessageDTO message)
 		{
 			IntervalActivity activity = IntervalActivity.getIntervalActivityRepository().findOne(message.getActivityID());
+			if (activity == null)
```
Bert-R: Don't think this logging should go here. If ``MessageControler.addLinks`` has an issue with null, that should do the checks. Besides this, the current code masks the issue.

I would add the below two lines to ``MessageControler.addLinks``:
```
			Objects.requireNonNull(activity, "Activity being commented should already exist");
			Objects.requireNonNull(activity.getGoal(), "Activity does not have a goal linked to it");
```
That way, we get an explicit error when these things are wrong, and the problem is not masked.

aristotelos: Done

Bert-R:
The format strings were wrong; corrected them.

PR #249, opened 2016-09-27
YD-352 Fixing LazyInitializationException for cached DayActivity

aristotelos:
I am still busy writing a reproduction SpringJUnit4ClassRunner integration test that:
- adds a DayActivity to the repository
- tries to retrieve it from the cache and update it

### DIFF
--- a/core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
+++ b/core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
@@ -15,6 +15,7 @@ import java.util.stream.Collectors;
 
 import javax.persistence.CascadeType;
 import javax.persistence.Entity;
+import javax.persistence.FetchType;
 import javax.persistence.OneToMany;
 import javax.persistence.Table;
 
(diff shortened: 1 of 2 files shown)


aristotelos:
The unit test is quite hard because you have to terminate a Hibernate session. I think I rather skip it.

core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
```diff
@@ -34,7 +35,11 @@ public static DayActivityRepository getRepository()
 		return (DayActivityRepository) RepositoryProvider.getRepository(DayActivity.class, UUID.class);
 	}
 
-	@OneToMany(cascade = CascadeType.ALL)
+	// because we are caching DayActivity instances and using this list, the fetch type has to be eager
+	// (or we should re-initialize a Hibernate session after returning from the cache)
+	// see http://jira.yona.nu/browse/YD-352 and
+	// http://stackoverflow.com/questions/26507446/how-to-resolve-lazyinitializationexception-in-spring-data-jpa
+	@OneToMany(cascade = CascadeType.ALL, fetch = FetchType.EAGER)
```
Bert-R: How certain are you that this works? I have tried this approach earlier and failed. That was inexplicable to me, but still...
Can we have a test case that reproduces the issue?

aristotelos: Am not sure. The test case is difficult. Will proceed with refactoring DayActivity entity caching...

PR #250, opened 2016-10-01
YD-295 Fix concurrency issues in activity creation

Bert-R:
* The activity reports are now solely based on DTOs
* After creating the report, a request is done on the analysis service to have that create the entities for the inactive days
* The analysis engine service implements locking to ensure only one thread updates a user at a time. This locking is currently only in place for creation of day and week activities.


analysisservice/src/main/java/nu/yona/server/analysis/rest/AnalysisEngineController.java
```diff
@@ -4,34 +4,50 @@
...
 @Controller
-@RequestMapping(value = "/analysisEngine", produces = { MediaType.APPLICATION_JSON_VALUE })
+@RequestMapping(value = "", produces = { MediaType.APPLICATION_JSON_VALUE })
 public class AnalysisEngineController
 {
 	@Autowired
 	private AnalysisEngineService analysisEngineService;
 
-	@RequestMapping(value = "/", method = RequestMethod.POST)
+	@RequestMapping(value = "userAnonymized/{userAnonymizedID}/networkActivity/", method = RequestMethod.POST)
```
aristotelos: For consistency, start path with a slash

core/src/main/java/nu/yona/server/analysis/service/IntervalInactivity.java
```diff
@@ -0,0 +1,75 @@
...
+import java.time.temporal.ChronoUnit;
+import java.util.Optional;
+import java.util.UUID;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class IntervalInactivity
```
aristotelos: Postfix with DTO for consistency

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
```diff
@@ -73,50 +80,88 @@
...
+
+		return retVal;
+	}
+
+	private void createInactivityEntities(Set<IntervalInactivity> missingInactivities)
+	{
+		Map<UUID, Set<IntervalInactivity>> activitiesByUserAnonymizedID = missingInactivities.stream()
+				.collect(Collectors.groupingBy(mia -> mia.getUserAnonymizedID().get(), Collectors.toSet()));
+
+		activitiesByUserAnonymizedID.forEach((u, mias) -> analysisEngineProxyService.createInactivityEntities(u, mias));
```
aristotelos: For consistency rename `mias` to `mia` :)

core/src/main/java/nu/yona/server/analysis/service/WeekActivityDTO.java
```diff
@@ -84,4 +98,63 @@ static WeekActivityDTO createInstance(WeekActivity weekActivity, LevelOfDetail l
 								dayActivity -> DayActivityDTO.createInstance(dayActivity, levelOfDetail))),
 				weekActivity.hasPrevious(), weekActivity.hasNext());
 	}
+
+	public static WeekActivityDTO createInstanceInactivity(UserAnonymizedDTO userAnonymized, Goal goal, ZonedDateTime startOfWeek,
```
aristotelos: I see you moved this logic from the entity to the DTO and pass the `missingInactivities` set to modify it. Is it in the end not cleaner to move this logic out of the DTO to the `ActivityService` class?

core/src/main/java/nu/yona/server/CoreConfiguration.java
```diff
@@ -122,6 +125,7 @@ ObjectMapper objectMapper()
 		ObjectMapper springHateoasObjectMapper = beanFactory.getBean(SPRING_HATEOAS_OBJECT_MAPPER, ObjectMapper.class);
 		Jackson2ObjectMapperBuilder builder = beanFactory.getBean(Jackson2ObjectMapperBuilder.class);
 		builder.configure(springHateoasObjectMapper);
+		springHateoasObjectMapper.disable(DeserializationFeature.ADJUST_DATES_TO_CONTEXT_TIME_ZONE);
```
aristotelos: Can you add a comment why this was disabled?

core/src/main/java/nu/yona/server/rest/RestClientErrorHandler.java
```diff
@@ -0,0 +1,32 @@
...
+	private static final Logger logger = LoggerFactory.getLogger(RestClientErrorHandler.class);
+
+	@Override
+	public void handleError(ClientHttpResponse response) throws IOException
+	{
+		logger.error("Response error: {} {}", response.getStatusCode(), response.getStatusText());
+	}
+
+	@Override
+	public boolean hasError(ClientHttpResponse response) throws IOException
```
aristotelos: Why do you need to override this? Could you add a comment?

core/src/main/java/nu/yona/server/rest/RestClientErrorHandler.java
```diff
@@ -0,0 +1,32 @@
...
+package nu.yona.server.rest;
+
+import java.io.IOException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.client.ClientHttpResponse;
+import org.springframework.web.client.ResponseErrorHandler;
+
+public class RestClientErrorHandler implements ResponseErrorHandler
```
aristotelos: Can you add a comment what this class does? It seems to log errors when they happen.

core/src/main/java/nu/yona/server/analysis/service/DayActivityDTO.java
```diff
@@ -27,11 +41,14 @@
 	private final int totalMinutesBeyondGoal;
 	private final Set<MessageDTO> messages;
 
-	private DayActivityDTO(UUID goalID, ZonedDateTime startTime, boolean shouldSerializeDate, List<Integer> spread,
-			Optional<Integer> totalActivityDurationMinutes, boolean goalAccomplished, int totalMinutesBeyondGoal,
+	private final UUID activityCategoryID;
```
aristotelos: Why is this only in `DayActivityDTO` and not in `WeekActivityDTO`? It seems something generic to put into `IntervalActivityDTO`.

analysisservice/src/main/java/nu/yona/server/analysis/service/UserAnonymizedSynchronizer.java
```diff
@@ -0,0 +1,97 @@
...
+ *******************************************************************************/
+package nu.yona.server.analysis.service;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import nu.yona.server.exceptions.YonaException;
+
+public class UserAnonymizedSynchronizer
```
aristotelos: It would be great to have a unit test for this class.

analysisservice/src/main/java/nu/yona/server/analysis/service/UserAnonymizedSynchronizer.java
```diff
@@ -0,0 +1,97 @@
...
+			throw new IllegalStateException("Map already contains a locking thread for user anonymized ID " + userAnonymizedID);
+		}
+	}
+
+	private void unlock(UUID userAnonymizedID)
+	{
+		synchronized (lockedUsers)
+		{
+			lockedUsers.remove(userAnonymizedID);
+			lockedUsers.notifyAll();
```
aristotelos: Should it not be `notify`? Could it otherwise not be the case that two threads call `storeLock` at once after the `lockedUsers.wait` call?

aristotelos: Sorry, I really am not acquainted with the Java mechanism of locking, so please bear with my ignorance. 
- The `synchronized` keyword is adding a lock to a single object, so why do you still need `wait` and `notifyAll` calls?
- Why not use the `synchronized` keyword on the method instead of on the `lockedUsers` object? I think that is practically the same, isn't it?

analysisservice/src/main/java/nu/yona/server/analysis/rest/AnalysisEngineController.java
Bert-R: That was a mistake indeed. Corrected it.

core/src/main/java/nu/yona/server/analysis/service/IntervalInactivity.java
Bert-R: Done.

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
Bert-R: The plural name is correct here. It's a set of "missing inactivities".

core/src/main/java/nu/yona/server/analysis/service/WeekActivityDTO.java
Bert-R: That's not a bad idea, but I prefer to postpone this now and take that along when further refactoring ``ActivityService`` (I'm still unhappy with that class).

core/src/main/java/nu/yona/server/CoreConfiguration.java
Bert-R: Done.

core/src/main/java/nu/yona/server/rest/RestClientErrorHandler.java
Bert-R: Good that you point at this. Initially, it was only there to allow the callers of the REST template to handle the errors themselves. That lead to duplicate code, as every caller needed to do the same. Now I've moved the error handling code into this handler.

core/src/main/java/nu/yona/server/rest/RestClientErrorHandler.java
Bert-R: It's one of the methods of the implemented interface. Explaining that in a comment sounds like overkill to me.

core/src/main/java/nu/yona/server/analysis/service/DayActivityDTO.java
Bert-R: It's here to be able to convert from``DayActivityDTO`` to ``DayActivityWithBuddiesDTO``. As there is no ``WeekActivityWithBuddiesDTO``, this is not needed for its week counterpart.

analysisservice/src/main/java/nu/yona/server/analysis/service/UserAnonymizedSynchronizer.java
Bert-R: Let me talk you through this next week, that'll be a lot quicker than through comments here.

Bert-R:
Pushed an update that addresses the review comments.

PR #251, opened 2016-10-08
YD-370 Generalized LockPool and added unit test

Bert-R:
* Renamed UserAnonymizedSynchronizer into LockPool
* Moved it into nu.yona.util, inside Core
* Made it generic, with the ID type as type parameter
* Added a unit test that verifies:
 * Nonconcurrency when using the lock in a nonreentrant way
 * Nonconcurrency when using the lock in a reentrant way
 * Immediate access when using different keys

### DIFF
--- a/core/src/test/java/nu/yona/server/util/LockPoolTest.java
+++ b/core/src/test/java/nu/yona/server/util/LockPoolTest.java
@@ -1,0 +1,165 @@
+package nu.yona.server.util;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.junit.Assert.assertThat;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.junit.Test;
+
+import nu.yona.server.exceptions.YonaException;
+
+public class LockPoolTest
+{
+	private final AtomicInteger freeLockID = new AtomicInteger(0);
+
+	@Test
+	public void testLockingNonReentrant()
+	{
+		final int numThreads = 10;
+		final int iterations = 10000;
+		assertThat(testConcurrently(numThreads, iterations, this::attemptNonReentrantAccess), equalTo(0));
+	}
+
+	@Test
+	public void testLockingReentrant()
+	{
+		final int numThreads = 10;
+		final int iterations = 10000;
+		assertThat(testConcurrently(numThreads, iterations, this::attemptReentrantAccess), equalTo(0));
+	}
+
+	@Test
+	public void testConcurrencyDifferentKeys()
+	{
+		final int numThreads = 10;
+		final int iterations = 5;
+		assertThat(testConcurrently(numThreads, iterations, this::attempImmediateAccessRandomID), equalTo(0));
+	}
+
+	private int testConcurrently(int numThreads, int iterations, AccessAttempt accessAttempt)
+	{
+		LockPool<Integer> lockPool = new LockPool<>();
+		ExecutorService threadPool = Executors.newFixedThreadPool(numThreads);
+		CountDownLatch startSignal = new CountDownLatch(1);
+		CountDownLatch doneSignal = new CountDownLatch(numThreads);
+		AtomicBoolean concurrencyIndicator = new AtomicBoolean(false);
+		AtomicInteger failureCounter = new AtomicInteger(0);
+		for (int i = 0; (i < numThreads); i++)
+		{
+			threadPool.execute(() -> tryAccess(iterations, startSignal, lockPool, concurrencyIndicator, failureCounter,
+					doneSignal, accessAttempt));
+		}
+		startSignal.countDown();
+		awaitWithoutInterrupt(doneSignal);
+		threadPool.shutdown();
+		return failureCounter.get();
+	}
+
+	private void tryAccess(int iterations, CountDownLatch startSignal, LockPool<Integer> lockPool,
+			AtomicBoolean concurrencyIndicator, AtomicInteger failureCounter, CountDownLatch doneSignal,
+			AccessAttempt accessAtempt)
+	{
+		awaitWithoutInterrupt(startSignal);
+		for (int i = 0; (i < iterations); i++)
+		{
+			accessAtempt.attempt(lockPool, concurrencyIndicator, failureCounter);
+		}
+		doneSignal.countDown();
+	}
+
+	private void attemptNonReentrantAccess(LockPool<Integer> lockPool, AtomicBoolean concurrencyIndicator,
+			AtomicInteger failureCounter)
+	{
+		int lockID = 0;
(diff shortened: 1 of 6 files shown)


core/src/test/java/nu/yona/server/util/LockPoolTest.java
```diff
@@ -0,0 +1,165 @@
...
+		}
+		startSignal.countDown();
+		awaitWithoutInterrupt(doneSignal);
+		threadPool.shutdown();
+		return failureCounter.get();
+	}
+
+	private void tryAccess(int iterations, CountDownLatch startSignal, LockPool<Integer> lockPool,
+			AtomicBoolean concurrencyIndicator, AtomicInteger failureCounter, CountDownLatch doneSignal,
+			AccessAttempt accessAtempt)
```
aristotelos: Small spelling error `accessAtempt`

core/src/test/java/nu/yona/server/util/LockPoolTest.java
```diff
@@ -0,0 +1,165 @@
...
+		{
+			verifyNonConcurrentExecution(concurrencyIndicator, failureCounter);
+		}
+	}
+
+	private void attemptReentrantAccess(LockPool<Integer> lockPool, AtomicBoolean concurrencyIndicator,
+			AtomicInteger failureCounter)
+	{
+		int lockID = 0;
+		try (LockPool<Integer>.Lock lock1 = lockPool.lock(0))
```
aristotelos: Here, `0` should be substituted with `lockID` to show the intent of the method I guess.

core/src/test/java/nu/yona/server/util/LockPoolTest.java
```diff
@@ -0,0 +1,165 @@
...
+	private void attempImmediateAccessRandomID(LockPool<Integer> lockPool, AtomicBoolean concurrencyIndicator,
+			AtomicInteger failureCounter)
+	{
+		int lockID = freeLockID.incrementAndGet();
+		int sleepDuration = 500;
+		long startTime = System.currentTimeMillis();
+		long accessTime;
+		try (LockPool<Integer>.Lock lock1 = lockPool.lock(lockID))
+		{
+			try (LockPool<Integer>.Lock lock2 = lockPool.lock(lockID))
```
aristotelos: Why do you use two locks here? I think 1 would be anough to test this scenario?

core/src/test/java/nu/yona/server/util/LockPoolTest.java
```diff
@@ -0,0 +1,165 @@
...
+			{
+				accessTime = System.currentTimeMillis();
+				sleepWithoutInterrupt(sleepDuration);
+			}
+		}
+		if (accessTime - startTime > sleepDuration / 2)
```
aristotelos: I would say: if it is larger or equal to `sleepDuration`, it seems that some thread blocked you.

core/src/test/java/nu/yona/server/util/LockPoolTest.java
```diff
@@ -0,0 +1,165 @@
...
+			failureCounter.incrementAndGet();
+		}
+	}
+
+	private void verifyNonConcurrentExecution(AtomicBoolean concurrencyIndicator, AtomicInteger failureCounter)
+	{
+		if (concurrencyIndicator.compareAndSet(false, true))
+		{
+			// No one else is here
+			concurrencyIndicator.set(false);
```
aristotelos: For easier detection, you could add a sleep of 1 millisecond before resetting it to `false` I think.

core/src/test/java/nu/yona/server/util/LockPoolTest.java
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	private int testConcurrently(int numThreads, int iterations, AccessAttempt accessAttempt)
+	{
+		LockPool<Integer> lockPool = new LockPool<>();
+		ExecutorService threadPool = Executors.newFixedThreadPool(numThreads);
+		CountDownLatch startSignal = new CountDownLatch(1);
+		CountDownLatch doneSignal = new CountDownLatch(numThreads);
+		AtomicBoolean concurrencyIndicator = new AtomicBoolean(false);
+		AtomicInteger failureCounter = new AtomicInteger(0);
```
aristotelos: Because the `failureCounter` is incremented at multiple places, it would be easier for test fixing to use a failure log which you push a message to. It doesn't matter much for the completeness of the test though.

aristotelos:
Great, thanks for the test!

core/src/test/java/nu/yona/server/util/LockPoolTest.java
Bert-R: I made it numerous times. One survived. Fixed it.

core/src/test/java/nu/yona/server/util/LockPoolTest.java
Bert-R: Correct. Fixed it.

core/src/test/java/nu/yona/server/util/LockPoolTest.java
Bert-R: This is reentrance. You'll never see this as straight as you see it here, but if two methods lock on the user anonymized ID, where one method calls the other, it would cause a deadlock of reentrance wasn't supported.

core/src/test/java/nu/yona/server/util/LockPoolTest.java
Bert-R: The contract for sleep is this: 
> Causes the currently executing thread to sleep (temporarily cease execution) for the specified number of milliseconds, subject to the precision and accuracy of system timers and schedulers. The thread does not lose ownership of any monitors.

Now I allow an accuracy of 50%. I'll add an explanatory comment.

core/src/test/java/nu/yona/server/util/LockPoolTest.java
Bert-R: Yes, I've thought of that and didn't like it because it slows down the test. When it's not necessary, I don't like to add sleeps. At the other hand, 1ms is very little. I'll give it a try and see what happens

Bert-R: The probability increases a lot and with an iteration count of 100, it still fails in every run if I comment a few lines in the ``LockPool``. So I've added a sleep.

Bert-R:
Processed the review comments.

Bert-R:
I've merged this myself, so it's part of the build that I'm about to deliver. Hope that's fine. Feel free to add additional review comments for a next build.

core/src/test/java/nu/yona/server/util/LockPoolTest.java
aristotelos: OK, so you're testing two things at once. I think that is a bit unnecessary but it doesn't harm much.

core/src/test/java/nu/yona/server/util/LockPoolTest.java
aristotelos: Hey, that's going the wrong way! ;)

aristotelos:
Thanks for the changes, I would have merged it now.

PR #253, opened 2016-10-12
YD-369 Enable activity category administration

Bert-R:
The ActivityCategoryController in the admin service now supports add, update and delete of activity categories.


adminservice/src/intTest/groovy/nu/yona/server/LocalizationTest.groovy
```diff
@@ -1,56 +0,0 @@
...
- * v.2.0. If a copy of the MPL was not distributed with this file, You can
- * obtain one at https://mozilla.org/MPL/2.0/.
- *******************************************************************************/
-package nu.yona.server
-
-import groovy.json.*
-import spock.lang.Shared
-import spock.lang.Specification
-
-class LocalizationTest extends Specification
```
Bert-R: Given that the admin service now returns a localizable name rather than a localized one, this test does not make sense anymore. There is still another ``LocalizationTest`` for the app service that test the localization there.

core/src/main/java/nu/yona/server/goals/service/ActivityCategoryService.java
```diff
@@ -53,18 +59,41 @@ public ActivityCategoryDTO getActivityCategory(UUID id)
 	@Transactional
 	public ActivityCategoryDTO addActivityCategory(ActivityCategoryDTO activityCategoryDTO)
 	{
-		logger.info("Adding activity category '{}'", activityCategoryDTO.getName(Translator.EN_US_LOCALE));
+		logger.info("Adding activity category '{}' with ID '{}'", activityCategoryDTO.getName(Translator.EN_US_LOCALE),
+				activityCategoryDTO.getID());
+		verifyNoDuplicateNames(Collections.emptySet(), activityCategoryDTO.getLocalizableNameByLocale());
```
aristotelos: How about duplicate Smoothwall categories, applications, IDs?

core/src/test/java/nu/yona/server/goals/service/ActivityCategoryServiceIntegrationTest.java
```diff
@@ -62,14 +62,19 @@ public void caching()
...
 				new HashSet<String>(Arrays.asList("games")), Collections.emptySet());
+		// Add to collection, so it appears as if it is in the database
+		// It hasn't been loaded, so should not be in the cache
 		activityCategories.add(gaming);
 		when(mockRepository.findOne(gaming.getID())).thenReturn(gaming);
 
 		assertGetAllActivityCategoriesResult("Set expected to be cached", "gambling", "news");
 
-		service.addActivityCategory(ActivityCategoryDTO.createInstance(gaming));
+		// Add dummy activity category to trigger cache eviction
```
aristotelos: I don't understand this. Why are you adding a dummy but don't mock it from the database?

adminservice/src/intTest/groovy/nu/yona/server/ActivityCategoriesTest.groovy
```diff
@@ -27,5 +27,158 @@ class ActivityCategoriesTest extends Specification
...
+		String programmingActivityCategoryJson = createActivityCategoryJson(["nl-NL": "Zomaar wat", "en-US" : "Gambling"], false, ["programming", "scripting"], ["Eclipse", "Visual Studio"])
+		when:
+		def response = adminService.yonaServer.createResource(AdminService.ACTIVITY_CATEGORIES_PATH, programmingActivityCategoryJson)
+
+		then:
+		response.status == 400
+		response.responseData.code == "error.activitycategory.duplicate.name"
+	}
+
+	def 'Try update to duplicate Dutch name' ()
```
Bert-R: This is a different test. The previous one tries to **add** a duplicate name while the second one tries to **update** an existing causing a duplicate name. The first one uses English and the second Dutch.

adminservice/src/main/java/nu/yona/server/AdminServiceApplication.java
```diff
@@ -32,4 +38,18 @@ protected SpringApplicationBuilder configure(SpringApplicationBuilder applicatio
...
 	}
+
+	@Bean
+	@Primary
+	public PlatformTransactionManager txManager(EntityManagerFactory entityManagerFactory)
+	{
+		// This transaction manager replaces the one set by Spring Batch Admin
+		// (org.springframework.jdbc.datasource.DataSourceTransactionManager).
+		// With the DataSourceTransactionManager, JPA updates were not persisted.
+		// TODO: Verify wither the Spring Batch Admin functionality still works.
```
Bert-R: Yes, and honestly, I don't care that much at the moment.

core/src/main/java/nu/yona/server/goals/service/ActivityCategoryDTO.java
```diff
@@ -106,13 +144,15 @@ public static ActivityCategoryDTO createInstance(ActivityCategory activityCatego
 
 	public ActivityCategory createActivityCategoryEntity()
```
Bert-R: No, it's not a getter, so it'll anyway be ignored.

adminservice/src/main/java/nu/yona/server/goals/rest/ActivityCategoryController.java
```diff
@@ -40,18 +44,45 @@
...
 	{
 		return createOKResponse(activityCategoryService.getAllActivityCategories(), getAllActivityCategoriesLinkBuilder());
 	}
 
+	@RequestMapping(value = "/", method = RequestMethod.POST)
+	@ResponseBody
+	@JsonView(ActivityCategoryDTO.AdminView.class)
+	public HttpEntity<ActivityCategoryResource> addActivityCategory(@RequestBody ActivityCategoryDTO activityCategory)
+	{
+		activityCategory.setID(UUID.randomUUID());
```
Bert-R: Given that it's a DTO, I'm not concerned about the public availability of the setter. Here, you _might_ move it to the service, but on the update, it has to be in the controller (at least, I feel it's a weird service contract to pass a DTO containing an ID and then the same ID in a parameter). Given that, I favor keeping this set also in the controller.

core/src/main/java/nu/yona/server/goals/service/ActivityCategoryService.java
Bert-R: That's not a problem. Smoothwall already classifies sites under multiple categories, so if you do things that fall under multiple categories, you see activities under multiple categories. Summing up the activities might go beyond 100%

adminservice/src/intTest/groovy/nu/yona/server/ActivityCategoriesTest.groovy
aristotelos: OK, I see now

core/src/test/java/nu/yona/server/goals/service/ActivityCategoryServiceIntegrationTest.java
Bert-R: Adding it and mocking it is doing it twice. Earlier, we were mocking the gaming category and then adding it. Now we check for duplicates and don't accept the "add" anymore.

The test now mocks "gaming" in the database and then adds "dummy". The addition of "dummy" triggers cache eviction and due to that, the "gaming" category is read from the mocked database.

Bert-R: This test is to verify that cache eviction works when a new category is added, and that's what the test still does. We don't want to verify here that save works.

PR #254, opened 2016-10-13
YD-271 Added localizable description to activity category

Bert-R:
Along with this added a new activity category: Gambling.

**Note**: this pull request is based on #253. Once that one is merged, built and delivered, this one can be merged. Not earlier, as that would make it impossible to bring YD-279 separately into the acceptance test.


aristotelos:
Reviewed, no suggestions for improvement. Will not merge because #253 has to be delivered first.

PR #256, opened 2016-10-22
YD-373 Implemented PUT for all act cats

Bert-R:
* Implemented a PUT option on ``/activityCategories/``. This allows setting the production activity categories through a command like this: ``curl -X PUT http://localhost:8080/activityCategories/ -d @dbinit/data/productionActivityCategories.json --header "Content-Type: application/json"``
* There is an issue in HSQLDB, Spring or JPA that makes an implicit flush fail with ``java.sql.SQLException: invalid transaction state: read-only SQL-transaction``. Implemented an explicit flush to work around that issue.
* Added an extra JSON file with the production activity categories
* Removed  ``Games`` from the activity categories file used for testing
* Reformatted the activity categories in the JSON file, to increase readability and simplify comparison
* Renamed local variable in ``ActivityCacheService.fetchLastActivityForUser`` for consistency


Bert-R:
Once this PR is merged, the functional people can maintain the activity categories on GitHub. Loading them into an environment can be done through this command:
```
curl https://raw.githubusercontent.com/yonadev/yona-server/master/dbinit/data/productionActivityCategories.json | curl -X PUT --header "Content-Type: application/json" -d @- http://localhost:8080/activityCategories/
```

PR #258, opened 2016-10-27
YD-361 YD-354 Correct goal deletion

Bert-R:
Removing a goal merely removed it from the collection of goals. The entity stayed in the system as a orphan
* YD-361: 500 when deleting a user for which goal conflict messages were sent. The goal conflict message prevented deletion of the goals through a foreign key constraint
* YD-354: After deletion of a goal, the goal conflict messages stayed in the system, but the related goals were not part of the user profile

The following changes were made:
* A ``GoalChangeMessage`` now associates with an ``ActivityCategory`` instead of a ``Goal``. This allows the goal to be deleted while the messages stay in the system.
* The ``creationTime`` and ``endTime`` on ``Goal`` are now stored as SQL dates (in UTC) rather than serialized ZonedDateTimes
* The ``LocalDateAttributeConverter`` is now marked as "auto-apply"
* Deletion of a ``Goal`` now also deletes the chain of ``previousInstanceOfThisGoal``, with the related messages and activities


core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
```diff
@@ -167,7 +165,7 @@ private void createAndSaveInactivityDays(UserAnonymizedDTO userAnonymized, Set<G
 	private long getTotalPageableItems(UserAnonymizedDTO userAnonymized, ChronoUnit timeUnit)
 	{
 		long activityMemoryDays = yonaProperties.getAnalysisService().getActivityMemory().toDays();
-		Optional<ZonedDateTime> oldestGoalCreationTime = userAnonymized.getOldestGoalCreationTime();
+		Optional<LocalDateTime> oldestGoalCreationTime = userAnonymized.getOldestGoalCreationTime();
 		long activityRecordedDays = oldestGoalCreationTime.isPresent() ? (Duration
 				.between(oldestGoalCreationTime.get(), ZonedDateTime.now(ZoneId.of(userAnonymized.getTimeZoneId()))).toDays() + 1)
```
aristotelos: Does the `Duration.between` convert the second argument `ZonedDateTime` to local date time in UTC correctly? I am not completely sure from the documentation...

aristotelos:
Looks great, I hope this solves a lot of problems.

You said: "A `GoalConflictMessage` now associates with an `ActivityCategory` instead of a `Goal`." You of course meant a `GoalChangeMessage`, as `GoalConflictMessages` are deleted at removal of the goal.

You could still create an integration test that if you remove a goal, it also removes the chain of previous instances.

Bert-R:
> You said: "A `GoalConflictMessage` now associates with an `ActivityCategory` instead of a `Goal`." You of course meant a `GoalChangeMessage`, as `GoalConflictMessages` are deleted at removal of the goal.

Indeed. Corrected that comment.

Bert-R:
> You could still create an integration test that if you remove a goal, it also removes the chain of previous instances.

That's a good one. Will try to do that today or tomorrow.

Bert-R:
Created the test and fixed the bug that was found through the test. Good that you asked for the test!

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
Bert-R: Yes, I noticed it last night and made a fix, but didn't wait for the test results. These are positive, so I've pushed the update now.

PR #266, opened 2016-10-31
YD-261 Fixed delete failure after user overwrite

Bert-R:
The solution comprises of the following as part of ``UserService.deleteUser``:
* Process any pending buddy response messages, to ensure a standard state
* Handle buddy users removed while offline
* When handling buddy users removed while offline, also remove the messages sent by this user to the user that was removed while offline

Besides this:
* Fixed DST issues
* Added missing clean-ups

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
@@ -241,6 +241,12 @@ public class BuddyService
 		return messageToBeProcessed.isPresent();
 	}
 
+	public void processPossiblePendingBuddyResponseMessages(User userEntity)
+	{
+		getBuddyEntitiesOfUser(userEntity.getID()).stream()
+				.forEach(b -> processPossiblePendingBuddyResponseMessage(userEntity, b));
+	}
+
 	private Optional<UUID> getUserID(BuddyConnectResponseMessage message)
 	{
 		return message.getSenderUser().map(u -> u.getID());
@@ -279,6 +285,7 @@ public class BuddyService
 		{
 			// Buddy request was accepted
 			removeMessagesSentByBuddy(user, buddy);
+			removeMessagesSentByUserToBuddy(user, buddy);
 
 			// Send message to "self", to notify the user about the removed buddy user
 			UUID buddyUserAnonymizedID = getUserAnonymizedIDForBuddy(buddy);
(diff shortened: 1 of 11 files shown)


core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -279,6 +285,7 @@ void removeBuddyInfoForRemovedUser(User user, Buddy buddy)
 		{
 			// Buddy request was accepted
 			removeMessagesSentByBuddy(user, buddy);
+			removeMessagesSentByUserToBuddy(user, buddy);
```
aristotelos: For my understanding, is this only required if the user was removed with `UserService.deleteExistingUserToOverwriteIt`? I see that the user at removal at `UserService.deleteUser` already removes the named and anonymous messages in `removeBuddyInfoForBuddy`.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -337,16 +344,34 @@ public void setBuddyAcceptedWithSecretUserInfo(UUID buddyID, UUID userAnonymized
...
 		UserAnonymizedDTO userAnonymized = userAnonymizedService.getUserAnonymized(user.getUserAnonymizedID());
-		removeAnonymousMessagesSentByUser(userAnonymized, userAnonymizedID.get());
+		removeAnonymousMessagesSentByUser(userAnonymized, buddyUserAnonymizedID.get());
+	}
+
+	private void removeMessagesSentByUserToBuddy(User user, Buddy buddy)
+	{
+		UUID buddyUserAnonymizedID = getUserAnonymizedIDForBuddy(buddy);
+		UserAnonymizedDTO buddyUserAnonymized = userAnonymizedService.getUserAnonymized(buddyUserAnonymizedID);
+		removeAnonymousMessagesSentByUser(buddyUserAnonymized, user.getUserAnonymizedID());
```
aristotelos: Why only the anonymous messages? I would expect the named messages too.

aristotelos: It would be great to include that in an integration test.

Bert-R: The integration tests cover this. It was actually the first cause of a test failure once I added the assertion on deletes in the cleanup.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
Bert-R: Yes, that's the case. The failing test was ``OverwriteUserTest.'Overwrite Richard while being a buddy of Bob and verify that Bob can move on with everything'``. The other one that failed is ``OverwriteUserTest.'Overwrite Richard while being a buddy of Bob and verify Richard does not receive Bob\'s goal conflicts anymore'``

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
Bert-R: > Why only the anonymous messages? I would expect the named messages too.

Added an explanatory comment:
```
// We are not removing the named messages because we don't have User entity anymore
// (this method is being called from removeBuddyInfoForRemovedUser) and thus we don't know the named destination.
// Given that the user and the named destination are both removed, this is not causing any issues.
```

Bert-R:
Responded to the comments and added a commit with an extra comment.

PR #267, opened 2016-11-03
YD-376 Corrected database schema

Bert-R:
Before we can go in production, the database schema needs to be right.

Made the following changes:
* Java types are never serialized to be stored
* Do not store ``ZonedDateTime`` but store ``LocalDateTime`` or ``LocalDate`` and add a separate time zone attribute in cases where it's needed. In other cases, store the time in UTC.
* Marked all UUID properties with uuid-char, to prevent binary storage
* Smaller clean up things


Bert-R:
Merged it myself, as I need to move on with the schema changes.

PR #270, opened 2016-11-14
YD-377 Database schema evolution

Bert-R:
This PR brings database schema evolution based on Liquibase. This includes several significant changes:
* The database schema is not created by running the ``main`` of ``dbinit`` but by running Liquibase update and loading the activity categories through ``curl``.
* Moved to another Gradle plug in for docker. The original one (``se.transmode.gradle:gradle-docker``) is not maintained anymore and didn't work with Docker on Windows. The new one (``com.bmuschko:gradle-docker-plugin``) is under active maintenance and works very well on Windows.
* Created a new Docker image to run Liquibase. This is based on another new image ([yonadev/hsqldb-liquibase-update](https://github.com/yonadev/hsqldb-liquibase-update)).
* Upgraded the Yona Docker compose file to version 2
* Created a new Docker compose file to run HSQLDB
* The Yona services and HSQLDB now run on the ``yonanet`` network
* The procedure to create a schema update is yet to be documented. I plan to do that after merging this PR, based on a small update that for instance adds a creation time to one of the entities that still needs one.


adminservice/src/main/java/nu/yona/server/AdminServiceApplication.java
```diff
@@ -30,7 +32,19 @@
...
-		SpringApplication.run(AdminServiceApplication.class, args);
+		try
+		{
+			SpringApplication.run(AdminServiceApplication.class, args);
+		}
+		catch (Exception ex)
+		{
+
+			// issue in Hazelcast: it doesn't shutdown automatically,
+			// while we want this for the short running database initializer
```
aristotelos: Copy-paste error, this is not about the database initializer

analysisservice/src/main/java/nu/yona/server/AnalysisServiceApplication.java
```diff
@@ -21,7 +23,19 @@
...
-		SpringApplication.run(AnalysisServiceApplication.class, args);
+		try
+		{
+			SpringApplication.run(AnalysisServiceApplication.class, args);
+		}
+		catch (Exception ex)
+		{
+
+			// issue in Hazelcast: it doesn't shutdown automatically,
+			// while we want this for the short running database initializer
```
aristotelos: Copy-paste error, this is not about the database initializer

appservice/src/main/java/nu/yona/server/AppServiceApplication.java
```diff
@@ -26,7 +28,19 @@
...
-		SpringApplication.run(AppServiceApplication.class, args);
+		try
+		{
+			SpringApplication.run(AppServiceApplication.class, args);
+		}
+		catch (Exception ex)
+		{
+
+			// issue in Hazelcast: it doesn't shutdown automatically,
+			// while we want this for the short running database initializer
```
aristotelos: Copy-paste error, this is not about the database initializer

adminservice/src/main/java/nu/yona/server/AdminServiceApplication.java
aristotelos: Is this only on exceptions? Should it not be in a finally clause?

batchservice/src/main/java/nu/yona/server/BatchServiceApplication.java
```diff
@@ -23,7 +25,19 @@
...
-		SpringApplication.run(BatchServiceApplication.class, args);
+		try
+		{
+			SpringApplication.run(BatchServiceApplication.class, args);
+		}
+		catch (Exception ex)
+		{
+
+			// issue in Hazelcast: it doesn't shutdown automatically,
+			// while we want this for the short running database initializer
```
aristotelos: Same here

dbinit/liquibase.gradle
```diff
@@ -0,0 +1,78 @@
...
+applicationProps.load(new FileInputStream("core/src/main/resources/application.properties"))
+
+String databaseUrl = applicationProps.getProperty('spring.datasource.url')
+String hibernateUrl = "hibernate:spring:" + liquibaseProps.getProperty('liquibase.domain.package') + "?dialect=" + applicationProps.getProperty('spring.jpa.database-platform') + "&hibernate.ejb.naming_strategy=" + applicationProps.getProperty('spring.jpa.hibernate.naming_strategy')
+String userName = applicationProps.getProperty('spring.datasource.username')
+String password = applicationProps.getProperty('spring.datasource.password')
+
+String changeLogPath = liquibaseProps.getProperty('liquibase.changelog.path')
+
+task liquibaseGenerateChangeLogFromCode(type: JavaExec) {
```
aristotelos: When are those tasks called?

dbinit/liquibase.gradle
```diff
@@ -0,0 +1,78 @@
...
+}
+
+task liquibaseGenerateChangeLogFromDB(type: JavaExec) {
+  group = "liquibase"
+
+  classpath sourceSets.main.runtimeClasspath
+  classpath configurations.liquibase
+  main = "liquibase.integration.commandline.Main"
+
+  args "--changeLogFile=${changeLogPath}/base/batch.yml"
```
aristotelos: So these batch tables are the only content of the database?

adminservice/src/main/java/nu/yona/server/AdminServiceApplication.java
Bert-R: Corrected it. It's only for exceptions. If the startup is successful, the call to ``SpringApplication.run`` will return, so if we would put the Hazelcast shutdown in a finally block, we would shut down Hazel cast immediately after service startup.

Hopefully, they fix hazelcast/hazelcast#6339 quickly. I will see whether I can create a little repro for them.

analysisservice/src/main/java/nu/yona/server/AnalysisServiceApplication.java
Bert-R: Corrected it.

appservice/src/main/java/nu/yona/server/AppServiceApplication.java
Bert-R: Corrected it.

batchservice/src/main/java/nu/yona/server/BatchServiceApplication.java
Bert-R: Corrected it.

dbinit/liquibase.gradle
Bert-R: I will create a wiki page to describe the process around Liquibase, including the purpose of these tasks. I will do that along with the first schema update.

dbinit/liquibase.gradle
Bert-R: No, there is more to it.
I'll describe that on the wiki page I mentioned.

Bert-R:
> It's hard for me to review because I am not familiar with Liquibase. What I think the purpose is:
>
> * Generate a database definition in a file (entity.yml and batch.yml, both included in base.yml)
> * Generate a diff file from the last database definition
> * This is done with every build?
> * The purpose is that you can update the database without dropping all existing content

That's quite right. More precisely: Liquibase allows updating a database from a definition file (the change log). This change log can be handcrafted or generated by Liquibase. In our case, I have generated the base schema files from the code (for the entities) and the database (for the Spring Batch schema) and plan to generate the incremental updates from the code.

> I hope that I have not got all wrong. But that leaves me with some questions:
> * Is the diff tool of Liquibase smart enough to create a minimal change set?

Pretty smart, though there is an annoying bug: liquibase/liquibase-hibernate#134. We will need to remove the drop/create statements manually from the generated change log, to prevent unnecessary overhead.

> * How are the applicable diffs applied when I e.g. merge master in my local Git feature branch (or when the Yona servers are updated)? When I run servers_start? Is this done automagically?

I'm going to describe that. The idea is that we manually generate the delta change log, put the file in the ``liquibase/updates`` folder and include it inside ``liquibase/updates.yml`` (which does not exist yet). Inside ``servers_start``, I've put a ``liquibase update`` statement that loads the base schema (we continue to drop the database before running our regression tests) and the updates that we created after that. Next, we load the activity categories through ``curl``.

> * What if you add a column that should not be NULL for existing data? Do you need to provide a manual update script?

That's possible through Liquibase.

> * Is the dbinit service still necessary? Or can we drop that?

It's not necessary anymore, but I want to retain it for some time, till we're confidently running with the new process.

PR #271, opened 2016-11-15
YD-379 Fixed introduced NullPointerException

aristotelos:
Exception could occur around 00:00 when trying to combine with the
previous activity.


Bert-R:
Good find!
Going to merge it.

PR #272, opened 2016-11-19
YD-227 Time stamp on BuddyAnonymized

Bert-R:
Added lastStatusChangeTime to BuddyAnonymized. It is updated when the buddy request is created, accepted, rejected or disconnected.

As this is the first change that requires a schema change, several other changes are included:
* Introduction of updates.yml, the file accumulating includes for all schema changes
* Corrected destination path in liquibase.gradle in task liquibaseDiffChangelog
* Removed the devices table from entity.yml. This shouldn't have been there as the Device class was removed already as part of PR #267. This was missed when merging these changes into the work for the Liquibase introduction.

Unfortunately, we always need to update the generated change log manually due to liquibase/liquibase-hibernate#134. This is included in the wiki procedure: http://wiki.yona.nu/display/DEV/Schema+evolution+with+liquibase

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/BasicBuddyTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/BasicBuddyTest.groovy
@@ -11,6 +11,7 @@ import groovy.json.*
 import java.time.Duration
 import java.time.ZonedDateTime
 
+import nu.yona.server.test.AppService
 import nu.yona.server.test.Buddy
 import nu.yona.server.test.Goal
 import nu.yona.server.test.User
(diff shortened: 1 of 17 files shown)


aristotelos:
Looks great. The only observation I have is that the new `lastStatusChangeTime` property is quite different from the existing `creationTime` properties. What were the exact requirements for adding this timestamp? Is it only the processing of the response messages? Should there also not still be a `creationTime` timestamp on a buddy?

Bert-R:
The idea was to be more informational. Knowing when the buddy link was requested but not when it was accepted isn't that meaningful. By making it a last status change time, you know when someone became your buddy, which is useful information.

Bert-R:
Addressed the merge conflicts

aristotelos:
OK, so it is a functional addition for the app to show this to the users.

Bert-R:
Indeed, in case they want to use this.

PR #273, opened 2016-11-19
YD-380 Greatly simplified the Gradle scripts

Bert-R:


dbinit/build.gradle
```diff
@@ -1,71 +1,48 @@
...
 }
 
 dependencies {
 	compile project(":core")
-	compile("org.springframework.boot:spring-boot-starter-data-jpa")
-	compile("org.springframework.batch:spring-batch-core:3.0.7.RELEASE")
-    compile("org.liquibase:liquibase-core:3.5.3")
-
-	runtime("org.hsqldb:hsqldb")
+	runtime "org.hsqldb:hsqldb"
```
aristotelos: Can't HSQLDB be included in the root build.gradle?

aristotelos:
Looks great, added a single question.

Bert-R:
Where? I don't see your question.

dbinit/build.gradle
aristotelos: OK, agreed

PR #274, opened 2016-11-22
YD-382 Simplify Hazelcast shutdown

Bert-R:
As pointed out by @wilkinsona on spring-projects/spring-boot#7418, our Hazelcast initialization was wrong. Spring couldn't see the Hazelcast bean and therefore couldn't shut it down. Corrected that now.


aristotelos:
Thanks for the fix. Do you know why the build is failing?

Bert-R:
> Do you know why the build is failing?

Yes, I've seen that multiple times for the pull request builder. I have no idea what's going wrong. Do you have an idea?

PR #275, opened 2016-11-24
YD-381 Upgrade to Spring Boot 1.4.2

Bert-R:
This includes:
* Upgrade of Jackson from 2.6.6 to 2.8.5
* Moving the unique constraint from IntervalActivity.java to extra.yml, as Liquibase and Hibernate couldn't agree on the names of the columns
* Use of org.hibernate:hibernate-java8. Now our attribute converters for LocalDate and LocalDateTime aren't necessary anymore. Will be deleted in a separate check-in
* Moved ZoneIdAttributeConverter from auto-apply to explicit apply, as Liquibase doesn't consider the autoapply attribute converters
* Removed @Table from DayActivity and WeekActivity, as they share the IntervalActivity table
* Changed the way the message source is touched to reencrypt it. This is necessary because of a change in Hibernate
* Now UserService.retrieveUserEncryptedData does not use a separate transaction anymore. This is not necessary and caused issues with the new version of Hibernate

Things to do:
* Remove unnecessary attribute converters (will be part of this PR)
* Move away from Velocity, as that is deprecated now
* Address a few warnings from Hibernate


core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -14,21 +14,21 @@
...
 import nu.yona.server.entities.EntityUtil;
 import nu.yona.server.entities.EntityWithID;
 import nu.yona.server.entities.RepositoryProvider;
+import nu.yona.server.entities.ZoneIdAttributeConverter;
 import nu.yona.server.goals.entities.Goal;
 import nu.yona.server.subscriptions.entities.UserAnonymized;
 
 @Entity
-@Table(name = "INTERVAL_ACTIVITIES", uniqueConstraints = {
-		@UniqueConstraint(columnNames = { "dtype", "user_anonymized", "startDate", "goal" }) })
```
aristotelos: Please refer with a comment that this unique constraint has been added to extra.yml

core/src/main/java/nu/yona/server/messaging/entities/MessageSource.java
```diff
@@ -57,7 +57,16 @@ public MessageSource(UUID id, PrivateKey privateKey, MessageDestination messageD
 	{
 		super(id);
 		this.messageDestination = messageDestination;
-		this.privateKeyBytes = PublicKeyUtil.privateKeyToBytes(privateKey);
+		this.privateKeyBytes = toAttributeArray(privateKey);
+	}
+
+	private static byte[] toAttributeArray(PrivateKey privateKey)
```
aristotelos: You could choose a more speaking name `toArrayWithModificationByte`

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -365,11 +365,10 @@ private EncryptedUserData retrieveUserEncryptedData(User originalUserEntity, Str
 	{
 		// use a separate transaction to read within the crypto session
```
aristotelos: Please update the comment

Bert-R:
Done.

PR #276, opened 2016-11-28
YD-383 Switched over to MariaDB

Bert-R:
This comes with the following:
* New Liquibase schema definition
* Changes in dependencies in Gradle
* Different properties for Spring Batch
* Different JDBC driver
* As ``change`` is a reserved word for MySQL, quoted the column name for ``change`` in ``GoalChangeMessage``
* Changed the transaction structure for "create user on buddy request" inside ``BuddyService`` and ``UserService``, to address a transaction isolation issue.
* Extended ``CryptoSession`` with the option to run a void method


core/src/main/java/nu/yona/server/goals/entities/GoalChangeMessage.java
```diff
@@ -22,6 +23,7 @@
 	@ManyToOne
 	private ActivityCategory activityCategoryOfChangedGoal;
 
+	@Column(name = "`change`")
```
aristotelos: Maybe add a comment that change is a reserved word in MySQL.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -112,32 +118,48 @@ static boolean canIncludePrivateData(Buddy buddyEntity)
...
 		logger.info(
 				"User with mobile number '{}' and ID '{}' sent buddy connect message to {} user with mobile number '{}' and ID '{}' as buddy",
-				requestingUser.getMobileNumber(), requestingUser.getID(), (buddyUserEntity == null) ? "new" : "existing",
+				requestingUser.getMobileNumber(), requestingUser.getID(), (buddyUserExists) ? "new" : "existing",
 				buddy.getUser().getMobileNumber(), buddy.getUser().getID());
 
 		return savedBuddy;
 	}
 
+	private void createBuddyUser(UUID idOfRequestingUser, BuddyDTO buddy, BiFunction<UUID, String, String> inviteURLGetter)
```
aristotelos: Maybe rename to `createAndInviteBuddyUser`

PR #279, opened 2016-12-10
Remove pairing tables

Bert-R:
* Removed the following tables:
  * interval_activities_activities
  * users_anonymized_goals (YD-388)
* This implied definition of foreign keys
* Along with this changed the way persistence is handled: more intensive use of CASCADE_ALL.
* Fixed an issue where we fetched the activity category information during the analysis of each and every app or network event


analysisservice/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -132,53 +158,61 @@ private void analyze(ActivityPayload payload, Set<ActivityCategoryDTO> matchingA
 		// because the lock is per user, it doesn't matter much that we block early.
 		try (LockPool<UUID>.Lock lock = userAnonymizedSynchronizer.lock(payload.userAnonymized.getID()))
 		{
-			analyzeInsideLock(payload, matchingActivityCategories);
+			UserAnonymizedEntityHolder userAnonymizedHolder = new UserAnonymizedEntityHolder(payload.userAnonymized.getID());
+			analyzeInsideLock(userAnonymizedHolder, payload, matchingActivityCategories);
+			if (userAnonymizedHolder.isEntityFetched())
```
aristotelos: Can't we rename this to `isEntityUpdated` or `isEntityDirty`? I think not all cases of when the entity is fetched are cases in which the entity is dirty, is it? The method `updateUserAnonymized` is sometimes called already.

analysisservice/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -132,53 +158,61 @@ private void analyze(ActivityPayload payload, Set<ActivityCategoryDTO> matchingA
...
+			{
+				userAnonymizedService.updateUserAnonymized(userAnonymizedHolder.getEntity());
+			}
 		}
 	}
 
-	private void analyzeInsideLock(ActivityPayload payload, Set<ActivityCategoryDTO> matchingActivityCategories)
+	private void analyzeInsideLock(UserAnonymizedEntityHolder userAnonymizedHolder, ActivityPayload payload,
```
aristotelos: This results in 1 extra `userAnonymizedHolder` argument per method. Can't the `userAnonymizedHolder` be included in the `ActivityPayload`?

dbinit/src/main/liquibase/base/extra.yml
```diff
@@ -1,9 +0,0 @@
-databaseChangeLog:
```
aristotelos: Where has this unique constraint gone?

scripts/cleanDB.sql
```diff
@@ -0,0 +1,23 @@
+delete from message_destinations_messages;
```
aristotelos: This script is never referenced. Is it complete? How do we ensure it stays complete?

analysisservice/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -41,13 +40,40 @@
 @Service
 public class AnalysisEngineService
 {
+	private class UserAnonymizedEntityHolder
```
aristotelos: Why is this an improvement? The `UserAnonymizedService` already caches the result.

core/src/main/java/nu/yona/server/goals/service/ActivityCategoryService.java
```diff
@@ -34,6 +34,33 @@
 @Service
 public class ActivityCategoryService
 {
+	/**
```
aristotelos: Hmm, that's a bad bug. Glad you found and fixed it...

analysisservice/src/test/java/nu/yona/server/analysis/service/AnalysisEngineServiceTests.java
```diff
@@ -70,13 +71,17 @@
 import nu.yona.server.util.TimeUtil;
 
 @RunWith(MockitoJUnitRunner.class)
+@Ignore // TODO YD-385 The day activities are now not directly saved to a repository, so this test can't work. Any idea how we'll
```
aristotelos: What is the problem exactly?

aristotelos:
It's really a big PR, I have not been able to review the last commit fully yet.

analysisservice/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
Bert-R: I bet you've seen the purpose when you looked further down the code. I've moved the inner class to the end, along with ``ActivityPayload``.

analysisservice/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
Bert-R: The issue is that we neither know nor need to know whether the entity is dirty. If it was fetched, we should save it and leave it to Hibernate whether or not it needs to go to the database.

dbinit/src/main/liquibase/base/extra.yml
Bert-R: Nowhere. That's a mistake. I removed it when I tried to implement a different foreign key, but after that failed, I forgot to bring the unique constraint back.

I'll do it.

scripts/cleanDB.sql
Bert-R: I run it manually. It is currently complete.
It's very convenient to see what data was exactly created by a single test: clean the DB and run the test.

analysisservice/src/test/java/nu/yona/server/analysis/service/AnalysisEngineServiceTests.java
Bert-R: Hibernate takes care of saving the activities, without using our repository.

aristotelos:
Hmm. I looked at the commit "Remove unnessecary database access" (
https://github.com/yonadev/yona-server/pull/279/commits/3061ca3a0bc38ac9c807e6056da6aede28483438)
but I think the UserAnonymizedHolder is not an improvement to that, as the
UserAnonymizedService already caches that entity. So the real reason is
that the activities pairing table is removed, so that the cached interval
activities at the user anonymized have to be reloaded, isn't it? Did this
not work properly before or is it affected by removing the relation between
intervalactivities and activities? Maybe you should explain this to me by
phone :).

In my opinion the code becomes more complex by introducing the
UserAnonymizedHolder class. The UserAnonymized entity is also still a field
of the ActivityPayload. I hope there is an alternative way of solving the
problem.

Bert-R:
> Hmm. I looked at the commit "Remove unnessecary database access" (https://github.com/yonadev/yona-server/pull/279/commits/3061ca3a0bc38ac9c807e6056da6aede28483438)
but I think the UserAnonymizedHolder is not an improvement to that, as the UserAnonymizedService already caches that entity.

That's a mistake. The entity isn't cached there but the DTO.

> So the real reason is that the activities pairing table is removed, so that the cached interval activities at the user anonymized have to be reloaded, isn't it? Did this not work properly before or is it affected by removing the relation between intervalactivities and activities? Maybe you should explain this to me by
phone :).

Saving the entities is now done through the cascading relationships. So to save an activity, you now call save the ``UserAnonymized`` entity to its repository. This means that you need to have that entity if you want to save it. However, if you fetch it before doing the analysis (which is what happened before the commit you're referring to), then you're going to the database always (to fetch the ``UserAnonymized`` entity), even though the outcome of the analysis might be that nothing needs to happen.

To prevent that, the ``UserAnonymized`` entity is only fetched when it's truly needed. And if it has been fetched, we save it to its repository.

We can have a call if you want, either this evening or Friday evening.

PR #280, opened 2016-12-13
YD-391 Store the spread cells as tinyblob

Bert-R:
Along with this, reduced the set of constructor parameters, preventing the null/empty values.


core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -169,6 +164,20 @@ public int getTotalActivityDurationMinutes()
 		return computeTotalActivityDurationMinutes();
 	}
 
+	private List<Integer> spreadBytesAsIntegerList()
+	{
+		if (spread == null)
```
aristotelos: This should never be true. The spread should be computed in this case. Actually, it should always have the length of 96.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -55,8 +55,8 @@ public static IntervalActivityRepository getIntervalActivityRepository()
 	 */
 	private LocalDate startDate;
 
-	@ElementCollection
-	private List<Integer> spread;
+	@Column(length = SPREAD_COUNT)
```
aristotelos: Currently, `spread` is never actually set because the aggregates are never computed (there is no batch script yet).

core/src/main/java/nu/yona/server/goals/entities/TimeZoneGoal.java
```diff
@@ -28,8 +30,8 @@
 	@ElementCollection
 	private List<String> zones;
 
-	@ElementCollection
-	private List<Integer> spreadCells;
+	@Column(length = 24 * 14) // 15 minute intervals
```
aristotelos: Shouldn't it be 24 * 4?

core/src/main/java/nu/yona/server/goals/entities/TimeZoneGoal.java
```diff
@@ -106,6 +113,16 @@ public int computeTotalMinutesBeyondGoal(DayActivity dayActivity)
 		return Math.min(dayActivity.getTotalActivityDurationMinutes(), sumOfSpreadOutsideGoal);
 	}
 
+	private static byte[] integerListToByteArray(List<Integer> spreadCells)
```
aristotelos: You might consider to move this method to a `SpreadUtil` class.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -169,6 +164,20 @@ public int getTotalActivityDurationMinutes()
 		return computeTotalActivityDurationMinutes();
 	}
 
+	private List<Integer> spreadBytesAsIntegerList()
```
aristotelos: You might consider to move this to a `SpreadUtil` class.

core/src/main/java/nu/yona/server/goals/entities/TimeZoneGoal.java
Bert-R: It's only used here. so there is currently no reason to centralize it.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
Bert-R: It's only used here. so there is currently no reason to centralize it.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
Bert-R: OK. Replaced this if-statement with ``assert spread.length == SPREAD_COUNT;``.

core/src/main/java/nu/yona/server/goals/entities/TimeZoneGoal.java
Bert-R: Indeed.

aristotelos:
OK you can merge it.

PR #281, opened 2016-12-14
YD-392 Spread cells of TimeZoneGoal now stored as CLOB

Bert-R:
Added a test to verify that max length is handled properly.


core/src/main/java/nu/yona/server/goals/entities/TimeZoneGoal.java
```diff
@@ -25,8 +26,8 @@
 @Entity
 public class TimeZoneGoal extends Goal
 {
-	@ElementCollection
-	private List<String> zones;
+	@Column(length = 24 * 4 * 6)
```
aristotelos: Why this length? I would expect that there are 24*4 quarters of an hour in a day, so that you can indeed create 24*4 non-overlapping zone specifications. However, every zone specification is 11 characters "hh:mm-hh:mm".

Bert-R:
Corrected the length and added a comment.

PR #282, opened 2016-12-14
YD-389 CamelCaseACRONYM as CamelCaseAcronym

Bert-R:
Along with this resolved a naming issue where we used activityId where we meant intervalActivityId.


Bert-R:
Question: shall we rename xxxDTO into xxxDTO? That would be in alignment with our current convention.

I might do the same rename on the Groovy files and push that to this PR.

Bert-R:
Updated the Groovy files too.

aristotelos:
Looks great. The "DTO" suffix can become "Dto" as discussed. After that (and conflict resolving) I will merge it.

PR #283, opened 2016-12-17
YD-393 Added support for long URLs

Bert-R:
Along with this, added version info, so we can enhance the crypto mechanism over time.


core/src/main/java/nu/yona/server/crypto/CryptoSession.java
```diff
@@ -136,36 +129,34 @@ public static CryptoSession getCurrent()
...
 
-	private SecretKey getSecretKey(char[] password)
+	private static SecretKey getSecretKey(String password)
 	{
 		try
 		{
-			SecretKeyFactory factory = SecretKeyFactory.getInstance("PBKDF2WithHmacSHA256");
-			KeySpec spec = new PBEKeySpec(password, SALT, 65536, 128);
+			SecretKeyFactory factory = SecretKeyFactory.getInstance(SECRET_KEY_ALGORITHM);
+			KeySpec spec = new PBEKeySpec(password.toCharArray(), SALT, 65536, SECRET_KEY_LENGTH_BITS);
```
aristotelos: Why do you use so many (65536) iterations? Some recommendations on the number are [here](http://security.stackexchange.com/questions/3959/recommended-of-iterations-when-using-pkbdf2-sha256), but I think we should make this less because it is used very often and the password has a high entropy. Would a low number of iterations (e.g. 100) not be enough?

core/src/main/java/nu/yona/server/crypto/CryptoSession.java
```diff
@@ -30,21 +28,16 @@
...
-	}
-
-	public interface VoidPredicate
-	{
-		boolean test();
-	}
+	private static final String SECRET_KEY_ALGORITHM = "PBKDF2WithHmacSHA256";
+	private static final int SECRET_KEY_LENGTH_BITS = 128;
+	/**
+	 * As we are using secure random passwords, a fixed salt suffices.
```
aristotelos: You could as a hacker try a lot of random passwords with this fixed salt appended. The chance that you hit a match is low. But it is higher than trying to find a combination of user ID and password. So I think it is acceptible but decreases security a little.

core/src/main/java/nu/yona/server/crypto/CryptoSession.java
```diff
@@ -30,21 +28,16 @@
...
+	 */
+	private static final byte[] SALT = "0123456789012345".getBytes();
```
aristotelos: Maybe good to give this a more speaking name `SECRET_KEY_SALT`.

appservice/src/intTest/groovy/nu/yona/server/BasicBuddyTest.groovy
```diff
@@ -382,6 +380,46 @@ class BasicBuddyTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(bob)
 	}
 
+	def 'Goal conflict of Richard with an 2k long URL is reported to Richard and Bob'()
```
Bert-R: It'll fail, probably with a vague error message. However, such extremely long URLs are very unusual. [According to Stack Overflow](http://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers), the defacto limit is 2000 characters. Because of that, I didn't want to invest time in being nice in such cases.

core/src/main/java/nu/yona/server/crypto/CryptoSession.java
Bert-R: The value of the salt is that passwords automatically become complex and to ensure that two identical passwords to encrypt the same data deliver different results. Given that we use long random passwords, the salts do not add value. If we would use varying salts, we would need to store the salt along with the data, thus bloating it (even more).

core/src/main/java/nu/yona/server/crypto/CryptoSession.java
Bert-R: It's a private static inside ``CryptoSession``, so what else can it be? Salt is the common crypto term.

core/src/main/java/nu/yona/server/crypto/CryptoSession.java
Bert-R: This I'm not sure of. [Wikipedia](https://en.wikipedia.org/wiki/Key_derivation_function) talks about  "often tens or hundreds of thousands". Given our random passwords, it might be sufficient to go down. However, given that the performance tests till now indicate that we're fast enough, I'd rather not go down (this number is there since the first version).

I still intend to a security expert to review our solution. We have one in our network that already offered to look into it.

appservice/src/intTest/groovy/nu/yona/server/BasicBuddyTest.groovy
aristotelos: The post gives an advice to stay under 2000 characters because of search engine preferences and old IE versions. However, all modern browsers seem to support 8192 or more characters. So I would like that it is truncated.

core/src/main/java/nu/yona/server/crypto/CryptoSession.java
aristotelos: But it is not used for encryption, it is used for generating a secret key out of a password using a hashing method, isn't it?

core/src/main/java/nu/yona/server/crypto/CryptoSession.java
aristotelos: Usually password hashing is only used at login or changing user password. We use it at every `CryptoSession` which is almost all app requests. So that is why I think it could be a costlier situation for us than for others. But still, it may take only milliseconds, I have not measured it yet :).

appservice/src/intTest/groovy/nu/yona/server/BasicBuddyTest.groovy
Bert-R: OK, will do.

core/src/main/java/nu/yona/server/crypto/CryptoSession.java
Bert-R: Indeed, salts are used for to generate the secret key, but that's what salts are for.

core/src/main/java/nu/yona/server/crypto/CryptoSession.java
Bert-R: Fair point. I'm already working on an optimization, for the scenario where we're using the key just once.

Bert-R: See also [my question on StackOverflow](http://stackoverflow.com/questions/41306171/are-iterations-and-a-salt-necessary-when-using-random-password-with-pbekeyspec).

Bert-R:
Pushed an update. The test execution is a lot faster now.
With my first build (using 65536 iterations) it took 10:34. Then I split the iteration counts: 65536 for the user key and 1000 for the per message key. That brought it down to 8:15. Then reduced to what's in the current PR: 1000/100. This brought it down to 3:47.

analysisservice/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -79,6 +80,10 @@ public void analyze(UUID userAnonymizedId, AppActivityDto appActivities)
 	@Transactional
 	public void analyze(UUID userAnonymizedId, NetworkActivityDto networkActivity)
 	{
+		if (networkActivity.getUrl().length() > MAX_SUPPORTED_URL_LENGTH)
```
aristotelos: Why not truncate it? Wouldn't that be better for analysis?

Bert-R:
With this length, I'd rather think someone is trying to fool us.

Bert-R:
There is one more step I want to take: have the server create the password when the user is created, so we can ensure it's done the right way.

PR #284, opened 2016-12-23
YD-384 Use autoincrement IDs for (Interval)Activity

Bert-R:


Bert-R:
I'll use the same branch to use autoincrement IDs for messages too.

core/src/main/java/nu/yona/server/analysis/entities/WeekActivity.java
```diff
@@ -84,11 +82,13 @@ public void addDayActivity(DayActivity dayActivity)
 			throw new IllegalStateException("Week is already full (" + dayActivities.size() + " days present)");
 		}
 
+		dayActivity.setWeekActivity(this);
 		dayActivities.add(dayActivity);
 	}
 
 	public void removeAllDayActivities()
 	{
+		dayActivities.forEach(da -> da.setWeekActivity(null));
```
aristotelos: I guess this has to do with the orphan removal. Why do you not just call delete on the `DayActivityRepository`?

Bert-R: Indeed, the deletion is taken care of through orphan removal. I believe I had an issue calling delete directly, but I'm not entirely sure. Do you believe it's worthwhile retesting that? I expect it to be identical in performance.

aristotelos: Well, the delete is implicit here, so you can leave it with this but then you should maybe add a comment that this works because of orphan removal. It doesn't matter much.

PR #285, opened 2017-01-02
YD-384 use autoincrement IDs for messages

Bert-R:


core/src/main/java/nu/yona/server/messaging/entities/MessageDestination.java
```diff
@@ -107,10 +114,11 @@ private PublicKey loadPublicKey()
 		return publicKey;
 	}
 
-	public void removeMessagesFromUser(UUID sentByUserAnonymizedId)
+	public List<Message> removeMessagesFromUser(UUID sentByUserAnonymizedId)
```
aristotelos: This does not execute a remove any more. Please rename the method.

core/src/main/java/nu/yona/server/messaging/service/BuddyMessageDto.java
```diff
@@ -41,6 +43,17 @@ public String getMessage()
 	public static abstract class Manager extends MessageDto.Manager
 	{
 		@Override
+		protected Optional<UUID> getSenderUserAnonymizedId(UserDto actingUser, Message messageEntity)
+		{
+			BuddyMessage buddyMessageEntity = (BuddyMessage) messageEntity;
+			if (actingUser.getId().equals(buddyMessageEntity.getSenderUserId()))
```
aristotelos: Why has this changed?

core/src/main/java/nu/yona/server/messaging/service/MessageService.java
```diff
@@ -238,7 +238,13 @@ public void sendMessage(Message message, MessageDestinationDto destination)
 	{
 		MessageDestination destinationEntity = MessageDestination.getRepository().findOne(destination.getId());
 		destinationEntity.send(message);
-		MessageDestination.getRepository().save(destinationEntity);
+		MessageDestination.getRepository().saveAndFlush(destinationEntity);
```
aristotelos: Please add a comment why the flush is necessary.

core/src/main/java/nu/yona/server/messaging/service/MessageService.java
```diff
@@ -238,7 +238,13 @@ public void sendMessage(Message message, MessageDestinationDto destination)
...
+	}
+
+	@Transactional
+	public void sendMessage(Message message, MessageDestination destinationEntity)
```
aristotelos: It is a bit confusing that this method does not call a flush on the `MessageDestination` repostitory as the other `sendMessage` method above. Maybe we should rename one of both?

core/src/main/java/nu/yona/server/messaging/service/MessageService.java
```diff
@@ -250,8 +256,11 @@ public void removeMessagesFromUser(MessageDestinationDto destination, UUID sentB
 		}
 
 		MessageDestination destinationEntity = MessageDestination.getRepository().findOne(destination.getId());
-		destinationEntity.removeMessagesFromUser(sentByUserAnonymizedId);
-		MessageDestination.getRepository().save(destinationEntity);
+		List<Message> messages = destinationEntity.removeMessagesFromUser(sentByUserAnonymizedId);
+		messages.stream().forEach(m -> m.clearThreadHeadSelfReference());
```
aristotelos: Please add a comment to explain why this is necessary

appservice/src/main/java/nu/yona/server/messaging/rest/MessageController.java
```diff
@@ -250,7 +250,7 @@ public MessageDto toResource(MessageDto message)
 
 		private void addRelatedMessageLink(MessageDto message, MessageDto messageResource)
 		{
-			if (message.getRelatedMessageId() != null)
+			if (message.getRelatedMessageId() != 0)
```
aristotelos: Shouldn't we use `Optional<Long>` here instead of `long`, as with `repliedMessageId` below?

aristotelos:
Overall a lot of great work has been done to fix the relations!

core/src/main/java/nu/yona/server/messaging/entities/BuddyMessage.java
```diff
@@ -85,14 +86,16 @@ public void encrypt(Encryptor encryptor)
 	{
 		senderUserIdCiphertext = encryptor.encrypt(senderUserId);
 		messageCiphertext = encryptor.encrypt(message);
-		senderNicknameCiphertext = encryptor.encrypt(senderNickname);
+		// senderNicknameCiphertext = encryptor.encrypt(senderNickname);
```
Bert-R: Certainly!

core/src/main/java/nu/yona/server/messaging/entities/MessageDestination.java
Bert-R: Done.

core/src/main/java/nu/yona/server/messaging/service/BuddyMessageDto.java
Bert-R: Because of a change in the use of {{relatedUserAnonymizedId}} in {{Message}}. It now also has this purpose: "If this message is a sent item, it is the target user of buddy copy of this message". The "related" aspect used to be interpreted as "sender", but that's not necessarily the case; it can also be the receiver.

core/src/main/java/nu/yona/server/messaging/service/MessageService.java
Bert-R: Done.

core/src/main/java/nu/yona/server/messaging/service/MessageService.java
Bert-R: Made both methods do a flush.

core/src/main/java/nu/yona/server/messaging/service/MessageService.java
Bert-R: Done

appservice/src/main/java/nu/yona/server/messaging/rest/MessageController.java
Bert-R: Yes, that's better.

PR #287, opened 2017-01-03
YD-397 Moved password generation to the server

Bert-R:
The app can still provide the password upon user sign-up, but this is only temporary. Once both app implementations are updated, this will be disabled.

The Yona password is now returned as part of the private user data. This happens when signing up, but also in other scenarios where the private user data is returned (e.g. GET on user). The app should take the password as returned in the user data and use that in subsequent requests.

Changes:
* Added yonaPassword to the Swagger spec
* Included it in the UserPrivateDto
* Rearranged some crypto-related code
* Made UserController.addUser generate the key if the Yona-Password header was not provided
* Updated the Groovy tests. Most tests now use the server-generated password. Some tests still generated it in the client, which is good for backward compatibility testing.

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
@@ -122,6 +122,7 @@ class CreateUserOnBuddyRequestTest extends AbstractAppServiceIntegrationTest
 		def newNickname = "Bobby"
 		def newPassword = "B o b"
 		def updatedBobJson = bob.convertToJson()
+		updatedBobJson.yonaPassword = newPassword
 		updatedBobJson.nickname = newNickname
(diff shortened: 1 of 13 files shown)


appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -122,6 +122,7 @@ class CreateUserOnBuddyRequestTest extends AbstractAppServiceIntegrationTest
 		def newNickname = "Bobby"
 		def newPassword = "B o b"
 		def updatedBobJson = bob.convertToJson()
+		updatedBobJson.yonaPassword = newPassword
```
aristotelos: With update user on buddy request, previously the Yona-Password header was set to the new password and the overwrite user confirmation code was sent. Now both the JSON yonaPassword property and the Yona-Password header are sent. However, the first is ignored, isn't it? It might be good to document that or do a validation that both are the same?

PR #288, opened 2017-01-06
YD-398 Refactor cryptosession consumers and split crypto package

Bert-R:
* Refactored ``CryptoSession`` consumers for readability. Along with that, consistently named the user ID URL parameter as ``userId``
* Split the ``crypto`` package in two:
  * ``seckey`` with secret key encryption
  * ``pubkey`` with public key encryption


appservice/src/main/java/nu/yona/server/analysis/rest/BuddyActivityController.java
```diff
@@ -1,9 +1,6 @@
 /*******************************************************************************
- * Copyright (c) 2016 Stichting Yona Foundation
- *
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ * Copyright (c) 2016 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License, v.
```
aristotelos: I would expect 2016-2017 in the copyright... We can solve that in another PR.

core/src/main/java/nu/yona/server/crypto/pubkey/DecryptionInfo.java
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	public byte[] getInitializationVector()
+	{
+		return initializationVector;
+	}
+
+	public byte[] convertToByteArray()
+	{
+		byte[] secreteKeyBytes = secretKey.getEncoded();
```
aristotelos: `secretKeyBytes`

core/src/main/java/nu/yona/server/crypto/seckey/SecretKeyUtil.java
```diff
@@ -0,0 +1,196 @@
...
+		return CryptoSession.getCurrent().decrypt(ciphertext);
+	}
+
+	/**
+	 * Encrypts the given plaintext bytes.
+	 * 
+	 * @param plaintext the bytes to be encrypted
+	 * @return the encrypted bytes, with the crypto variant number prepended to it.
+	 */
+	public static byte[] encrypt(byte cryptoVariantNumber, Cipher cipher, byte[] plaintext)
```
aristotelos: This method should certainly be in the shared `CryptoUtil` class, because it is used both by public key and secret key encryption. The cipher is passed.

core/src/main/java/nu/yona/server/crypto/seckey/SecretKeyUtil.java
```diff
@@ -0,0 +1,196 @@
...
+		return new SecretKeySpec(secretKeyBytes, "AES");
+	}
+
+	/**
+	 * Decrypts the given ciphertext bytes.
+	 * 
+	 * @param ciphertext the bytes to be decrypted, with a leading crypto variant number
+	 * @return the decrypted bytes
+	 */
+	public static byte[] decrypt(byte cryptoVariantNumber, Cipher cipher, byte[] ciphertext)
```
aristotelos: This method should be in the `CryptoUtil` class like the `encrypt`.

core/src/main/java/nu/yona/server/crypto/pubkey/Encryptor.java
```diff
@@ -2,7 +2,7 @@
  * Copyright (c) 2015, 2016 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License,
  * v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
  *******************************************************************************/
-package nu.yona.server.crypto;
+package nu.yona.server.crypto.pubkey;
 
 public interface Encryptor
```
aristotelos: We should maybe use 'encrypter'/'decrypter' instead of 'encryptor'/'decryptor'. The -or ending is less common, see https://en.oxforddictionaries.com/spelling/nouns-ending-in-er-or-and-ar. Currently we are using both by the way (e.g. `StringFieldEncrypter`.

appservice/src/main/java/nu/yona/server/analysis/rest/BuddyActivityController.java
Bert-R: Indeed

core/src/main/java/nu/yona/server/crypto/pubkey/DecryptionInfo.java
Bert-R: Done

core/src/main/java/nu/yona/server/crypto/seckey/SecretKeyUtil.java
Bert-R: Indeed. Done

core/src/main/java/nu/yona/server/crypto/seckey/SecretKeyUtil.java
Bert-R: Indeed. Done.

core/src/main/java/nu/yona/server/crypto/pubkey/Encryptor.java
Bert-R: Two variants is not good. Encrypt and Decrypt have Latin origins, so I prefer the Lattin "or" suffix. That's also pretty common, see [this link](https://books.google.com/ngrams/graph?content=encryptor%2C+encrypter&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cencryptor%3B%2Cc0%3B.t1%3B%2Cencrypter%3B%2Cc0).

Bert-R:
Pushed corrections.

PR #289, opened 2017-01-11
YD-399 Added Spring Boot Actuator

Bert-R:
To fetch management information from the running system.

See [Spring Boot Actuator](http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready).


Bert-R:
I've tested this extensively in through the regular {{servers_start}} and in Docker containers and I don't see the vague class loading issues anymore that I had yesterday night.

aristotelos:
I also got the vague class loading issue on master, so it is not due to this functionality. I restarted and everything worked fine.

core/src/main/resources/application.properties
```diff
@@ -33,6 +33,10 @@ spring.jackson.deserialization.fail-on-reading-dup-tree-key=true
 spring.jackson.deserialization.fail-on-ignored-properties=true
 spring.jackson.deserialization.accept-float-as-int=false
 spring.jackson.serialization.write-dates-as-timestamps=false
+spring.jackson.serialization.indent_output=true
+
+# Management (Actuator)
+management.security.enabled=false
```
aristotelos: Maybe add a comment this can be done because the ports are not accessible from outside.

core/src/main/java/nu/yona/server/CoreConfiguration.java
```diff
@@ -121,6 +122,7 @@ public JavaMailSender javaMailSender()
 	private BeanFactory beanFactory;
 
 	@Bean
+	@Primary
```
Bert-R: I didn't fix it. I don't think it matters much, as all end points are anyway documented. I expect we'll directly use an endpoint.

core/src/main/resources/application.properties
Bert-R: Makes sense. Done.

PR #290, opened 2017-01-11
YD-401 Limit the number of users

Bert-R:
It's done through a configuration property. This requires the service to be restarted when the maximum is updated. Given that this is a temporary feature and restarting is fast, this shouldn't be a problem.


Bert-R:
Tested manually.

aristotelos:
Looks great. I can reuse the `verifyUserIsAllowed` method.

PR #294, opened 2017-01-19
YD-407 Fixed the issue by handling the transaction inside the lock

Bert-R:


Bert-R:
I don't know how we could do that. It would be a Spring integration test that would run a bunch of threads to see it can force the issue. That would require mocking a bunch of repositories. It would be no means be trivial.

I've thought of writing a test from Groovy, but I doubt the usefulness of it. If it fails, you can prove it's wrong, but you can't prove it's right. There might still be that small interval that is missed in the test but used in reality.

PR #296, opened 2017-02-03
YD-409 Improved batch job triggering

Bert-R:
Instead of polling for empty pin reset requests, the batch job now gets triggered by the app service to generate one at the specified time.


aristotelos:
Maybe good that we discuss this next week?

Bert-R:
Yes, that's fine.

PR #299, opened 2017-02-07
YD-411 Fixed week already full error

Bert-R:
The cause was that upon a goal change, we create a copy of the goal, but both copies had the same start time. The history item should carry the original creation time, but the "actual" goal should have its creation time updated to the current time. Once the data is corrected, the "week already full" issue will disappear. The fix of this change set will prevent from creating the wrong data again.


appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
```diff
@@ -15,6 +15,7 @@ import nu.yona.server.test.BudgetGoal
 import nu.yona.server.test.Goal
 import nu.yona.server.test.TimeZoneGoal
 import nu.yona.server.test.User
+import spock.lang.IgnoreRest
```
aristotelos: This could be removed

PR #300, opened 2017-02-08
YD-413 Improved start script

Bert-R:
Improved the test run script in the following ways:
* Now doesn't run Gradle in parallel anymore. This addresses the random class loading error and makes the service start up faster
* Now automatically starts the tests when all services are up and running (no need to hit enter anymore)
* Gives a more clear error message when something fails

Note that this requires curl 7.51.1 or later, as we use curl/curl#1064

### DIFF
--- a/scripts/servers_start.cmd
+++ b/scripts/servers_start.cmd
@@ -41,7 +41,8 @@ if ERRORLEVEL 1 goto error
 echo.
 echo Load the activity categories
 echo.
-curl -X PUT --header "Content-Type: application/json" -d @dbinit/data/activityCategories.json http://localhost:8080/activityCategories/
+curl -f -X PUT --header "Content-Type: application/json" -d @dbinit/data/activityCategories.json http://localhost:8080/activityCategories/
+if ERRORLEVEL 1 goto error
 
 echo.
 echo.
--- a/scripts/servers_start.cmd
+++ b/scripts/servers_start.cmd
@@ -2,31 +2,57 @@
  
 call servers_stop.cmd
 
-call gradlew %1 build
-if ERRORLEVEL 1 goto end
+call gradlew build
+if ERRORLEVEL 1 goto error
 
+if "%1"=="-keepDB" goto updateDB
 echo.
 echo Recreating Yona database
 echo.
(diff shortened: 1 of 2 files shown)


scripts/servers_start.cmd
```diff
@@ -2,31 +2,58 @@
...
+echo.
 call gradlew :dbinit:bootRun
-if ERRORLEVEL 1 goto end
+if ERRORLEVEL 1 goto error
 
-start "Admin service" cmd /c gradlew %1 :adminservice:bootRun
-start "Analysis service" cmd /c gradlew %1 :analysisservice:bootRun
-start "App service" cmd /c gradlew %1 :appservice:bootRun
-start "Batch service" cmd /c gradlew %1 :batchservice:bootRun
+call gradlew adminservice:build && start "Admin service" java -Xdebug -Xrunjdwp:transport=dt_socket,address=8840,server=y,suspend=n -jar adminservice\build\libs\adminservice-0.0.8-SNAPSHOT-full.jar --server.port=8080 --management.port=9080 --spring.jpa.hibernate.ddl-auto=none
```
aristotelos: This will enable debug by default?

scripts/servers_start.cmd
```diff
@@ -2,31 +2,58 @@
...
+
+echo.
+echo Load the activity categories
+echo.
+curl -f -X PUT --header "Content-Type: application/json" -d @dbinit/data/activityCategories.json http://localhost:8080/activityCategories/
+if ERRORLEVEL 1 goto error
+
+echo.
+echo.
+echo Start the integration tests
```
aristotelos: I do sometimes start the services without running the integration tests, to do manual testing, so I would prefer a `pause so I can escape the integration tests there.

scripts/servers_start.cmd
Bert-R: Indeed. At OpenText, all our development machines run with debug enabled by default, with no negative side effects.

scripts/servers_start.cmd
Bert-R: Can't you simply hit ctrl-c when the tests run?
The paus option means that you've to look after it and come back in the middle. In your case, I expect you will want to load the activity categories, so you can't break it at "pause", but need to make it do the next step, right?

aristotelos: Indeed, I locally added an extra `pause`. Ctrl+C is fine too.

PR #303, opened 2017-02-10
YD-410 Added white list properties

Bert-R:
There are now two properties, one whether invited users are checked against the white list and one for free sign up (users that are not invited).

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
@@ -60,6 +60,11 @@ public class UserService
 	/** Holds the regex to validate a valid phone number. Start with a '+' sign followed by only numbers */
 	private static Pattern REGEX_PHONE = Pattern.compile("^\\+[1-9][0-9]+$");
 
+	private enum UserSignUp
+	{
+		INVITED, FREE
+	}
+
 	private static final Logger logger = LoggerFactory.getLogger(UserService.class);
 
 	@Autowired(required = false)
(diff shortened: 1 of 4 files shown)


core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -247,27 +252,33 @@ private void verifyUserDoesNotExist(String mobileNumber)
...
+	private void verifyUserIsAllowed(String mobileNumber, UserSignUp origin)
+	{
+		verifyBelowMaxUsers();
+		verifyWhiteList(mobileNumber, origin);
+	}
+
+	private void verifyBelowMaxUsers()
 	{
 		int maxUsers = yonaProperties.getMaxUsers();
-		if (maxUsers <= 0)
```
aristotelos: This line was removed, which means `maxUsers` is always required.

core/src/main/resources/application.properties
```diff
@@ -46,6 +46,8 @@ yona.supportedLocales=en-US,nl-NL
 yona.appleAppId = SFWNSYN8YK.com.xaton.yona
 # The maximum number of users accounts allowed. If set to negative or zero, it will fall back to white list mobile number verification.
```
aristotelos: This comment is no longer correct.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
Bert-R: Indeed. There is no value in allowing negative values. If you don't like to use this property, just set it to a very large value (as we do today).

core/src/main/resources/application.properties
Bert-R: Removed it.

PR #304, opened 2017-02-11
YD-406 corrected locale handling for Thymeleaf

Bert-R:


aristotelos:
Have you found on the Internet that this is a known issue?

Bert-R:
No, but your test proved it didn't work like that.

PR #308, opened 2017-02-20
YD-418 Fixed "job instance already exists"

Bert-R:
Added the current date/time as a uniqueInstanceId job parameter. That way, it can be run repeatedly.

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/PinResetRequestTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/PinResetRequestTest.groovy
@@ -58,6 +58,42 @@ class PinResetRequestTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(richard)
 	}
 
+	def 'Request pin reset twice'()
+	{
+		given:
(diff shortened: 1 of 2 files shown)


appservice/src/intTest/groovy/nu/yona/server/PinResetRequestTest.groovy
```diff
@@ -58,6 +58,42 @@ class PinResetRequestTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(richard)
 	}
 
+	def 'Request pin reset twice'()
+	{
+		given:
+		User richard = addRichard()
+		def firstResponse = appService.yonaServer.postJson(richard.pinResetRequestUrl, [:], ["Yona-Password" : richard.password, "Accept-Language" : "nl-NL"])
+		assert firstResponse.status == 200
+		sleepTillPinResetCodeIsGenerated(richard, firstResponse.responseData.delay)
```
aristotelos: Maybe we could add a test with the situation that the PIN reset is requested before the delay interval...

PR #313, opened 2017-03-06
YD-286 YD-299 Validate buddy data

Bert-R:
* Verify first and last name is given, as well as a valid mobile number
* Verify that mobile number is not of the inviting user, not an already existing buddy

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
@@ -124,8 +124,7 @@ public class BuddyService
 	{
 		UserDto requestingUser = userService.getPrivateUser(idOfRequestingUser);
 		assertMobileNumberOfRequestingUserConfirmed(idOfRequestingUser);
-		validateBuddyFields(buddy);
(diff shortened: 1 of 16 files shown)


core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -165,6 +166,24 @@ private void assertMobileNumberOfRequestingUserConfirmed(UUID idOfRequestingUser
 		requestingUser.assertMobileNumberConfirmed();
 	}
 
+	private void assertValidBuddy(UserDto requestingUser, BuddyDto buddy)
+	{
+		userService.assertValidUserFields(buddy.getUser(), true);
```
aristotelos: An integration test for violating this validation could be created too.

Bert-R: OK, will do.

Bert-R: Done, see ``BuddyValidationTest.groovy``

PR #314, opened 2017-03-06
YD-162 Daily aggregation batch job part 1

aristotelos:
- Pruning of activities not implemented yet, because the first part is already quite a lot of work

### DIFF
--- a/analysisservice/src/test/java/nu/yona/server/analysis/service/IntervalActivityTestsBase.java
+++ b/analysisservice/src/test/java/nu/yona/server/analysis/service/IntervalActivityTestsBase.java
@@ -1,0 +1,112 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License, v.
+ * 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *******************************************************************************/
+package nu.yona.server.analysis.service;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.junit.Assert.assertThat;
+
+import java.text.MessageFormat;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+import java.util.UUID;
+
+import org.junit.Before;
+
+import nu.yona.server.analysis.entities.Activity;
+import nu.yona.server.analysis.entities.DayActivity;
+import nu.yona.server.analysis.entities.IntervalActivity;
+import nu.yona.server.crypto.pubkey.PublicKeyUtil;
+import nu.yona.server.goals.entities.ActivityCategory;
+import nu.yona.server.goals.entities.BudgetGoal;
+import nu.yona.server.goals.entities.Goal;
+import nu.yona.server.goals.entities.TimeZoneGoal;
+import nu.yona.server.messaging.entities.MessageDestination;
+import nu.yona.server.subscriptions.entities.UserAnonymized;
+import nu.yona.server.util.TimeUtil;
+
+public abstract class IntervalActivityTestsBase
+{
+	protected ZoneId testZone;
+
+	protected BudgetGoal budgetGoal;
+	protected TimeZoneGoal timeZoneGoal;
+
+	protected UserAnonymized userAnonEntity;
+
+	@Before
+	public void setUp()
+	{
+		testZone = ZoneId.of("Europe/Amsterdam");
+		ActivityCategory activityCategory = ActivityCategory.createInstance(UUID.randomUUID(),
+				Collections.singletonMap(Locale.US, "being bored"), false, Collections.emptySet(), Collections.emptySet(),
+				Collections.singletonMap(Locale.US, "Descr"));
+		budgetGoal = BudgetGoal.createInstance(TimeUtil.utcNow(), activityCategory, 60);
+		timeZoneGoal = TimeZoneGoal.createInstance(TimeUtil.utcNow(), activityCategory,
+				Arrays.asList("05:00-12:30", "13:00-24:00"));
+		// Set up UserAnonymized instance.
+		MessageDestination anonMessageDestinationEntity = MessageDestination
+				.createInstance(PublicKeyUtil.generateKeyPair().getPublic());
+		Set<Goal> goals = new HashSet<Goal>(Arrays.asList(budgetGoal));
+		userAnonEntity = UserAnonymized.createInstance(anonMessageDestinationEntity, goals);
+	}
+
+	protected ZonedDateTime getDate(DayActivity d, String timeString)
+	{
+		int hourIndex = timeString.indexOf(':');
+		int hour = Integer.parseInt(timeString.substring(0, hourIndex));
+		int minuteIndex = timeString.indexOf(':', hourIndex + 1);
+		if (minuteIndex == -1)
+		{
+			int minute = Integer.parseInt(timeString.substring(hourIndex + 1));
+			return d.getStartTime().withHour(hour).withMinute(minute);
+		}
+		else
+		{
+			int minute = Integer.parseInt(timeString.substring(hourIndex + 1, minuteIndex));
+			int second = Integer.parseInt(timeString.substring(minuteIndex + 1));
+			return d.getStartTime().withHour(hour).withMinute(minute).withSecond(second);
+		}
+	}
+
+	protected void addActivity(DayActivity d, String startTimeString, String endTimeString)
+	{
+		d.addActivity(Activity.createInstance(testZone, getDate(d, startTimeString).toLocalDateTime(),
+				getDate(d, endTimeString).toLocalDateTime()));
+	}
+
+	protected void assertSpreadItemsAndTotal(IntervalActivity d, String expectedSpreadItems,
+			int expectedTotalActivityDurationMinutes)
+	{
+		assertThat(d.areAggregatesComputed(), equalTo(false));
+		assertSpreadItems(d.getSpread(), expectedSpreadItems);
+		assertThat(d.getTotalActivityDurationMinutes(), equalTo(expectedTotalActivityDurationMinutes));
+
+		d.computeAggregates();
+		assertThat(d.areAggregatesComputed(), equalTo(true));
+		assertSpreadItems(d.getSpread(), expectedSpreadItems);
+		assertThat(d.getTotalActivityDurationMinutes(), equalTo(expectedTotalActivityDurationMinutes));
+	}
+
+	protected void assertSpreadItems(List<Integer> actualSpread, String expectedSpreadItems)
+	{
+		Arrays.stream(expectedSpreadItems.split(","))
+				.forEach(expectedSpreadItem -> assertSpreadItem(actualSpread, expectedSpreadItem));
+	}
+
+	protected void assertSpreadItem(List<Integer> actualSpread, String expectedSpreadItem)
+	{
+		int separatorIndex = expectedSpreadItem.indexOf('=');
+		int spreadIndex = Integer.parseInt(expectedSpreadItem.substring(0, separatorIndex));
+		int expectedValue = Integer.parseInt(expectedSpreadItem.substring(separatorIndex + 1));
+		assertThat(MessageFormat.format("expected value {0} at spread index {1}", expectedValue, spreadIndex),
+				actualSpread.get(spreadIndex), equalTo(expectedValue));
+	}
+}
--- a/analysisservice/src/test/java/nu/yona/server/analysis/service/IntervalActivityTestsBase.java
+++ b/analysisservice/src/test/java/nu/yona/server/analysis/service/IntervalActivityTestsBase.java
@@ -10,6 +10,9 @@ import static org.junit.Assert.assertThat;
 import java.text.MessageFormat;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoField;
+import java.time.temporal.TemporalAccessor;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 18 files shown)


Bert-R:
This is named "part 1". What is yet to come?

aristotelos:
Part 2, obviously :). It is the pruning of individual activities from days
and of older day activities from weeks. I have left it out because it may
take more time than expected (e.g. writing integration tests how goal
conflict messages related to an individual activity keep behaving, how url
disclosure request stops working or remains working, how messages related
to pruned data are removed).

Bert-R:
OK, that's fine. That can well be done separately.

aristotelos:
Maybe we should use `fetch=FetchType.LAZY` on the `DayActivity.activities` in the meantime, to get a performance improvement?

Bert-R:
Don't think that helps. If a DayActivity is fetched, it's for the report, and that needs the Activity instances.

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -156,8 +156,8 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		def expectedValuesRichardLastWeek = [
 			"Mon" : [[goal:budgetGoalNewsRichard, data: [goalAccomplished: false, minutesBeyondGoal: 20, spread: [13 : 15, 14 : 5]]]],
 			"Tue" : [[goal:budgetGoalNewsRichard, data: [goalAccomplished: false, minutesBeyondGoal: 25, spread: [35 : 15, 36 : 10]]]],
-			"Wed" : [[goal:budgetGoalNewsRichard, data: [goalAccomplished: true, minutesBeyondGoal: 0, spread: []]], [goal:timeZoneGoalSocialRichard, data: [goalAccomplished: false, minutesBeyondGoal: 1, spread: [68 : 1]]]],
-			"Thu" : [[goal:budgetGoalNewsRichard, data: [goalAccomplished: true, minutesBeyondGoal: 0, spread: []]], [goal:timeZoneGoalSocialRichard, data: [goalAccomplished: true, minutesBeyondGoal: 0, spread: [48 : 1]]]],
+			"Wed" : [[goal:budgetGoalNewsRichard, data: [goalAccomplished: true, minutesBeyondGoal: 0, spread: []]], [goal:timeZoneGoalSocialRichard, data: [goalAccomplished: false, minutesBeyondGoal: 1, spread: [60 : 1]]]],
```
Bert-R: How can this be wrong while the test was successful?

batchservice/src/main/java/nu/yona/server/batch/jobs/ActivityAggregationBatchJob.java
```diff
@@ -0,0 +1,176 @@
...
+			reader.setParameterValues(Collections.singletonMap("lastWeek",
+					TimeUtil.getStartOfWeek(ZonedDateTime.now(DEFAULT_TIME_ZONE).toLocalDate()).minusWeeks(1)));
+			reader.setEntityManagerFactory(entityManager.getEntityManagerFactory());
+			reader.setPageSize(WEEK_ACTIVITY_CHUNK_SIZE);
+			reader.afterPropertiesSet();
+			reader.setSaveState(true);
+
+			return reader;
+		}
+		catch (Exception e)
```
Bert-R: Please change this t a multicatch with list of exceptions being caught.

batchservice/src/main/java/nu/yona/server/batch/jobs/ActivityAggregationBatchJob.java
```diff
@@ -0,0 +1,176 @@
...
+			reader.setParameterValues(
+					Collections.singletonMap("yesterday", ZonedDateTime.now(DEFAULT_TIME_ZONE).minusDays(1).toLocalDate()));
+			reader.setEntityManagerFactory(entityManager.getEntityManagerFactory());
+			reader.setPageSize(DAY_ACTIVITY_CHUNK_SIZE);
+			reader.afterPropertiesSet();
+			reader.setSaveState(true);
+
+			return reader;
+		}
+		catch (Exception e)
```
Bert-R: Please change this t a multicatch with list of exceptions being caught.

batchservice/src/main/java/nu/yona/server/batch/jobs/ActivityAggregationBatchJob.java
```diff
@@ -0,0 +1,176 @@
...
+	public JpaPagingItemReader<WeekActivity> weekActivityReader()
+	{
+		try
+		{
+			String jpqlQuery = "SELECT w FROM WeekActivity w WHERE w.startDate <= :lastWeek AND w.aggregatesComputed = false";
+
+			JpaPagingItemReader<WeekActivity> reader = new JpaPagingItemReader<>();
+			reader.setQueryString(jpqlQuery);
+			reader.setParameterValues(Collections.singletonMap("lastWeek",
+					TimeUtil.getStartOfWeek(ZonedDateTime.now(DEFAULT_TIME_ZONE).toLocalDate()).minusWeeks(1)));
```
Bert-R: In ``dayActivityReader``, you're first subtracting a day and then converting it to ``LocalDate``. Here you're doing the conversion first. Please do it the same way in both places.

core/src/main/java/nu/yona/server/analysis/entities/Activity.java
```diff
@@ -88,6 +83,11 @@ public ZonedDateTime getEndTimeAsZonedDateTime()
 	public void setEndTime(LocalDateTime endTime)
 	{
 		this.endTime = endTime;
+
+		if (dayActivity != null)
```
Bert-R: When can this be null?

core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
```diff
@@ -86,6 +86,19 @@ public void addActivity(Activity activity)
...
+
+		this.resetAggregatesComputed();
+	}
+
+	@Override
+	protected void resetAggregatesComputed()
+	{
+		super.resetAggregatesComputed();
+
+		if (weekActivity != null)
```
Bert-R: When can this be null?

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -171,11 +193,21 @@ public int getTotalActivityDurationMinutes()
 		List<Integer> integers = new ArrayList<>(spread.length);
 		for (int i = 0; (i < spread.length); i++)
 		{
-			integers.set(i, (int) spread[i]);
+			integers.add((int) spread[i]);
 		}
 		return integers;
 	}
 
+	private static byte[] spreadIntegersAsByteList(List<Integer> spread)
```
Bert-R: Please rename into ``spreadIntegersAsByteArray``.

core/src/main/resources/application.properties
```diff
@@ -21,6 +21,9 @@ liquibase.enabled=false
 spring.batch.initializer.enabled=false
 spring.batch.job.enabled=false
 
+# Batch service
+yona.batchService.activityAggregationJobCronExpression=0 0 3 * * *
```
Bert-R: See above.

core/src/main/java/nu/yona/server/properties/BatchServiceProperties.java
```diff
@@ -8,6 +8,8 @@
 {
 	private String serviceUrl = "http://localhost:8083";
 
+	private String activityAggregationJobCronExpression = "0 0 3 * * *";
```
Bert-R: According to [Wikipedia](https://en.wikipedia.org/wiki/Cron) and [this site](http://www.corntab.com/?c=0_0_3_*_*_*), this job now runs at midnight, ever third of every month. That's not what you had in mind, I guess. What about using 0 3 * * * * (i.e. [every day at three in the morning](http://www.corntab.com/?c=0_3_*_*_*_*))?

core/src/testUtils/groovy/nu/yona/server/test/BatchService.groovy
```diff
@@ -0,0 +1,25 @@
...
+{
+	final ACTIVITY_AGGREGATION_TRIGGER_PATH = "/batch/aggregateActivities/"
+
+	BatchService ()
+	{
+		super("yona.batchservice.url", "http://localhost:8083")
+	}
+
+
+	def triggerAggregateActivities()
```
Bert-R: Please rename into ``triggerActivityAggregationBatchJob``

Bert-R: I ran the tests and got a couple of failures:
```
nu.yona.server.ActivityAggregationBatchJobTest > Days and weeks in the past are aggregated once FAILED
    org.spockframework.runtime.SpockComparisonFailure at ActivityAggregationBatchJobTest.groovy:63

nu.yona.server.ActivityAggregationBatchJobTest > Aggregation is reset and repeated when posting new app activity within combine interval FAILED
    org.spockframework.runtime.SpockComparisonFailure at ActivityAggregationBatchJobTest.groovy:117

nu.yona.server.ActivityAggregationBatchJobTest > Aggregation is reset and repeated when posting new app activity outside combine interval FAILED
    org.spockframework.runtime.SpockComparisonFailure at ActivityAggregationBatchJobTest.groovy:166

170 tests completed, 3 failed
:appservice:intTest FAILED
```

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
aristotelos: Because this test does not assert at the detail level, only at the aggregation level.

batchservice/src/main/java/nu/yona/server/batch/jobs/ActivityAggregationBatchJob.java
aristotelos: Actually, the `JpaPagingItemReader.afterPropertiesSet` has `throws Exception`, so it cannot be made more specific.

batchservice/src/main/java/nu/yona/server/batch/jobs/ActivityAggregationBatchJob.java
aristotelos: Actually, the `JpaPagingItemReader.afterPropertiesSet` has `throws Exception`, so it cannot be made more specific.

batchservice/src/main/java/nu/yona/server/batch/jobs/ActivityAggregationBatchJob.java
aristotelos: Thanks for pointing this out. I also forgot to use `TimeUtil.getStartOfDay` in the other case, so fixed that too.

core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
aristotelos: At no path that I know of, but I just tried to be careful for the future.

core/src/main/java/nu/yona/server/properties/BatchServiceProperties.java
aristotelos: According to [this](http://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/support/CronSequenceGenerator.html) and [this](https://docs.spring.io/spring/docs/current/spring-framework-reference/html/scheduling.html) Spring documentation, my expression causes the job to run every day at three in the morning. Standard CRON expressions do not work with Spring, I encountered that at the test...

core/src/main/resources/application.properties
aristotelos: See below

core/src/testUtils/groovy/nu/yona/server/test/BatchService.groovy
aristotelos: The tests indeed failed when not run in isolation. Fixed that now.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
aristotelos: Done

core/src/testUtils/groovy/nu/yona/server/test/BatchService.groovy
aristotelos: Done

aristotelos:
Processed review comments

batchservice/src/main/java/nu/yona/server/batch/jobs/ActivityAggregationBatchJob.java
Bert-R: OK, so be it.

batchservice/src/main/java/nu/yona/server/batch/jobs/ActivityAggregationBatchJob.java
Bert-R: OK, so be it.

core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
Bert-R: Yes. Scattering the code with null checks is a common pattern, but it reduces readability. Let's only put them in when they're really needed. The same comment also occurs in ``Activity``.

core/src/main/java/nu/yona/server/properties/BatchServiceProperties.java
Bert-R: I see. They have seconds as an extra field. It's weird that they're not more explicit about this.

appservice/src/intTest/groovy/nu/yona/server/ActivityAggregationBatchJobTest.groovy
```diff
@@ -0,0 +1,174 @@
...
+				[goal:timeZoneGoalSocialRichard, data: [goalAccomplished: true, minutesBeyondGoal: 0, spread: [ : ]]]
+			]]
+		assertActivityValues(richard, 1, expectedValuesRichardLastWeek, 2)
+
+		when:
+		def response = batchService.triggerActivityAggregationBatchJob()
+
+		then:
+		response.status == 200
+		response.responseData.writeCountPerStep?.aggregateWeekActivities >= 2 //may aggregate existing activities from other tests too, if not run in isolation
```
Bert-R: A better way to handle this is to do an initial aggregation right after the users are created. That way, the test can exactly assert what it expects, without resorting to "at least".

analysisservice/src/test/java/nu/yona/server/analysis/service/IntervalActivityTestsBase.java
```diff
@@ -0,0 +1,112 @@
...
+		MessageDestination anonMessageDestinationEntity = MessageDestination
+				.createInstance(PublicKeyUtil.generateKeyPair().getPublic());
+		Set<Goal> goals = new HashSet<Goal>(Arrays.asList(budgetGoal));
+		userAnonEntity = UserAnonymized.createInstance(anonMessageDestinationEntity, goals);
+	}
+
+	protected ZonedDateTime getDate(DayActivity d, String timeString)
+	{
+		int hourIndex = timeString.indexOf(':');
+		int hour = Integer.parseInt(timeString.substring(0, hourIndex));
```
Bert-R: Just use standard Java time parsing.

analysisservice/src/test/java/nu/yona/server/analysis/service/IntervalActivityTestsBase.java
```diff
@@ -0,0 +1,112 @@
...
+		timeZoneGoal = TimeZoneGoal.createInstance(TimeUtil.utcNow(), activityCategory,
+				Arrays.asList("05:00-12:30", "13:00-24:00"));
+		// Set up UserAnonymized instance.
+		MessageDestination anonMessageDestinationEntity = MessageDestination
+				.createInstance(PublicKeyUtil.generateKeyPair().getPublic());
+		Set<Goal> goals = new HashSet<Goal>(Arrays.asList(budgetGoal));
+		userAnonEntity = UserAnonymized.createInstance(anonMessageDestinationEntity, goals);
+	}
+
+	protected ZonedDateTime getDate(DayActivity d, String timeString)
```
Bert-R: These single-letter variable names don't say much.

PR #315, opened 2017-03-07
YD-286 YD-299 Validate email address

Bert-R:
For buddy requests


appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
```diff
@@ -95,6 +95,7 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
 	{
 		def richard = addRichard(false)
 		def bob = addBob(false)
+		bob.emailAddress = "bob@dunn.com"
```
aristotelos: Can't this be inlined in `addBob`? That removes a lot of duplicate code that complicates the tests.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -713,6 +717,15 @@ void assertValidUserFields(UserDto userResource, boolean nickNameIsOptional)
...
 
 		assertValidMobileNumber(userResource.getMobileNumber());
+
+		if (userPurpose == UserPurpose.BUDDY)
+		{
+			if (StringUtils.isBlank(userResource.getEmailAddress()))
+			{
+				throw InvalidDataException.blankEmailAddress();
+			}
+			assertValidEmailAddress(userResource.getEmailAddress());
```
aristotelos: Asserting that it is valid can also be done on `UserPurpose.USER` when the e-mail address is not blank.

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
Bert-R: It's attractive indeed, but cumbersome and error prone. In general, a user does not have an e-mail address, which means that we loose it upon lots of actions: reload, update, number confirmation, etc.

I tried it out, but that won't make things simpler.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
Bert-R: Will do.

Bert-R:
Added the extra check.

aristotelos:
Can be merged now according to me. Should I merge it?

Bert-R:
OK. Let's wait with merging it till the Git problem is resolved. Once that's done, I'll merge it.

PR #316, opened 2017-03-10
Declarative Pipelines' credentials support =)

abayer:
See https://jenkins.io/doc/book/pipeline/syntax/#environment for more info. =) You can also put all the `FOO = credentials('id')` into the `environment` at the top level and avoid the duplication if you prefer. Hope this is helpful!


yonabuild:

PR #321, opened 2017-03-17
YD-265 Prevent information leakage

aristotelos:


Bert-R:
See [my comment on the ticket](http://jira.yona.nu/browse/YD-265?focusedCommentId=12784&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-12784) for some feedback on the approach.

appservice/src/main/java/nu/yona/server/admin/rest/AdminController.java
```diff
@@ -4,29 +4,52 @@
...
+	@Autowired
+	private YonaProperties yonaProperties;
 
 	@Autowired
 	private UserService userService;
 
 	@RequestMapping(value = "/requestUserOverwrite/", method = RequestMethod.POST)
-	@ResponseStatus(HttpStatus.OK)
-	public void setOverwriteUserConfirmationCode(@RequestParam String mobileNumber)
+	public ResponseEntity<String> setOverwriteUserConfirmationCode(@RequestParam String mobileNumber, HttpServletRequest request)
```
Bert-R: See [my comment on the ticket.](http://jira.yona.nu/browse/YD-265). I don't think we need the ``dosProtectionService`` here.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -272,6 +280,12 @@ static ControllerLinkBuilder getAddUserLinkBuilder()
 		return linkTo(methodOn.addUser(Optional.empty(), null, null, null));
 	}
 
+	static ControllerLinkBuilder getUpdateUserLinkBuilder(UUID userId)
+	{
+		UserController methodOn = methodOn(UserController.class);
+		return linkTo(methodOn.updateUser(Optional.empty(), null, userId, null, null));
```
Bert-R: This effectively makes it impossible for a user to update their account more than x times in the defined time window. That sounds quite reasonable to me.

Looks like we should make the window rather large (e.g. an hour) and the number of attempts relatively high (10 times). That way, normal usage will never hit this, but someone probing numbers will only be able to do 10 per hour.

For {{addUser}}, the scenario is different, as that is global, for all users. If we would use the same window and limit, then we could only add 10 users per hour, which is too low. In that case, it might make more sense to make the window shorter, so we hinder people that probe numbers but we do not block onboarding of considerable numbers of users (e.g. in the break of a Yona promotion meeting). If we make the window shorter (e.g. 1 minute) and the limit low (e.g. 1), we can add a user per minute, which is quite OK, but someone probing numbers can only try 60 per hour.

Makes sense?

core/src/main/resources/application.properties
```diff
@@ -60,6 +60,8 @@ yona.security.passwordLength = 32
 yona.security.dosProtectionEnabled = false
 yona.security.dosProtectionWindow = PT5M
```
Bert-R: See my comment above on ``updateUser``. I question there whether it's sufficient to have a single window setting.

core/src/main/java/nu/yona/server/subscriptions/service/UserServiceException.java
```diff
@@ -43,7 +43,7 @@ public static UserServiceException userExists(String mobileNumber)
 		return new UserServiceException("error.user.exists", mobileNumber);
 	}
 
-	public static UserServiceException usernotCreatedOnBuddyRequest(UUID id)
+	public static UserServiceException userNotCreatedOnBuddyRequest(UUID id)
```
Bert-R: :thumbsup:

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
aristotelos: Yes, it does make sense that the single user has a low frequency which can be asserted over a larger window. Even a window that lasts for a day can be used to restrict the frequency even further.

For new users, one IP might indeed register multiple users if a proxy is used. I hope that in the Yona promotion meetings they have different IP addresses though... The advantage of having a shorter window is that the server does not have to remember `addUser` calls by IP address that long, so can free memory. However, because `addUser` calls do not occur many times, it is also not very much of a problem to have a large window here too. So maybe we could still use the same large window (e.g. 1 hour or 1 day) for both cases.

The advantage of having the same window is that there are less settings to think about. The window setting only influences the memory on the server, not the frequency actually. Only if there are very infrequent calls versus very frequent calls it makes sense to use 2 windows, a shorter window for the very frequently called method and a longer window for the very infrequently called method. Do you think the `addUser` is called much more frequently than `updateUser`?

appservice/src/main/java/nu/yona/server/admin/rest/AdminController.java
aristotelos: Thanks, updated according to your suggestion.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
Bert-R: I forgot we're using the IP address too. With that, I'm not very concerned. One hour would be a reasonable window. If someone ever legitimately runs into, it, they don't need to wait very long for it to expire.

PR #323, opened 2017-03-21
YD-206 Added ISO8601 parse and format tests

aristotelos:


core/src/main/java/nu/yona/server/analysis/service/WeekActivityDto.java
```diff
@@ -79,6 +79,9 @@ public static LocalDate parseDate(String iso8601)
 
 	public static String formatDate(LocalDate sundayDate)
 	{
+		if (sundayDate.getDayOfWeek() != DayOfWeek.SUNDAY)
```
Bert-R: Please add curly braces around the next statement.

PR #324, opened 2017-03-22
YD-416 Add Quartz scheduler

Bert-R:


batchservice/src/main/java/nu/yona/server/batch/service/BatchTaskService.java
```diff
@@ -17,44 +16,41 @@
...
+import nu.yona.server.batch.quartz.SchedulingService;
+import nu.yona.server.batch.quartz.SchedulingService.ScheduleGroup;
+import nu.yona.server.batch.quartz.jobs.PinResetConfirmationCodeSenderQuartzJob;
 import nu.yona.server.exceptions.YonaException;
 import nu.yona.server.util.TimeUtil;
 
 @Service
 public class BatchTaskService
 {
+	private static final String JOB_NAME = "PinResetConfirmationCode";
```
aristotelos: Please inline this or give it a more speaking name.

batchservice/src/main/java/nu/yona/server/batch/quartz/JobController.java
```diff
@@ -0,0 +1,148 @@
...
+
+	@RequestMapping(value = "/{group}/", method = RequestMethod.GET)
+	@ResponseBody
+	public HttpEntity<Resources<JobResource>> getJobsInGroup(@PathVariable String group)
+	{
+		return new ResponseEntity<>(createJobsCollectionResource(jobManagementService.getJobsInGroup(group),
+				getJobsInGroupLinkBuilder(group).withSelfRel()), HttpStatus.OK);
+	}
+
+	@RequestMapping(value = "/{group}/", method = RequestMethod.PUT)
```
aristotelos: As far as I can see, this is the only method used currently, isn't it? And the only `group` used currently is `OTHER`? Why not simplify this and remove the group layer? It would reduce the amount of code.

batchservice/src/main/java/nu/yona/server/batch/quartz/TriggerController.java
```diff
@@ -0,0 +1,147 @@
...
+
+	@RequestMapping(value = "/cron/{group}/", method = RequestMethod.GET)
+	@ResponseBody
+	public HttpEntity<Resources<TriggerResource>> getTriggersInGroup(@PathVariable String group)
+	{
+		return new ResponseEntity<>(createTriggersCollectionResource(triggerManagementService.getTriggersInGroup(group),
+				getTriggersInGroupLinkBuilder(group).withSelfRel()), HttpStatus.OK);
+	}
+
+	@RequestMapping(value = "/cron/{group}/", method = RequestMethod.PUT)
```
aristotelos: Same here, this is the only method used? And only with `group`=`OTHER`? Why not simplify this and remove the group layer?

batchservice/src/main/java/nu/yona/server/batch/quartz/SchedulingService.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.stereotype.Service;
+
+import nu.yona.server.exceptions.YonaException;
+
+@Service
+public class SchedulingService
+{
+	public enum ScheduleGroup
+	{
+		CLEAN_UP, GENERATE, OTHER
```
aristotelos: Currently, the only value of this enum used is `OTHER`. Why do you have the other values there? Why do you even use a group name?

batchservice/src/main/java/nu/yona/server/batch/quartz/AutowiringSpringBeanJobFactory.java
```diff
@@ -0,0 +1,27 @@
+package nu.yona.server.batch.quartz;
+
+import org.quartz.spi.TriggerFiredBundle;
+import org.springframework.beans.factory.config.AutowireCapableBeanFactory;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.scheduling.quartz.SpringBeanJobFactory;
+
+public final class AutowiringSpringBeanJobFactory extends SpringBeanJobFactory implements ApplicationContextAware
```
Bert-R: Done.

batchservice/src/main/java/nu/yona/server/batch/quartz/JobController.java
Bert-R: True. Seen from the back, it would have been better to start with just one group. However, I expect we will see a number of jobs for various purposes. That's why I went down this path. Now that it's there, don't like throwing it away.

batchservice/src/main/java/nu/yona/server/batch/quartz/JobController.java
```diff
@@ -0,0 +1,148 @@
...
+		{
+			super(JobController.class, JobResource.class);
+		}
+
+		@Override
+		public JobResource toResource(JobDto job)
+		{
+			JobResource jobResource = instantiateResource(job);
+			addSelfLink(jobResource);
+			addEditLink(jobResource);/* always editable */
```
Bert-R: Fair point. I wanted to bring the editing at the job level, but then implemented it at the group level. I'm removing the ``addEditLink`` method.

batchservice/src/main/java/nu/yona/server/batch/quartz/TriggerController.java
Bert-R: Same here.

batchservice/src/main/java/nu/yona/server/batch/quartz/TriggerController.java
```diff
@@ -0,0 +1,147 @@
...
+		{
+			super(TriggerController.class, TriggerResource.class);
+		}
+
+		@Override
+		public TriggerResource toResource(CronTriggerDto trigger)
+		{
+			TriggerResource triggerResource = instantiateResource(trigger);
+			addSelfLink(triggerResource);
+			addEditLink(triggerResource);/* always editable */
```
Bert-R: Same here.

batchservice/src/main/java/nu/yona/server/batch/service/BatchTaskService.java
Bert-R: Done.

batchservice/src/main/java/nu/yona/server/batch/quartz/SchedulingService.java
Bert-R: See my earlier comment.

Bert-R:
Pushed an update that addresses several of your comments.

PR #325, opened 2017-03-23
YD-428 Remove Spring Batch admin UI

Bert-R:
It caused many issues already and stopped working recently. Dropping it now. For now, we can monitor by querying the database. Later, we might add a custom UI.


aristotelos:
Looks great. Will merge.

PR #327, opened 2017-03-28
YD-417 Prevent identification

aristotelos:
Moved lastStatusChangeTime to Buddy, encrypted
Set creation time of mandatory goals to 1 January 2017 (resetting to now in tests to avoid clutter)
Refactored Buddy/Goal a little

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
@@ -295,7 +295,7 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
 		dateTime.get(ChronoField.MINUTE_OF_DAY)/15
 	}
 
-	def setMandatoryGoalsToNow(User user)
+	def setCreationTimeOfMandatoryGoalsToNow(User user)
 	{
 		def mandatoryGoals = user.goals.find { it.activityCategoryUrl == GAMBLING_ACT_CAT_URL }
 		mandatoryGoals.each
--- a/appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
@@ -295,6 +295,18 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
 		dateTime.get(ChronoField.MINUTE_OF_DAY)/15
 	}
 
+	def setMandatoryGoalsToNow(User user)
+	{
+		def mandatoryGoals = user.goals.find { it.activityCategoryUrl == GAMBLING_ACT_CAT_URL }
+		mandatoryGoals.each
+		{ goal ->
+			goal.creationTime = YonaServer.now
+			def response = appService.updateGoal(user, goal.url, goal)
+			assert response.status == 200
+		}
+		return appService.reloadUser(user)
+	}
+
 	void assertWeekOverviewBasics(response, numberOfReportedGoals, expectedTotalElements, expectedPageSize = 2)
 	{
 		assert response.status == 200
--- a/appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
@@ -297,7 +297,7 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
 
 	def setCreationTimeOfMandatoryGoalsToNow(User user)
 	{
-		def mandatoryGoals = user.goals.find { it.activityCategoryUrl == GAMBLING_ACT_CAT_URL }
+		def mandatoryGoals = user.goals.findAll { it.activityCategoryUrl == GAMBLING_ACT_CAT_URL }
 		mandatoryGoals.each
 		{ goal ->
 			goal.creationTime = YonaServer.now
--- a/appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
@@ -304,7 +304,6 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
 			def response = appService.updateGoal(user, goal.url, goal)
 			assert response.status == 200
 		}
-		return appService.reloadUser(user)
 	}
 
 	void assertWeekOverviewBasics(response, numberOfReportedGoals, expectedTotalElements, expectedPageSize = 2)
--- a/appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
@@ -67,7 +67,7 @@ class EditGoalsTest extends AbstractAppServiceIntegrationTest
 		gamblingGoals.size() == 1
 		gamblingGoals[0]."@type" == "BudgetGoal"
 		!gamblingGoals[0]._links.edit //mandatory goal
-		assertEquals(gamblingGoals[0].creationTime, creationTime)
+		assertEquals(gamblingGoals[0].creationTime, '20170101120000')
 		def newsGoals = filterGoals(response, NEWS_ACT_CAT_URL)
 		newsGoals.size() == 1
 		newsGoals[0]."@type" == "BudgetGoal"
--- a/appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
@@ -7,7 +7,7 @@
 package nu.yona.server
 
 import java.time.Duration
-import java.time.ZoneId
+import java.time.ZoneOffset
 import java.time.ZonedDateTime
 
 import groovy.json.*
@@ -68,7 +68,7 @@ class EditGoalsTest extends AbstractAppServiceIntegrationTest
 		gamblingGoals.size() == 1
 		gamblingGoals[0]."@type" == "BudgetGoal"
 		!gamblingGoals[0]._links.edit //mandatory goal
-		assertEquals(gamblingGoals[0].creationTime, ZonedDateTime.of(2017,1,1,12,0,0,0, ZoneId.of("Z")))
+		assertEquals(gamblingGoals[0].creationTime, ZonedDateTime.of(2017,1,1,12,0,0,0, ZoneOffset.UTC))
 		def newsGoals = filterGoals(response, NEWS_ACT_CAT_URL)
 		newsGoals.size() == 1
 		newsGoals[0]."@type" == "BudgetGoal"
--- a/appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
@@ -7,6 +7,7 @@
 package nu.yona.server
 
 import java.time.Duration
+import java.time.ZoneId
 import java.time.ZonedDateTime
 
 import groovy.json.*
@@ -67,7 +68,7 @@ class EditGoalsTest extends AbstractAppServiceIntegrationTest
 		gamblingGoals.size() == 1
 		gamblingGoals[0]."@type" == "BudgetGoal"
 		!gamblingGoals[0]._links.edit //mandatory goal
-		assertEquals(gamblingGoals[0].creationTime, '20170101120000')
+		assertEquals(gamblingGoals[0].creationTime, ZonedDateTime.of(2017,1,1,12,0,0,0, ZoneId.of("Z")))
 		def newsGoals = filterGoals(response, NEWS_ACT_CAT_URL)
 		newsGoals.size() == 1
 		newsGoals[0]."@type" == "BudgetGoal"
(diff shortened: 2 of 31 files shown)


dbinit/src/main/liquibase/updates/changelog-0004-yd-417.yml
```diff
@@ -0,0 +1,18 @@
...
+        columns:
+        - column:
+            name: last_status_change_time
+            type: varchar(255)
+        tableName: buddies
+- changeSet:
+    id: 1490127260476-21
+    author: avdijk (generated)
+    changes:
+    - dropColumn:
```
Bert-R: This is not going to work. It would cause all users to loose their "last status change time" for their buddies.
We need to retain both columns till we have migrated the data for all users from their ``BuddyAnonymized`` entity to their ``Buddy`` entity.

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
```diff
@@ -295,6 +295,17 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
 		dateTime.get(ChronoField.MINUTE_OF_DAY)/15
 	}
 
+	def setMandatoryGoalsToNow(User user)
```
Bert-R: Please rename into ``setCreationTimeForMandatoryGoalsToNow``

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
```diff
@@ -295,6 +295,17 @@ abstract class AbstractAppServiceIntegrationTest extends Specification
...
+	{
+		def mandatoryGoals = user.goals.find { it.activityCategoryUrl == GAMBLING_ACT_CAT_URL }
```
Bert-R: The ``find`` operation returns just one object, so the method can be this simple:
```
def goal = user.goals.find { it.activityCategoryUrl == GAMBLING_ACT_CAT_URL }
goal.creationTime = YonaServer.now
def response = appService.updateGoal(user, goal.url, goal)
assert response.status == 200
```

appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
```diff
@@ -67,7 +68,7 @@ class EditGoalsTest extends AbstractAppServiceIntegrationTest
 		gamblingGoals.size() == 1
 		gamblingGoals[0]."@type" == "BudgetGoal"
 		!gamblingGoals[0]._links.edit //mandatory goal
-		assertEquals(gamblingGoals[0].creationTime, creationTime)
+		assertEquals(gamblingGoals[0].creationTime, ZonedDateTime.of(2017,1,1,12,0,0,0, ZoneId.of("Z")))
```
Bert-R: There is a constant for UTC: ``ZoneOffset.UTC``.

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
aristotelos: Done

appservice/src/intTest/groovy/nu/yona/server/AbstractAppServiceIntegrationTest.groovy
aristotelos: Changed to `findAll`

appservice/src/intTest/groovy/nu/yona/server/EditGoalsTest.groovy
aristotelos: Thanks, replaced

aristotelos:
I have added a migration mechanism now. It has a simple integration test on the migration functionality, but the actual migration of `lastStatusChangeTime` has to be manually tested still (I will try to do that on a local DB or so).

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -114,6 +115,9 @@
 	@Autowired(required = false)
 	private WhiteListedNumberService whiteListedNumberService;
 
+	@Autowired(required = false)
+	private MigratePrivateUserDataService migratePrivateUserDataService;
```
Bert-R: Please rename class and member into ``PrivateUserDataMigrationService``. That's a more common naming pattern.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -725,6 +760,11 @@ void assertValidUserFields(UserDto userResource, UserPurpose userPurpose)
 				throw InvalidDataException.blankEmailAddress();
 			}
 			assertValidEmailAddress(userResource.getEmailAddress());
+
+			if (!userResource.getPrivateData().getGoals().isEmpty())
```
Bert-R: Good addition!

core/src/main/java/nu/yona/server/subscriptions/service/migration/EncryptBuddyLastStatusChangeTime.java
```diff
@@ -0,0 +1,19 @@
...
+import nu.yona.server.subscriptions.service.MigratePrivateUserDataService;
+
+public class EncryptBuddyLastStatusChangeTime extends MigratePrivateUserDataService.MigrationStep
+{
+	@SuppressWarnings("deprecation")
+	@Override
+	public void upgrade(User userEntity)
+	{
+		userEntity.getBuddies()
+				.forEach(buddy -> buddy.setLastStatusChangeTime(buddy.getBuddyAnonymized().getLastStatusChangeTime()));
```
Bert-R: Please also nullify the time stamp on {{BuddyAnonymized}}. With that, we prevent identity matching (the purpose of this ticket) and we can see how far we are with the migration.

dbinit/src/main/liquibase/updates/changelog-0004-yd-417.yml
```diff
@@ -0,0 +1,23 @@
...
+        tableName: buddies
+- changeSet:
+    id: 1491330640527-1
+    author: avdijk (generated)
+    changes:
+    - addColumn:
+        columns:
+        - column:
+            constraints:
+              nullable: false
```
Bert-R: Shouldn't you define a default value for this one for the records that exist when adding the column?

core/src/main/java/nu/yona/server/exceptions/InvalidDataException.java
```diff
@@ -85,4 +85,9 @@ public static InvalidDataException excessEmailAddress()
 	{
 		return new InvalidDataException("error.user.email.address.not.supported");
 	}
+
+	public static InvalidDataException excessGoals()
```
aristotelos: I think the current name shows the intent better. The method above, `excessEmailAddress`, has the same purpose and message ID.

core/src/main/java/nu/yona/server/subscriptions/entities/Buddy.java
```diff
@@ -113,12 +120,20 @@ public Status getReceivingStatus()
 
 	public void setReceivingStatus(Status status)
 	{
-		getBuddyAnonymized().setReceivingStatus(status);
+		BuddyAnonymized buddyAnonymized = getBuddyAnonymized();
+		buddyAnonymized.setReceivingStatus(status);
+		BuddyAnonymized.getRepository().save(buddyAnonymized);
```
aristotelos: The `BuddyAnonymized` was edited here, but not saved anywhere. Could it be the case that this was done automatically by JPA?

core/src/main/java/nu/yona/server/subscriptions/service/MigratePrivateUserDataService.java
```diff
@@ -0,0 +1,82 @@
...
+import org.springframework.stereotype.Service;
+
+import nu.yona.server.exceptions.YonaException;
+import nu.yona.server.subscriptions.entities.User;
+import nu.yona.server.subscriptions.service.migration.EncryptBuddyLastStatusChangeTime;
+
+@Service
+public class MigratePrivateUserDataService
+{
+	public static abstract class MigrationStep
```
aristotelos: I always choose an abstract class as the default option, because it is easier to refactor later.

core/src/main/java/nu/yona/server/subscriptions/service/migration/EncryptBuddyLastStatusChangeTime.java
aristotelos: Done

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
aristotelos: Done, indeed sounds better

core/src/main/java/nu/yona/server/subscriptions/service/MigratePrivateUserDataService.java
```diff
@@ -0,0 +1,82 @@
...
+		public abstract void upgrade(User userEntity);
+	}
+
+	private static List<Class<? extends MigrationStep>> migrationSteps = Arrays.asList(EncryptBuddyLastStatusChangeTime.class);
+
+	@Transactional
+	void upgrade(User userEntity)
+	{
+		int version = userEntity.getPrivateDataMigrationVersion();
+		for (int newVersion = version + 1; newVersion <= getCurrentVersion(); newVersion++)
```
aristotelos: Done

core/src/main/java/nu/yona/server/exceptions/InvalidDataException.java
Bert-R: The scenario really is that e-mail addresses are not supported here. Excess means "surpassing usual, proper, or specified limits". In this case, the limit is "zero". In that case, it's more clear to say "goals not supported", as the message ID says.

core/src/main/java/nu/yona/server/subscriptions/entities/Buddy.java
Bert-R: No, JPA doesn't do that. I guess it just happened to occur elsewhere. It's better to do it explictely, as you do now.

dbinit/src/main/liquibase/updates/changelog-0004-yd-417.yml
aristotelos: Why? If it is `NULL`, this indicates it is an old record.

core/src/main/java/nu/yona/server/subscriptions/service/migration/EncryptBuddyLastStatusChangeTime.java
```diff
@@ -4,16 +4,24 @@
...
-				.forEach(buddy -> buddy.setLastStatusChangeTime(buddy.getBuddyAnonymized().getLastStatusChangeTime()));
+		userEntity.getBuddies().forEach(buddy -> upgradeBuddy(buddy));
+	}
+
+	@SuppressWarnings("deprecation")
+	private void upgradeBuddy(Buddy buddy)
+	{
+		BuddyAnonymized buddyAnonymized = buddy.getBuddyAnonymized();
+		buddy.setLastStatusChangeTime(buddyAnonymized.getLastStatusChangeTime());
+		buddyAnonymized.clearLastStatusChangeTime();
```
Bert-R: Shouldn't you explicitly save this?

core/src/main/java/nu/yona/server/subscriptions/service/PrivateUserDataMigrationService.java
```diff
@@ -0,0 +1,82 @@
...
+import org.springframework.stereotype.Service;
+
+import nu.yona.server.exceptions.YonaException;
+import nu.yona.server.subscriptions.entities.User;
+import nu.yona.server.subscriptions.service.migration.EncryptBuddyLastStatusChangeTime;
+
+@Service
+public class PrivateUserDataMigrationService
+{
+	public static abstract class MigrationStep
```
Bert-R: > I always choose an abstract class as the default option, because it is easier to refactor later.

That's an uncommon approach. Given that you can inherit from just one class, the common approach is to try not to demand a base class. Given that this isn't necessary here, it's better not to do it. In Java, we can nowadays have default methods, which makes refactoring even easier.

dbinit/src/main/liquibase/updates/changelog-0004-yd-417.yml
Bert-R: In Java, it's an ``int``. What value will that ``int`` get? I'd rather say the default is zero rather than null.

core/src/main/java/nu/yona/server/subscriptions/service/migration/EncryptBuddyLastStatusChangeTime.java
aristotelos: It is done implicitly by JPA :). But I will do so.

dbinit/src/main/liquibase/updates/changelog-0004-yd-417.yml
aristotelos: OK sorry. The default value indeed is `0` for existing records. I will add the default here explicitly indeed.

core/src/main/java/nu/yona/server/exceptions/InvalidDataException.java
aristotelos: OK, you're right, will rename.

core/src/main/java/nu/yona/server/subscriptions/service/PrivateUserDataMigrationService.java
aristotelos: Done

aristotelos:
Reworked now, thanks for your suggestions! I have also performed a manual test which was successful.

Bert-R:
Great, so there's nothing left. I'm going to merge it.

PR #328, opened 2017-03-28
YD-424 Implemented signing of the mobile config

Bert-R:
The current implementation only signs when explicitly requested to do so through a query parameter. When the iOS team is happy with the result, we'll make signing the default.

Along with this, cleaned up some old properties.

### DIFF
--- a/appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
+++ b/appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
@@ -71,6 +71,7 @@ import nu.yona.server.exceptions.YonaException;
 import nu.yona.server.goals.rest.GoalController;
 import nu.yona.server.goals.service.GoalDto;
 import nu.yona.server.messaging.rest.MessageController;
+import nu.yona.server.properties.AppleMobileConfigSigningProperties;
 import nu.yona.server.properties.YonaProperties;
 import nu.yona.server.rest.Constants;
 import nu.yona.server.rest.ErrorResponseDto;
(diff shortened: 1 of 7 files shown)


appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -146,27 +147,50 @@
...
 		try (CryptoSession cryptoSession = CryptoSession.start(password, () -> userService.canAccessPrivateData(userId)))
 		{
-			return new ResponseEntity<>(getUserSpecificAppleMobileConfig(userService.getPrivateUser(userId)), headers,
+			return new ResponseEntity<>(getUserSpecificAppleMobileConfig(userService.getPrivateUser(userId), mustSign), headers,
 					HttpStatus.OK);
 		}
 	}
 
-	private byte[] getUserSpecificAppleMobileConfig(UserDto privateUser)
+	@RequestMapping(value = "/{userId}/apple.mobileconfig", method = RequestMethod.GET)
```
aristotelos: Why do you need this overload? If you just remove the `params = { "sign" }` in the method above, it will match both.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -146,27 +147,50 @@
...
+
+	private byte[] signIfRequested(boolean mustSign, byte[] unsignedMobileconfig)
+	{
+		if (mustSign)
+		{
+			AppleMobileConfigSigningProperties properties = yonaProperties.getSecurity().getAppleMobileConfigSigning();
+			String signingCertificateFile = properties.getSigningCertificateFile();
+			String signingKeyFile = properties.getSigningKeyFile();
+			String password = properties.getPassword();
+			return new Signer(signingCertificateFile, signingKeyFile, password).sign(unsignedMobileconfig);
```
aristotelos: Qua performance it would be good to cache the instance of the `Signer` with the certificate/key files preloaded. Can't this be done with `@Autowired`?

core/src/main/resources/application.properties
```diff
@@ -66,6 +56,11 @@ yona.security.maxCreateUserAttemptsPerTimeWindow = 2
 yona.security.maxUpdateUserAttemptsPerTimeWindow = 2
 yona.security.corsAllowed = true
 
+yona.security.appleMobileConfigSigning.isSigningEnabled = true
+yona.security.appleMobileConfigSigning.signingCertificateFile = smime.crt
+yona.security.appleMobileConfigSigning.signingKeyFile = smime.key
+yona.security.appleMobileConfigSigning.password = Yona1
```
aristotelos: This should be stored in a key container later I guess?

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
Bert-R: I added this because otherwise it would add the ``?sign=...`` parameter to the ``yona:appleMobileConfig`` link.

The ``sign`` query parameter will be taken out when we close this feature, so this is just temporarily.

core/src/main/resources/application.properties
Bert-R: Not sure yet. We need to talk to Richard Clark to see what's the best way here.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
Bert-R: We can't cache much, see [this](http://bouncy-castle.1462172.n4.nabble.com/is-CMSSignedDataGenerator-object-thread-safe-td4655398.html). The certificate can be cached (see [this](http://bouncy-castle.1462172.n4.nabble.com/Signing-an-Verify-X509Certificate-Hash-Thread-Safe-operations-td1467650.html)) and I'm confident about the private key too.

I'll take care of that.

PR #331, opened 2017-04-01
YD-434 Embed Swagger UI

Bert-R:
By embedding Swagger UI, we are not longer dependent on the availability of http://petstore.swagger.io, nor do changes there immediately impact us. The Swagger spec was updated to make it work with the current version of Swagger UI. The most notable update is that the parameters are now defined as reusable objects. This makes the spec considerably shorter and it resolves the issue that somehow parameters defined for the path (rather than the operation) weren't recognized anymore.

With this update, it is possible to go to any installation of the Yona server and access Swagger UI on it: http://localhost:8082/swagger-ui/index.html


appservice/src/main/resources/swagger/swagger-spec.yaml
```diff
@@ -1194,25 +928,13 @@ paths:
...
-        type: string
-      - name: Yona-Password
-        in: header
-        description: The device password
-        required: true
-        type: string
     get:
       summary: Get a message
       description: Fetches all messages of the user with the given ID
+      parameters:
```
aristotelos: Please add those parameters to the `delete` verb as well.

appservice/src/main/resources/swagger/swagger-spec.yaml
```diff
@@ -43,21 +93,12 @@ paths:
...
-        required: true
-        type: string
     get:
       summary: Get a user
       description: Fetches the user identified by the given ID
       parameters:
-        - name: Yona-Password
-          in: header
-          description: The device password
-          required: false
```
aristotelos: This has changed to `required: true` now.

Bert-R: Done.

appservice/src/main/resources/swagger/swagger-spec.yaml
Bert-R: Done

appservice/src/main/java/nu/yona/server/rest/SwaggerSpecController.java
```diff
@@ -22,7 +23,7 @@
 import nu.yona.server.exceptions.YonaException;
 
 @Controller
-@RequestMapping(value = "/swagger/swagger-spec.yaml", produces = { "text/yaml" })
+@RequestMapping(value = "/swagger/swagger-spec.yaml", produces = MediaType.APPLICATION_JSON_VALUE)
```
aristotelos: You could add a comment here.

appservice/src/main/resources/swagger/swagger-spec.yaml
```diff
@@ -98,7 +98,11 @@ paths:
       description: Fetches the user identified by the given ID
       parameters:
         - $ref: "#/parameters/userID"
-        - $ref: "#/parameters/yonaPassword"
+        - name: Yona-Password
```
aristotelos: You could add a comment here why it is not required.

PR #332, opened 2017-04-05
YD-433 Fixed activity comment message delete issue

Bert-R:
* Break the cyclic relations before performing the delete
* Use explicit delete instead of cascading
* Brought the cycle-breaking for the self-reference of the thread head under the same umbrella
* Added integration tests


appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
```diff
@@ -0,0 +1,552 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Stichting Yona Foundation
```
aristotelos: 2017...

core/src/main/java/nu/yona/server/messaging/entities/Message.java
```diff
@@ -167,7 +169,8 @@ public void setRead(boolean isRead)
 	 * <ul>
 	 * <li>If isSentItem == true, it is logically the target user</li>
 	 * <li>The sender of this message</li>
-	 * <li>In case of a system-originated message (e.g. a goal conflict message), it is the user for which this message is sent</li>
+	 * <li>In case of a system-originated message (e.g. a goal conflict message), it is the user for which this message is
```
aristotelos: This is a whitespace code style change.

core/src/main/java/nu/yona/server/analysis/entities/ActivityCommentMessage.java
```diff
@@ -76,6 +80,10 @@ public void clearThreadHeadSelfReference()
 	public void setBuddyMessage(ActivityCommentMessage message)
 	{
 		buddyMessage = message;
+		if (message == null)
```
aristotelos: Is this actually occurring now? I think `prepareForDelete` already sets to `null`?

appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
```diff
@@ -0,0 +1,552 @@
...
+
+		cleanup:
+		appService.deleteUser(richard)
+		appService.deleteUser(bob)
+	}
+
+	def 'Richard can delete a comment of Bob on which nobody replied'()
+	{
+		given:
+		/** Begin of standard test setup for message delete tests **/
```
aristotelos: Can you extract a method here please?

appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
```diff
@@ -0,0 +1,552 @@
...
+
+		then:
+		responseAddMessage.status == 200
+
+		cleanup:
+		appService.deleteUser(richard)
+		appService.deleteUser(bob)
+	}
+
+	def 'Richard can delete a comment of Bob on which Richard already replied'()
```
aristotelos: You could state something about the result, i.e. that the message threads are removed on both sides.

appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
Bert-R: Done.

appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
Bert-R: How about "Richard successfully deletes a comment..."? Otherwise, the test name becomes very long.

appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
Bert-R: No, I can't, as the variables are used in the assertions. I've considered that, but didn't see how I could usefully do that.

core/src/main/java/nu/yona/server/messaging/entities/Message.java
Bert-R: Yes, line length. I don't know how this occurs, as we both use the same code style, but I occasionally see this.

Reverting it doesn't help, as it'll hapen again next time.

appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
aristotelos: OK, you can leave it unaltered in that case

core/src/main/java/nu/yona/server/analysis/entities/ActivityCommentMessage.java
Bert-R: Indeed. Removed it.

appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
aristotelos: Could be done using a separate test fixture with common test setup and teardown. But that can be done later too.

core/src/main/java/nu/yona/server/messaging/entities/Message.java
aristotelos: I agree, just noting because you asked me :)

Bert-R: I see. The one that worries me more is with the curly braces. That was a weird one.

PR #333, opened 2017-04-06
YD-435 Moved certificate info out of JARs

Bert-R:
The following files are not resources in the JAR any longer but taken from the resources folder on the file system:
* appservice/src/main/resources/static/ssl/rootcert.cer
* appservice/src/main/resources/static/vpn/profile.ovpn
* core/src/main/resources/templates/other/apple.mobileconfig.xml

The install scripts now pull these files from GitHub and after that copy node specific files from the specified folder (currently /opt/ope-cloudbees/yona/resources as defined in the Jenkinsfile).

Extended the integration tests to verify the files are properly served.


kti-richard:
Looks like it does what was looking for - Thanks.    I will set those files up as k8s config maps.

appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
```diff
@@ -296,6 +296,26 @@ class UserTest extends AbstractAppServiceIntegrationTest
...
+		appService.deleteUser(richard)
+	}
+
+	def 'Retrieve SSL root certificate'()
+	{
+		given:
+		User richard = addRichard()
+
+		when:
+		assert richard.appleMobileConfig
```
aristotelos: Why this assert?

appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
```diff
@@ -296,6 +296,26 @@ class UserTest extends AbstractAppServiceIntegrationTest
...
+		def responseAppleMobileConfig = appService.yonaServer.restClient.get(path: richard.sslRootCertUrl, headers: ["Yona-Password":richard.password])
```
aristotelos: I guess this should be `responseSslRootCert`?

appservice/src/main/java/nu/yona/server/MimeTypeConfigurer.java
```diff
@@ -19,8 +16,7 @@
 	public void customize(ConfigurableEmbeddedServletContainer container)
 	{
 		MimeMappings mappings = new MimeMappings(MimeMappings.DEFAULT);
-		mappings.add("ovpn", "application/x-openvpn-profile"); // see vpn/profile.ovpn
-		mappings.add("cer", "application/pkix-cert"); // see ssl/rootcert.cer
+		mappings.add("cer", "application/pkix-cert"); // see resources/rootcert.cer
```
aristotelos: Is this still necessary, now you are using `FileSystemResource`?

core/src/main/java/nu/yona/server/properties/AppleMobileConfigSigningProperties.java
```diff
@@ -6,11 +6,22 @@
 
 public class AppleMobileConfigSigningProperties
```
aristotelos: This could be renamed to `AppleMobileConfigProperties`.

core/src/test/java/nu/yona/server/templates/TemplateTest.java
```diff
@@ -61,19 +61,6 @@ public void testAppleAppSiteAssociation()
...
-	public void testAppleMobileConfig()
-	{
-		Context ctx = ThymeleafUtil.createContext();
-		String ldapUsername = "DummyLdapUserName";
-		String ldapPassword = "DummyLdapPassword";
-		ctx.setVariable("ldapUsername", ldapUsername);
-		ctx.setVariable("ldapPassword", ldapPassword);
-
-		String result = otherTemplateEngine.process("apple.mobileconfig.xml", ctx);
-		assertThat(result, containsString("<string>" + ldapUsername + "\\n" + ldapPassword + "</string>"));
```
aristotelos: This test asserted that the `\n` was present between the LDAP username van LDAP password. Have you checked that this is still the case? Maybe move it to the integration test above?

resources/smime.key
```diff
@@ -0,0 +1,54 @@
+-----BEGIN RSA PRIVATE KEY-----
```
aristotelos: This file should not be on GitHub, isn't it?

appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
Bert-R: This is a copy/paste error. I'll correct it.

appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
Bert-R: This is a copy/paste error too. I'll correct it.

appservice/src/main/java/nu/yona/server/MimeTypeConfigurer.java
Bert-R: I left this one in as it is generic and could be of use later.

core/src/main/java/nu/yona/server/properties/AppleMobileConfigSigningProperties.java
Bert-R: Will do.

core/src/test/java/nu/yona/server/templates/TemplateTest.java
Bert-R: Yes, will add it.

resources/smime.key
Bert-R: It doesn't matter. This is a useless key, used for our test server only.

aristotelos:
Yes, sure.

Bert-R:
Pushed an update that addresses all issues.

PR #335, opened 2017-04-08
YD-424 Now use a PKCS 12

Bert-R:
This is what Apple provides and it's easier to consume this than to build .crt and .key files out of the .p12 of Apple.

Along with this removed the isSigningEnabled flag. We're shipping a dummy .p12 file now, so we can simply enable signing always.


appservice/src/main/java/nu/yona/server/AppServiceApplication.java
```diff
@@ -83,60 +80,81 @@ public void addCorsMappings(CorsRegistry registry)
...
 		{
-			logger.info("Apple mobile config signing certificate not loaded, as signing is disabled");
-			return null;
+			throw YonaException.unexpected(e);
 		}
-		return loadCertificateFromFile(properties.getSigningCertificateFile());
 	}
 
-	private X509Certificate loadCertificateFromFile(String signingCertificateFile)
+	private void assertKeyStoreContent(KeyStore keyStore)
```
aristotelos: Technically, this is no assert, so I would prefer the name `verifyKeyStoreContent`.

Bert-R: If you mean it's not a Java or JUnit assert, then it's true, but this method asserts the key store content (for correctness). All of these methods are nowadays called ``assertXxxx``. See ``UserService`` for examples.

aristotelos: Agreed

PR #341, opened 2017-04-19
YD-442 Remove BuddyAnonymized on disconnect

Bert-R:
Along with this added extra logging


Bert-R:
OK, just go ahead and merge it in such a case.
There is no need to wait for anything.

PR #343, opened 2017-04-21
YD-444 Improve error logging

Bert-R:
* For unhandled exceptions, include the request URL
* Concatenate the query string to the URL, when present


appservice/src/main/java/nu/yona/server/subscriptions/rest/PinResetRequestController.java
```diff
@@ -1,6 +1,6 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2017 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License, v.
- * 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ * Copyright (c) 2016, 2017 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
```
aristotelos: (Note: formatting difference)

PR #344, opened 2017-04-22
YD-443 Fixed IllegalStateException

Bert-R:
This sequence went wrong:
* Buddy disconnected
* User removed goal

While this one was successful:
* Buddy disconnected
* User retrieved messages (and along with that processed buddy disconnects)
* User removed goal

This is now fixed and a test case is added.

### DIFF
--- a/core/src/main/java/nu/yona/server/goals/service/GoalService.java
+++ b/core/src/main/java/nu/yona/server/goals/service/GoalService.java
@@ -21,6 +21,7 @@ import nu.yona.server.goals.entities.GoalChangeMessage;
 import nu.yona.server.messaging.service.MessageService;
 import nu.yona.server.subscriptions.entities.User;
 import nu.yona.server.subscriptions.entities.UserAnonymized;
+import nu.yona.server.subscriptions.service.BuddyService;
 import nu.yona.server.subscriptions.service.UserAnonymizedDto;
 import nu.yona.server.subscriptions.service.UserAnonymizedService;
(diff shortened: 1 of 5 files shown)


core/src/main/java/nu/yona/server/goals/service/GoalService.java
```diff
@@ -225,8 +229,7 @@ private void deleteGoalRelatedMessages(UserAnonymized userAnonymizedEntity, Goal
 	private void deleteGoalConflictMessagesForGoal(UserAnonymized userAnonymizedEntity, Goal goalEntity)
 	{
 		userAnonymizedEntity.getAnonymousDestination().removeGoalConflictMessages(goalEntity);
-		userAnonymizedEntity.getBuddiesAnonymized()
-				.forEach(ba -> ba.getUserAnonymized().getAnonymousDestination().removeGoalConflictMessages(goalEntity));
+		buddyService.getBuddyDestinations(userAnonymizedEntity).forEach(d -> d.removeGoalConflictMessages(goalEntity));
```
aristotelos: When there are messages sent to buddies that are now disconnected, are they removed ever? Or are they just kept?

Bert-R: These are removed when the buddy disconnects or unsubscribes.

aristotelos: OK, I guess in the `BuddyService.removeBuddyInfoForBuddy` method then. The name is not so easy to understand...

PR #348, opened 2017-04-27
YD-448 Addressed some code issues flagged by SonarCube

Bert-R:
* Removed duplicate Constants.PASSWORD_HEADER
* Removed unnecessary parenthesis and replaced method call with method reference
* Added 'L' to long constants
* Reinterrupted the thread instead of totally ignoring the interrupt

Along with this changed the Sonar version number to our regular convention: build-<number>


Bert-R:
I merged this one myself, to see how the SonarQube analysis works when correcting issues.

PR #354, opened 2017-05-09
YD-455 Small code and copyright corrections

Bert-R:
* Code corrections were flagged by SonarQube
* Copyrights correct automatically through Eclipse plug-in

### DIFF
--- a/core/src/main/java/nu/yona/server/exceptions/EmailException.java
+++ b/core/src/main/java/nu/yona/server/exceptions/EmailException.java
@@ -1,9 +1,6 @@
 /*******************************************************************************
- * Copyright (c) 2016 Stichting Yona Foundation
- *
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ * Copyright (c) 2016 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License, v.
+ * 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
  *******************************************************************************/
 package nu.yona.server.exceptions;
 
@@ -13,11 +10,6 @@ package nu.yona.server.exceptions;
 public class EmailException extends YonaException
 {
 	private static final long serialVersionUID = -7917208280838423613L;
-
-	private EmailException(String messageId, Object... parameters)
-	{
-		super(messageId, parameters);
-	}
 
 	private EmailException(Throwable t, String messageId, Object... parameters)
 	{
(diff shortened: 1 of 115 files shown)


core/src/main/java/nu/yona/server/exceptions/EmailException.java
```diff
@@ -1,9 +1,6 @@
 /*******************************************************************************
- * Copyright (c) 2016 Stichting Yona Foundation
```
aristotelos: There are some accidental format changes like this. Just to let you know.

core/src/main/java/nu/yona/server/goals/service/ActivityCategoryService.java
```diff
@@ -225,9 +225,9 @@ private String getName(ActivityCategory entity)
 
 	/**
 	 * This method calls flush on the activity category repository. It shouldn't be necessary to do this, but the implicit flush
-	 * (done just before JPA sends a query to the database) causes a
```
aristotelos: Formatting change here as well.

core/src/main/java/nu/yona/server/messaging/service/DisclosureResponseMessageDto.java
```diff
@@ -51,8 +51,7 @@ public String getType()
 	@Override
 	public Set<String> getPossibleActions()
 	{
-		Set<String> possibleActions = super.getPossibleActions();
-		return possibleActions;
+		return super.getPossibleActions();
```
aristotelos: I think SonarQube will warn about this again :).

core/src/main/java/nu/yona/server/sms/SmsTemplate.java
```diff
@@ -0,0 +1,21 @@
+package nu.yona.server.sms;
+
+public enum SmsTemplate
+{
+	ADD_USER_NUMBER_CONFIRMATION("add-user-number-confirmation"), CHANGED_USER_NUMBER_CONFIRMATION(
```
aristotelos: This formatting is strange...

core/src/main/java/nu/yona/server/exceptions/EmailException.java
Bert-R: These are reformatted when the file is updated. That's normal.

core/src/main/java/nu/yona/server/goals/service/ActivityCategoryService.java
Bert-R: This seems a slight difference in line length. That might be the cause of the accidental formatting difference. Please keep flagging them.

core/src/main/java/nu/yona/server/messaging/service/DisclosureResponseMessageDto.java
Bert-R: Very true :)
I'll remove the method.

Bert-R:
Removed the useless method.

PR #356, opened 2017-05-16
YD-452 Reduce entity queries

Bert-R:


core/src/main/java/nu/yona/server/subscriptions/entities/UserAnonymized.java
```diff
@@ -37,12 +41,14 @@ public static UserAnonymizedRepository getRepository()
...
 	private LocalDate lastMonitoredActivityDate;
 
-	@OneToOne
+	@OneToOne(fetch = FetchType.LAZY)
 	private MessageDestination anonymousDestination;
 
 	@OneToMany(mappedBy = "userAnonymized", cascade = CascadeType.ALL, orphanRemoval = true)
 	@Where(clause = "end_time is null") // The history items have the user anonymized ID set, so they would appear in this
 										// collection if not explicitly excluded
+	@Fetch(FetchMode.SELECT)
```
aristotelos: This is the default, so does not have to be here. See [Hibernate documentation](http://docs.jboss.org/hibernate/core/4.3/manual/en-US/html/ch20.html#performance-fetching-lazy).

core/src/main/java/nu/yona/server/subscriptions/entities/User.java
```diff
@@ -53,13 +53,13 @@ public static UserRepository getRepository()
 
 	private boolean isCreatedOnBuddyRequest;
 
-	@OneToOne(cascade = CascadeType.ALL, fetch = FetchType.EAGER, orphanRemoval = true)
+	@OneToOne(cascade = CascadeType.ALL, fetch = FetchType.LAZY, orphanRemoval = true)
```
aristotelos: This is the default, so does not have to be here, see [Hibernate docs](http://docs.jboss.org/hibernate/core/4.3/manual/en-US/html/ch20.html#performance-fetching-lazy).

core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
```diff
@@ -170,9 +170,15 @@ public UUID getUserAnonymizedId()
 
 	public Set<Buddy> getBuddiesRelatedToRemovedUsers()
 	{
+		loadAllBuddyUsers();
 		return buddies.stream().filter(b -> b.getUser() == null).collect(Collectors.toSet());
 	}
 
+	private void loadAllBuddyUsers()
```
aristotelos: Maybe the name `loadAllBuddyUsersAtOnce` is more speaking. The same for the other methods below.

Bert-R:
Here is an overview of the queries before and after the change.

Days:

| Initial situation | After YD-452 |
|-------------------|--------------|
| select ... from users | select ... from users |
| select ... from users_private | select ... from users_private |
| select ... from users | select ... from users |
| select ... from users | select ... from users_anonymized |
| select ... from users | select ... from goals |
| select ... from users | select ... from interval_activities |
| select ... from users_anonymized | select ... from activities |
| select ... from goals |  |
| select ... from interval_activities |  |
| select ... from interval_activities |  |
| select ... from activities |  |

Weeks:

| Initial situation | After YD-452 |
|-------------------|--------------|
| select ... from users | select ... from users |
| select ... from users_private | select ... from users_private |
| select ... from users | select ... from users |
| select ... from users | select ... from users_anonymized |
| select ... from users | select ... from goals |
| select ... from users | select ... from interval_activities |
| select ... from users_anonymized | select ... from activities |
| select ... from goals |  |
| select ... from interval_activities |  |
| select ... from activities |  |

core/src/main/java/nu/yona/server/subscriptions/entities/UserAnonymized.java
Bert-R: Done.

core/src/main/java/nu/yona/server/subscriptions/entities/User.java
Bert-R: Done for all occurrences.

core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
Bert-R: Done for all three methods.

core/src/main/java/nu/yona/server/subscriptions/entities/User.java
Bert-R: Despite the documentation, I can't remove all occurrences, as that causes the decryption check to fail. I've reverted this change, so all ``FetchType.LAZY`` instances are still in.

Bert-R: See [this answer](http://stackoverflow.com/a/42168093/4353482)

PR #358, opened 2017-05-23
Initial commit of k8s helm/local work for merge upstream

kti-richard:
This is the checkin of the k8s / helm / local instance work.
For this to work in full the changes to the dbinit docker container also need to be applied or it will just be liquibase db seed/updates not also ldap/quartz/category work


yonabuild:

k8s/containers/seed-update/Makefile
```diff
@@ -0,0 +1,13 @@
+SHELL=/bin/bash
+buildversion = $(shell cat version)
+
+.PHONY: all
+all: build
+
+.PHONY: increment
+increment:
+	echo $$(( $(buildversion) + 1 )) > version
```
Bert-R: This assumes that the build numbers will be maintained in the file rather than defined by Jenkins. That doesn't sound like a good idea. Jenkins provides the build number as an environment variable. Can't you use that instead of the file?

k8s/containers/seed-update/changelogs/base/base.yml
```diff
@@ -0,0 +1,13 @@
+databaseChangeLog:
```
Bert-R: Why are all Liquibase files copied? They're already in a different folder in Git. Can't they be taken from there?

k8s/containers/seed-update/Makefile
kti-richard: This was just some boiler plate I use from other project build - I am not suggesting that Jenkiins uses the make file, it was just for local dev as I had to do manual builds because Jenkins is not pushing this to docker hub.

k8s/containers/seed-update/changelogs/base/base.yml
kti-richard: That is a mistake in the commit - there should be no /changelogs - Again, I had to build these images locally for testing with the updated scripts

kti-richard: To make things simpler - I have just removed the seed-update.  The separate pull request for dbinit should have the required changes.

Bert-R:
I'm going to merge this and study further.
Thanks!

PR #361, opened 2017-05-23
YD-461 Test to generate data for query analysis

Bert-R:
The test creates:

* A user with 3 buddies
* Each of the 4 users gets 4 goals, three budget goals and a timezone goal
* Two if the timezone goals are updated 3 times, to create history items
* Over a period of the current week and the two preceding weeks, each user does 4 activities for each goal, in the early hours of the day
* The "main user" and one of his buddies is having a conversation (multiple activity comment messages) on an activity of yesterday

As the test is rather time consuming (2.5 minutes), it is skipped by default. To run it, do this:

```
gradlew :appservice:cleanIntTest :appservice:intTest -Pyona_test_gen_data=true -DintTest.single=PerformanceSetupTest
```

The user and his password are printed to stdout, like this (in the test output):

```
Test user URL: http://localhost:8082/users/26d92f7f-1a63-41af-8791-df8166f0e3ac, password: AES:128:KHoAy+XTPcKeiE/pfdDAXg==
```

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
@@ -7,6 +7,7 @@
 package nu.yona.server
 
 import groovy.json.*
+import nu.yona.server.test.AppService
 import nu.yona.server.test.Buddy
 import nu.yona.server.test.Goal
 import nu.yona.server.test.User
@@ -362,10 +363,23 @@ class ActivityCommentTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(bob)
 	}
 
-	private void assertCommentingWorks(User richard, User bob, boolean isWeek, Closure userOverviewRetriever, Closure buddyOverviewRetriever, Closure detailsRetriever)
+	public void assertCommentingWorks(User richard, User bob, boolean isWeek, Closure userOverviewRetriever, Closure buddyOverviewRetriever, Closure detailsRetriever)
 	{
-		Goal budgetGoalNewsBuddyBob = richard.buddies[0].findActiveGoal(NEWS_ACT_CAT_URL)
-		Goal budgetGoalNewsBob = bob.findActiveGoal(NEWS_ACT_CAT_URL)
+		assertCommentingWorks(appService, richard, bob, NEWS_ACT_CAT_URL, isWeek, userOverviewRetriever, buddyOverviewRetriever, detailsRetriever, { user, message -> assertMarkReadUnread(user, message)})
+	}
+
+	private getActivityDetailMessages(responseGetActivityDetails, User user, expectedData, int pageSize = 4)
+	{
+		return getActivityDetailMessages(appService, responseGetActivityDetails, user, expectedData, pageSize)
+	}
+
+	public static void assertCommentingWorks(AppService appService, User richard, User bob, def activityCategoryUrl, boolean isWeek, Closure userOverviewRetriever, Closure buddyOverviewRetriever, Closure detailsRetriever,
+			Closure assertMarkReadUnread)
+	{
+		Goal budgetGoalNewsBuddyBob = richard.buddies[0].findActiveGoal(activityCategoryUrl)
+		assert budgetGoalNewsBuddyBob
+		Goal budgetGoalNewsBob = bob.findActiveGoal(activityCategoryUrl)
+		assert budgetGoalNewsBob
 
 		def responseOverviewsBobAsBuddyAll = buddyOverviewRetriever(richard)
 		assert responseOverviewsBobAsBuddyAll.status == 200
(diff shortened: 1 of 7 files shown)


appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
```diff
@@ -362,10 +363,23 @@ class ActivityCommentTest extends AbstractAppServiceIntegrationTest
...
-		Goal budgetGoalNewsBob = bob.findActiveGoal(NEWS_ACT_CAT_URL)
+		assertCommentingWorks(appService, richard, bob, NEWS_ACT_CAT_URL, isWeek, userOverviewRetriever, buddyOverviewRetriever, detailsRetriever, { user, message -> assertMarkReadUnread(user, message)})
+	}
+
+	private getActivityDetailMessages(responseGetActivityDetails, User user, expectedData, int pageSize = 4)
+	{
+		return getActivityDetailMessages(appService, responseGetActivityDetails, user, expectedData, pageSize)
+	}
+
+	public static void assertCommentingWorks(AppService appService, User richard, User bob, def activityCategoryUrl, boolean isWeek, Closure userOverviewRetriever, Closure buddyOverviewRetriever, Closure detailsRetriever,
```
aristotelos: Pfew... So many parameters... I hope we can make the tests readable again later on.

appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
```diff
@@ -454,7 +468,7 @@ class ActivityCommentTest extends AbstractAppServiceIntegrationTest
 		assertCommentMessageDetails(activityCommentMessagesBob[2], bob, isWeek, bob.buddies[0], responseDetailsBob.responseData._links.self.href, "You're quiet!", activityCommentMessagesBob[2])
 	}
 
-	private void replyToMessage(messageToReply, User senderUser, messageToSend, boolean isWeek, responseGetActivityDetails, threadHeadMessage)
+	private static void replyToMessage(AppService appService, messageToReply, User senderUser, messageToSend, boolean isWeek, responseGetActivityDetails, threadHeadMessage)
```
aristotelos: Same here. I think it is a good idea to extract a class somewhere that holds most of the parameters.

appservice/src/intTest/groovy/nu/yona/server/PerformanceSetupTest.groovy
```diff
@@ -0,0 +1,134 @@
...
+		(1..numBuddies).each
+		{
+			User buddyUser = createBuddyUser(it)
+			appService.makeBuddies(user, buddyUser)
+			buddyUsers.add(appService.reloadUser(buddyUser))
+		}
+		return buddyUsers
+	}
+
+	User createMainUser()
```
aristotelos: It seems this method is not used. You are using `addRichard` above.

appservice/src/intTest/groovy/nu/yona/server/PerformanceSetupTest.groovy
```diff
@@ -0,0 +1,134 @@
...
+		User richard = appService.addUser(appService.&assertUserCreationResponseDetails, "Richard", "Quinn", "RQ",
+				"+$timestamp")
+		richard = appService.confirmMobileNumber(appService.&assertResponseStatusSuccess, richard)
+		setGoals(richard)
+		return richard
+	}
+
+	User createBuddyUser(int index)
+	{
+		User bob = appService.addUser(appService.&assertUserCreationResponseDetails, "Bob" + index, "Dunn" + index, "BD" + index,
```
aristotelos: Consider to rename this variable `bob` to `aBob` or `buddyUser`.

appservice/src/intTest/groovy/nu/yona/server/PerformanceSetupTest.groovy
```diff
@@ -0,0 +1,134 @@
...
+		hoursBegin.eachWithIndex
+		{ begin, idx ->
+			reportAppActivity(user, app, "$day $begin", "$day " + hoursEnd[idx])
+		}
+	}
+	void generateCommentThreadOnYesterdaysNewsActivity(User user, def buddyUsers)
+	{
+		User buddyUser = buddyUsers.find { it.nickname == user.buddies[0].nickname }
+		def yesterdayShortDay = LocalDate.now().getDayOfWeek().minus(1).getDisplayName(TextStyle.SHORT, Locale.US)
+		ActivityCommentTest.assertCommentingWorks(appService, user, buddyUser, NEWS_ACT_CAT_URL, false, {u -> appService.getDayActivityOverviews(u, ["size": 14])},
```
aristotelos: I would rather rename this method to reflect that it actually generates comments...

appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
Bert-R: Understood, also the previous comment. Let's do that when we touch this again.

appservice/src/intTest/groovy/nu/yona/server/PerformanceSetupTest.groovy
Bert-R: Correct. Removed it.

appservice/src/intTest/groovy/nu/yona/server/PerformanceSetupTest.groovy
Bert-R: Renamed into ``buddyUser``.

appservice/src/intTest/groovy/nu/yona/server/PerformanceSetupTest.groovy
Bert-R: The alternative would be to write another piece of code that just creates the comment thread, without all the asserts. That's in principle a better approach, but I didn't deem it worth the effort given the limited purpose of just creating a chunk of test data.

PR #362, opened 2017-05-23
YD-462 Fix some SonarQube reported issues

aristotelos:


core/src/main/java/nu/yona/server/exceptions/ResourceBasedException.java
```diff
@@ -31,27 +33,17 @@ void onContextStarted(ContextRefreshedEvent event)
...
-	 * @param parameters The parameters for the message
-	 */
-	protected ResourceBasedException(HttpStatus statusCode, String messageId, Object... parameters)
+	protected ResourceBasedException(HttpStatus statusCode, String messageId, Object... messageArgs)
 	{
 		super(messageId);
 
 		this.messageId = messageId;
-		this.parameters = parameters;
+		this.messageArgs = serializeMessageArgs(messageArgs);
```
Bert-R: This I don't like. The message format can in principle use format specifications that depend on the type (see [this](https://docs.oracle.com/javase/8/docs/api/java/text/MessageFormat.html)). By converting everything to string, that won't work anymore. Next, I expect the message format will show "null" when a null value is given, while the new approach would result in an NPE.

Given that this is only done to address a serialization issue that we don't have (as we don't serialize), I propose we revert this part of the change.

aristotelos: OK. I agree that this is a disadvantage. However, I think there is a good reason why exceptions should be serializable. Maybe a good solution is to format the message in the constructor and remove the messageArgs as a field?

Bert-R: That would solve it, and in our context I don't think it's an issue.

Bert-R: Actually not. Also in our context, we require late translation. We are logging all exceptions in English, even though we return them to the user in Dutch or even Chinese (later).

I'm creating PR with a fix.

PR #364, opened 2017-05-24
YD-464 Added base-64 encoded Mobile Config

Bert-R:
To work around the issue that the iOS app currently does not support binary responses (which occur when the mobile config is signed).


Bert-R:
Merging this myself, as it's urgent.

PR #366, opened 2017-05-30
YD-408 Sending system messages using batch job

aristotelos:
Currently starts the job immediately


core/src/main/java/nu/yona/server/messaging/service/MessageService.java
```diff
@@ -52,7 +51,7 @@
 	private BuddyService buddyService;
 
 	@Autowired
-	private UserAnonymizedService userAnonymizedService;
+	private BatchProxyService batchProxyService;
```
Bert-R: Let's call this from the admin service instead of the core, as this functionality is really admin specific.

aristotelos: OK, done

PR #367, opened 2017-05-31
K8s updates

kti-richard:
Various updates to the k8s material.   Should not effect anything but helm local and production.
- 'release' variable renamed to 'build'
- Mobiconfig user material templated
- Typo fix in mobiconfig
- Removed and added to gitignore the requirements.lock
- Cert material is no longer word wrapped as having to inject \n for only some uses of the cert variables was going to be cumbersome.  Cert 'wrapping' is should only be cosmetic thing, so no parsers 'should' have a problem with it.  OpenVPN doesn't care, but perhaps the iOS implementation will.


yonabuild:

Bert-R:
Looks good. I'm going to merge it.
We were considering to take the YONA_BATCH_SERVICE_SERVICE_URL along in our PR for the system messages, but we didn't feel confident enough. Next time, we'll do it and request your review.

PR #369, opened 2017-06-07
YD-468 Support different SMS sender numbers

Bert-R:
New properties are introduced:

yona.sms.defaultSenderNumber = +17184169858
yona.sms.alphaSenderId = Yona
yona.sms.alphaSenderSupportingCountryCallingCodes = +31 +49 +33

If the country calling code of the target phone number is in the alpha-sender-supporting country calling codes, then the alphaSenderId will be used as sender, otherwise the defaultSenderNumber.

### DIFF
--- a/core/src/test/java/nu/yona/server/sms/PlivoSmsServiceTest.java
+++ b/core/src/test/java/nu/yona/server/sms/PlivoSmsServiceTest.java
@@ -4,25 +4,21 @@ import static org.hamcrest.CoreMatchers.equalTo;
 import static org.junit.Assert.assertThat;
 import static org.mockito.Mockito.when;
 
-import java.util.Arrays;
-import java.util.Collection;
-
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
-import org.junit.runners.Parameterized.Parameter;
-import org.junit.runners.Parameterized.Parameters;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnit;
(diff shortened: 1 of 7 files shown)


core/src/test/java/nu/yona/server/sms/PlivoSmsServiceTest.java
```diff
@@ -0,0 +1,70 @@
...
+	{
+		SmsProperties smsProperties = new SmsProperties();
+		smsProperties.setAlphaSenderId(TEST_ALPHA_ID);
+		smsProperties.setDefaultSenderNumber(TEST_DEFAULT_NUMBER);
+		smsProperties.setAlphaSenderSupportingCountryCallingCodes(TEST_ALPHA_SUPPORTING_COUNTRY_CALLING_CODES);
+		when(mockYonaProperties.getSms()).thenReturn(smsProperties);
+	}
+
+	@Test
+	public void test()
```
aristotelos: Please use a speaking name (containing `determineSender`)

core/src/test/java/nu/yona/server/sms/PlivoSmsServiceTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.runners.Parameterized.Parameters;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+import nu.yona.server.properties.SmsProperties;
+import nu.yona.server.properties.YonaProperties;
+
+@RunWith(Parameterized.class)
```
aristotelos: This parametrizes the complete class, so I think the class should be renamed to a more specific name. Or you could consider to use [https://github.com/Pragmatists/JUnitParams]

PR #370, opened 2017-06-13
YD-469 Moved LDAP to separate repo and use AppVersion

Bert-R:
Instead of global.build

Once kubernetes/helm#2566 is addressed, we can easily build the Helm package from Jenkins.


k8s/helm/yona-server/requirements.yaml
```diff
@@ -4,5 +4,5 @@ dependencies:
     repository:  "https://kubernetes-charts.storage.googleapis.com"
   - name: ldap
     version: 0.1.0
-    repository:  "file://../ldap"
+    repository:  "https://yonadev.github.io/helm-charts"
```
kti-richard: This gave a merge conflict with my PR due to "condition: support_infrastructure.enabled" being added.  Might merge fine at master.

Bert-R: I've merged your PR into the master, synced this one into it and resolved the conflicts. I want to do two more things:
* Rename yona-server into yona
* Test it

But before that you can review the current change set already.

PR #374, opened 2017-06-20
Fix bug in checked-in certificate material in values.yaml

kti-richard:
Resolve issue with default root certificate material certs:root


kti-richard:
Also fixed the defaultSenderNumber quoting issue

Bert-R:
Looks good. I made one small change: replaced 543 with ``_ReplaceWithBuildNumber_``.

PR #375, opened 2017-06-20
YD-470 Reduce queries for GET messages

aristotelos:
Reduced from 84 (1st try)/76 (2nd try) to 31 JDBC statements

### DIFF
--- a/core/src/main/java/nu/yona/server/messaging/service/MessageDto.java
+++ b/core/src/main/java/nu/yona/server/messaging/service/MessageDto.java
@@ -188,7 +188,7 @@ public abstract class MessageDto extends PolymorphicDto
 
 				// Note: actingUser may be out of sync here, because the message action may have changed its state
 				// so retrieve afresh from buddy service
-				Optional<BuddyDto> buddy = buddyService.getBuddyOfUserByUserAnonymizedId(actingUser.getId(),
+				Optional<BuddyDto> buddy = buddyService.getBuddyOfUserByUserAnonymizedId(actingUser.getPrivateData(),
 						senderUserAnonymizedId.get());
 				if (buddy.isPresent())
 				{
--- a/core/src/main/java/nu/yona/server/messaging/service/MessageDto.java
+++ b/core/src/main/java/nu/yona/server/messaging/service/MessageDto.java
@@ -186,8 +186,6 @@ public abstract class MessageDto extends PolymorphicDto
 					return createSenderInfoForSelf(actingUser);
 				}
 
-				// Note: actingUser may be out of sync here, because the message action may have changed its state
-				// so retrieve afresh from buddy service
 				Optional<BuddyDto> buddy = buddyService.getBuddyOfUserByUserAnonymizedId(actingUser.getPrivateData(),
 						senderUserAnonymizedId.get());
 				if (buddy.isPresent())
--- a/core/src/main/java/nu/yona/server/subscriptions/service/BuddyConnectResponseMessageDto.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/BuddyConnectResponseMessageDto.java
@@ -140,19 +140,25 @@ public class BuddyConnectResponseMessageDto extends BuddyMessageLinkedUserDto
 						connectResponseMessageEntity.getRelatedUserAnonymizedId().get(),
 						connectResponseMessageEntity.getSenderNickname());
 			}
-			UserDto actingUserAfterUpdate = userService.getPrivateUser(actingUser.getId());
+			// refresh after actions
+			actingUser = userService.getPrivateUser(actingUser.getId());
 
 			connectResponseMessageEntity = updateMessageStatusAsProcessed(connectResponseMessageEntity);
 
+			logHandledAction_Process(actingUser, connectResponseMessageEntity);
+
+			return MessageActionDto
+					.createInstanceActionDone(theDtoFactory.createInstance(actingUser, connectResponseMessageEntity));
+		}
+
+		private void logHandledAction_Process(UserDto actingUser, BuddyConnectResponseMessage connectResponseMessageEntity)
+		{
 			Optional<User> senderUser = connectResponseMessageEntity.getSenderUser();
 			String mobileNumber = senderUser.map(User::getMobileNumber).orElse("already deleted");
 			String id = senderUser.map(u -> u.getId().toString()).orElse("already deleted");
 			logger.info(
 					"User with mobile number '{}' and ID '{}' processed buddy connect response from user with mobile number '{}' and ID '{}'",
 					actingUser.getMobileNumber(), actingUser.getId(), mobileNumber, id);
-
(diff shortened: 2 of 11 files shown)


core/src/main/java/nu/yona/server/messaging/service/MessageDto.java
```diff
@@ -186,9 +186,7 @@ protected final SenderInfo getSenderInfo(UserDto actingUser, Message messageEnti
 					return createSenderInfoForSelf(actingUser);
 				}
 
-				// Note: actingUser may be out of sync here, because the message action may have changed its state
```
Bert-R: Why is this comment removed?

core/src/main/java/nu/yona/server/subscriptions/service/BuddyConnectResponseMessageDto.java
```diff
@@ -137,6 +140,7 @@ MessageActionDto handleAction_Process(UserDto actingUser, BuddyConnectResponseMe
 						connectResponseMessageEntity.getRelatedUserAnonymizedId().get(),
 						connectResponseMessageEntity.getSenderNickname());
 			}
+			UserDto actingUserAfterUpdate = userService.getPrivateUser(actingUser.getId());
```
Bert-R: Why is this being fetched again? Is that related to my previous comment about the removed comment?

And why is the user fetched here already and only used 12 lines later?

core/src/main/java/nu/yona/server/subscriptions/service/BuddyDisconnectMessageDto.java
```diff
@@ -124,6 +127,7 @@ private MessageActionDto handleAction_Process(UserDto actingUser, BuddyDisconnec
 				MessageActionDto requestPayload)
 		{
 			buddyService.removeBuddyAfterBuddyRemovedConnection(actingUser.getId(), messageEntity.getSenderUserId());
+			UserDto actingUserAfterUpdate = userService.getPrivateUser(actingUser.getId());
```
Bert-R: See previous comment.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -392,9 +392,8 @@ public void setBuddyAcceptedWithSecretUserInfo(UUID buddyId, UUID userAnonymized
 		Buddy.getRepository().save(buddy);
 	}
 
-	public Set<BuddyDto> getBuddies(Set<UUID> buddyIds)
+	Set<BuddyDto> getBuddies(Set<Buddy> buddyEntities)
```
Bert-R: Please rename into ``getBuddyDtos``

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -620,9 +619,9 @@ private boolean buddyUserExists(BuddyDto buddy)
 		}
 	}
 
-	public Optional<BuddyDto> getBuddyOfUserByUserAnonymizedId(UUID forUserId, UUID userAnonymizedId)
+	public Optional<BuddyDto> getBuddyOfUserByUserAnonymizedId(UserPrivateDto forUser, UUID userAnonymizedId)
```
Bert-R: Please rename ``forUser`` into ``user``, like we do everywhere else.

core/src/main/java/nu/yona/server/subscriptions/service/UserPrivateDto.java
```diff
@@ -34,30 +34,28 @@
...
 	public UserPrivateDto(@JsonProperty("nickname") String nickname)
 	{
-		this(Optional.empty(), null, nickname, null, null, Collections.emptySet(), Collections.emptySet(), null, null,
+		this(Optional.empty(), null, nickname, null, null, Collections.emptySet(), Collections.emptySet(), null,
 				new VPNProfileDto(null));
 	}
 
 	UserPrivateDto(Optional<LocalDate> lastMonitoredActivityDate, String yonaPassword, String nickname, UUID namedMessageSourceId,
-			UUID anonymousMessageSourceId, Set<GoalDto> goals, Set<UUID> buddyIds,
-			Function<Set<UUID>, Set<BuddyDto>> buddyIdToDtoMapper, UUID userAnonymizedId, VPNProfileDto vpnProfile)
```
Bert-R: I'm trying to remember where the ``buddyIdToDtoMapper`` was for. I think it was because the mapping couldn't be made upfront, at that time. Do you have any idea?

core/src/main/java/nu/yona/server/messaging/service/MessageDto.java
aristotelos: Because it is now no longer the case, it is synced in the message action handlers before going here. I can still leave a warning if you like?

core/src/main/java/nu/yona/server/subscriptions/service/BuddyConnectResponseMessageDto.java
aristotelos: The user DTO has changed, and the response of this method should use the up-to-date user DTO. I update it here because the actions are fired here.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
aristotelos: OK

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
aristotelos: OK

core/src/main/java/nu/yona/server/subscriptions/service/BuddyConnectResponseMessageDto.java
Bert-R: Wouldn't it be fine if you would just inline ``userService.getPrivateUser(actingUser.getId())`` in the place where ``actingUserAfterUpdate`` is used on line 155 below?

core/src/main/java/nu/yona/server/messaging/service/MessageDto.java
Bert-R: No, it's fine with me. If the warning isn't applicable anymore, we shouldn't have it in the code.

core/src/main/java/nu/yona/server/subscriptions/service/UserPrivateDto.java
aristotelos: It was introduced for [YD-245](http://jira.yona.nu/browse/YD-245). See [this commit](https://github.com/yonadev/yona-server/pull/198/commits/da09c055159a728ff5b0ed4dc831581282d9eafd). I don't know the exact reason, because I was on holiday. The commit message reads: "UserDTO (with nested DTOs) now fetches the buddies when needed instead of having the controller set them explicitly".

PR #377, opened 2017-06-21
Work around go template or helm yaml poor default handling for booleans

kti-richard:
go template 'default' function seems to feel that false values = nill/empty/undef values.
As a result, any value we set to 'false' would always get the default value.

I added a 'required' function in place.  This seems to have no issue understanding false != undef, and provides an indication to developer that this is a required field, if they have somehow missed it in values.yaml.  An alternative would have been to implement some conditionals, but messy imho.

Validation/Test case for fix : 
```
# helm upgrade --dry-run --debug --install --namespace yona --set email.enabled=true yona ./yona | grep "yona.email.enabled"
yona.email.enabled=true

# helm upgrade --dry-run --debug --install --namespace yona --set email.enabled=false yona ./yona | grep "yona.email.enabled"
yona.email.enabled=false
```

### DIFF
--- a/k8s/helm/yona/Chart.yaml
+++ b/k8s/helm/yona/Chart.yaml
@@ -1,7 +1,7 @@
 name: yona
 home: https://yona.nu/
 version: 0.0.5
-appVersion: _ReplaceWithBuildNumber_
+appVersion: 552
 description: The Yona Server backs the Yona mobile app, providing user management, buddy relationship management, web traffic classification and buddy messaging
 sources:
   - https://github.com/yonadev/yona-server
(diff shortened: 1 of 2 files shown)


Bert-R:
Thanks for the quick fix!
I found an alternative that also works: putting quotes around ``false``. That has a downside:

```
2017/06/22 07:33:53 Warning: Condition path 'support_infrastructure.enabled' for chart mariadb returned non-bool value
2017/06/22 07:33:53 Warning: Condition path 'support_infrastructure.enabled' for chart ldap returned non-bool value
```

The quotes make it a string, which is what we want for the properties, but if we use the value in a condition, the above error is shown. 

Given that, I favor your approach and am going to merge it.

PR #380, opened 2017-06-27
YD-475 Small Helm improvements

Bert-R:
* Use yona-default as storage class
* Recycle the storage
* Remove unnecessary volumes from Admin service deployment
* Remove the wipe-database step from the Jenkinsfile as we start with a blank slate every time

### DIFF
--- a/k8s/helm/yona/values.yaml
+++ b/k8s/helm/yona/values.yaml
@@ -170,6 +170,7 @@ mariadb:
   mariadbPassword: develop
   mariadbDatabase: yona
   persistence:
+    storageClass: yona-default
     size: 2Gi
   user: develop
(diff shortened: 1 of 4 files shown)


k8s/helm/yona/values.yaml
```diff
@@ -170,6 +170,7 @@ mariadb:
   mariadbPassword: develop
   mariadbDatabase: yona
   persistence:
+    storageClass: yona-default
```
Bert-R: On the LDAP chart, we need to also specify this storage class.

seedickcode:
Reviewed, but not tested.   If it works, then I guess it is defaulting to the manual storage provider, so that's a win.

PR #382, opened 2017-06-30
YD-473 Upgrade to Spring Boot 1.4.7

aristotelos:
This fixes the performance issue
[https://github.com/spring-projects/spring-hateoas/issues/525], which
was solved in Spring Framework 4.3.5.
Edited the single broken test. It seems PagedResourceAssembler adds a
different self link in case no paging was passed, but I don't think that
is harmful.


Bert-R:
That's great news!

PR #383, opened 2017-07-04
YD-447 Removed app-provided pwd option

Bert-R:


appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -319,33 +318,14 @@ static ControllerLinkBuilder getConfirmMobileNumberLinkBuilder(UUID userId)
...
-	{
-		if (password.isPresent())
-		{
-			if (yonaProperties.getSecurity().isAppProvidedPasswordEnabled())
-			{
-				logger.warn("Creating user with app-provided password");
-			}
-			else
-			{
-				throw InvalidDataException.appProvidedPasswordNotSupported();
```
aristotelos: Has this exception been removed? Including translation?

core/src/main/resources/application.properties
```diff
@@ -57,7 +57,6 @@ yona.security.maxUpdateUserAttemptsPerTimeWindow = 2
 yona.security.corsAllowed = true
 yona.security.sslRootCertFile = resources/rootcert.cer
 yona.security.ovpnProfileFile = resources/profile.ovpn
-yona.security.appProvidedPasswordEnabled = true
```
aristotelos: Was this already tried with `false` on production?

Bert-R: Yes, the production server has this as false. That's why I know it can be removed.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
Bert-R: No, it hasn't, and it shouldn't, as it's still in use at line 229 (``updateUserCreatedOnBuddyRequest``). Honestly, I forgot to remove it, but when I tried to remove it, in follow up to your comment, I see it's still in use.

PR #384, opened 2017-07-06
YD-340 Only include own goals when shared with a buddy

Bert-R:
Along with this:
* Fixed issue in AnalysisEngineService where dayActivity was transient
* Added extra integration test
* Extended the Groovy classes with @ToString

### DIFF
--- a/analysisservice/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
+++ b/analysisservice/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
@@ -358,6 +358,7 @@ public class AnalysisEngineService
 	{
 		DayActivity dayActivity = DayActivity.createInstance(userAnonymizedEntity, matchingGoal, payload.startTime.getZone(),
 				TimeUtil.getStartOfDay(payload.userAnonymized.getTimeZone(), payload.startTime).toLocalDate());
+		dayActivityRepository.save(dayActivity);
 
 		ZonedDateTime startOfWeek = TimeUtil.getStartOfWeek(payload.userAnonymized.getTimeZone(), payload.startTime);
 		WeekActivity weekActivity = weekActivityRepository.findOne(payload.userAnonymized.getId(), matchingGoal.getId(),
--- a/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
@@ -1223,7 +1223,7 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(bob)
 	}
 
-	def 'Richard and Bob have different goals'()
+	def 'Richard\'s goals are only included when a buddy has that goal too'()
 	{
 		given:
 		def richardAndBob = addRichardAndBobAsBuddies()
--- a/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
@@ -15,7 +15,6 @@ import java.time.temporal.IsoFields
 
 import groovy.json.*
 import nu.yona.server.test.AppActivity
-import nu.yona.server.test.Buddy
 import nu.yona.server.test.BudgetGoal
 import nu.yona.server.test.Goal
 import nu.yona.server.test.TimeZoneGoal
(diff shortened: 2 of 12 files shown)


core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
```diff
@@ -169,6 +169,12 @@ private void createAndSaveInactivityDays(UserAnonymizedDto userAnonymized, Set<G
 		weekActivity.createRequiredInactivityDays(userAnonymized, goals, LevelOfDetail.WEEK_OVERVIEW, missingInactivities);
 	}
 
+	private long getTotalPageableItems(Set<BuddyDto> buddies, ChronoUnit timeUnit)
```
aristotelos: Please pick a more speaking name, e.g. `getTotalPageableItemsBuddiesCombinedTimeline`.

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
```diff
@@ -215,27 +221,26 @@ private long getTotalPageableItems(UserAnonymizedDto userAnonymized, ChronoUnit
...
-		if (!userAnonymized.hasAnyBuddies())
-		{
-			return new PageImpl<>(Collections.emptyList(), pageable, 0);
-		}
-
 		Interval interval = getInterval(getCurrentDayDate(userAnonymized), pageable, ChronoUnit.DAYS);
-
 		Set<BuddyDto> buddies = buddyService.getBuddiesOfUserThatAcceptedSending(userId);
 		Set<UUID> userAnonymizedIds = buddies.stream().map(this::getBuddyUserAnonymizedId).collect(Collectors.toSet());
+		Set<UUID> activityCategoryIds = buddies.stream().map(BuddyDto::getGoals).flatMap(Set::stream)
```
aristotelos: It would be good to explain here, e.g.:
`// Friend results are all included. Own results are only included if some friend has a goal in the same category.`

aristotelos:
Shouldn't we rename `UserWithBuddies` to `BuddiesCombined` or so? The name `UserWithBuddies` does not really say what it is.

analysisservice/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -358,6 +358,7 @@ private DayActivity createNewDayActivity(UserAnonymized userAnonymizedEntity, Ac
 	{
 		DayActivity dayActivity = DayActivity.createInstance(userAnonymizedEntity, matchingGoal, payload.startTime.getZone(),
 				TimeUtil.getStartOfDay(payload.userAnonymized.getTimeZone(), payload.startTime).toLocalDate());
+		dayActivityRepository.save(dayActivity);
```
Bert-R: Indeed, and most of the times it works, but sometimes not. You had that error last week and I had it this week. Explicitly saving it is not an issue and we do this often, but it shouldn't be necessary.

Sometime back, I tried removing all saves except for the root of the cascading tree (see [YD-457](http://jira.yona.nu/browse/YD-457), but that failed because at times, JPA/Hibernate seems to miss a cascading persist.

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -1224,6 +1223,117 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(bob)
 	}
 
+	def 'Richard and Bob have different goals'()
```
Bert-R: Agreed. At the time I started the test, I didn't have a clear idea what the test would do.

I'll rename it in "Richard's goals are only included when a buddy has that goal too"

core/src/main/java/nu/yona/server/analysis/service/ActivityService.java
Bert-R: Mmm. It's just an overload of the other method. One takes a ``userAnonymized`` and returns the number of pageable items for it, the other takes a set of buddies and does the same. I like to keep these method names the same.

Bert-R:
Pushed an update to address your comments

Bert-R:
> Shouldn't we rename ``UserWithBuddies`` to ``BuddiesCombined`` or so? The name ``UserWithBuddies`` does not really say what it is.

To me, it's more clear than ``BuddiesCombined``, as it exactly describes what it is: the user with their buddies.

aristotelos:
Before this change, it indeed was the user with buddies. Now, it is the buddies, with the user if and only if one of the buddies has a goal in the same category as the user. The UI design also calls it "Friends timeline" so it is not about the user at all...

Bert-R:
OK, I see what you mean. Still, I'm not in favor of renaming it. Externally, it's exposed as ``/users/{userID}/activity/withBuddies/days/``, which reflects "user ``{userID}`` with their buddies". If we would rename it internally, it would become a little more precise there, but it would also become inconsistent.

aristotelos:
I think it would be OK to leave the name like it is now, but then we may like to place a comment in the code somewhere to explain how it works (because the name 'user with buddies' is misleading).

Bert-R:
Added the requested comment

PR #387, opened 2017-08-01
YD-477 Renamed tests with less speaking names

aristotelos:
Postfixes like "rightResult" removed


Bert-R:
Looks good, going to merge it.

PR #390, opened 2017-08-22
YD-203 Added activity overview self-links

Bert-R:
For:
* User activity overviews (both day and week), with and without buddies
* Buddy activity overviews (both day and week)

Extended the integration tests accordingly

### DIFF
--- a/appservice/src/main/java/nu/yona/server/analysis/rest/ActivityControllerBase.java
+++ b/appservice/src/main/java/nu/yona/server/analysis/rest/ActivityControllerBase.java
@@ -81,10 +81,8 @@ abstract class ActivityControllerBase
 	protected static final int WEEKS_DEFAULT_PAGE_SIZE = 2;
 	protected static final int DAYS_DEFAULT_PAGE_SIZE = 3;
 	protected static final int MESSAGES_DEFAULT_PAGE_SIZE = 4;
-	protected static final String PREV_REL = "prev"; // IANA reserved, so will
-														// not be prefixed
-	protected static final String NEXT_REL = "next"; // IANA reserved, so will
-														// not be prefixed
+	protected static final String PREV_REL = "prev"; // IANA reserved, so will not be prefixed
+	protected static final String NEXT_REL = "next"; // IANA reserved, so will not be prefixed
 
 	protected HttpEntity<PagedResources<WeekActivityOverviewResource>> getWeekActivityOverviews(Optional<String> password,
 			UUID userId, PagedResourcesAssembler<WeekActivityOverviewDto> pagedResourcesAssembler,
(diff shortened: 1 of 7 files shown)


appservice/src/main/java/nu/yona/server/analysis/rest/ActivityControllerBase.java
```diff
@@ -79,12 +81,14 @@
 	protected static final int WEEKS_DEFAULT_PAGE_SIZE = 2;
 	protected static final int DAYS_DEFAULT_PAGE_SIZE = 3;
 	protected static final int MESSAGES_DEFAULT_PAGE_SIZE = 4;
-	protected static final String PREV_REL = "prev"; // IANA reserved, so will not be prefixed
```
aristotelos: There are many whitespace changes in this PR. Any idea why?

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -603,6 +648,21 @@ paths:
           description: Activity of the buddy.
           schema:
             $ref: '#/definitions/PagedDayActivityOverviews'
+  '/users/{userID}/buddies/{buddyID}/activity/days/{date}':
+    get:
+      summary: Get a day activity overview
+      description: Get network and app activity of the specified day
+      parameters:
+        - $ref: "#/parameters/yonaPassword"
```
aristotelos: `buddyID` is missing

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -819,6 +879,21 @@ paths:
           description: Activity of the buddy.
           schema:
             $ref: '#/definitions/PagedWeekActivityOverviews'
+  '/users/{userID}/buddies/{buddyID}/activity/weeks/{week}':
+    get:
+      summary: Get a week activity overview
+      description: Get network and app activity of the specified week
+      parameters:
+        - $ref: "#/parameters/yonaPassword"
```
aristotelos: `buddyID` is missing

appservice/src/main/resources/static/swagger/swagger-spec.yaml
Bert-R: Sharp! I'll add it.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
Bert-R: This one too.

appservice/src/main/java/nu/yona/server/analysis/rest/ActivityControllerBase.java
Bert-R: I don't think there are many. This comment was wrapped in an earlier formatting action (with a different formatter) and not joined again when reformatted with the Yona formatter. Corrected that and also the next one.

Besides this, I see a line wrapping change, but that was originally too long and now corrected.

Bert-R:
Pushed an update.

PR #395, opened 2017-09-07
Init and Logrotate material for handdansguardian parser

kti-richard:
Init script, boot hook script, and logrotate configuration
to support the Dansguardian category output parser.

See:
http://wiki.yona.nu/display/DEV/Perl+script+to+pass+events+from+SmoothWall+to+the+Yona+server

For details on how to deploy

### DIFF
--- a/scripts/smoothwall/HandleDansGuardianLog.pl
+++ b/scripts/smoothwall/HandleDansGuardianLog.pl
@@ -140,6 +140,21 @@ sub handle_records_from_file {
 	close INPUT_FILE;
 }
 
+BEGIN {
+	#Fork
+	my $pidFile = '/var/run/HandleDansGuardianLog.pid';
+	my $pid = fork;
+	if ($pid) # parent: save PID
+	{
+		open PIDFILE, ">$pidFile" or die "can't open $pidFile: $!\n";
+	    print PIDFILE $pid;
+	    close PIDFILE;
+	    exit 0;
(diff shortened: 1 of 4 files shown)


kti-richard:
Note that the changes to the perl script itself might be breaking changes for the way it is currently run on older infrastructure.  Since we manually deploy this component though, it should not be a problem.

scripts/smoothwall/HandleDansGuardianLog.pl
```diff
@@ -140,6 +140,21 @@ sub handle_records_from_file {
 	close INPUT_FILE;
 }
 
+BEGIN {
+	#Fork
```
Bert-R: Can you add an explanatory comment why the script needs to fork?

scripts/smoothwall/HandleDansGuardianLog.pl
```diff
@@ -140,6 +140,21 @@ sub handle_records_from_file {
...
+	my $pidFile = '/var/run/HandleDansGuardianLog.pid';
+	my $pid = fork;
+	if ($pid) # parent: save PID
+	{
+		open PIDFILE, ">$pidFile" or die "can't open $pidFile: $!\n";
```
Bert-R: When adding the comment, can you also correct the indentation here?

Bert-R:
Looks good.

PR #396, opened 2017-09-08
YD-145 Cache BuddyAnonymized as part of UserAnonymized

Bert-R:
The UserAnonymizedDto was already cached inside UserAnonymizedService. Now this DTO holds a set of BuddyAnonymizedDto. Updates to the buddy status now need to be saved as part of the UserAnonymized, so the cache is updated properly.

While making this enhancement, it was noticed that the UserAnonymizedService had an update method that didn't really update the cache. This was corrected and signature of the other update method was simplified.

Next, the update lead to multiple "object references an unsaved transient instance - save the transient instance before flushing" errors. Save calls were added, including a new repository (ActivityRepository).

RemoveUserTest didn't assert the posts to the analysis engine were successful. This was corrected and might need to be done in other tests too.


core/src/main/java/nu/yona/server/subscriptions/service/UserAnonymizedDto.java
```diff
@@ -31,25 +30,26 @@
...
 	private final Set<GoalDto> goals;
 	private final MessageDestinationDto anonymousMessageDestination;
-	private final Set<UUID> buddyAnonymizedIds;
+	private final Set<BuddyAnonymizedDto> buddiesAnonymized;
 
 	private static final ZoneId DEFAULT_TIME_ZONE = ZoneId.of("Europe/Amsterdam");
 
 	public UserAnonymizedDto(UUID id, Optional<LocalDate> lastMonitoredActivityDate, Set<GoalDto> goals,
-			MessageDestinationDto anonymousMessageDestination, Set<UUID> buddyAnonymizedIds)
+			MessageDestinationDto anonymousMessageDestination, Set<BuddyAnonymizedDto> buddyAnonymized)
```
aristotelos: `buddiesAnonymized`

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -556,7 +556,7 @@ public void deleteUser(UUID id, Optional<String> message)
 		MessageSource anonymousMessageSource = messageSourceRepository.findOne(userEntity.getAnonymousMessageSourceId());
 		UserAnonymized userAnonymizedEntity = userAnonymizedService.getUserAnonymizedEntity(userAnonymizedId);
 		userAnonymizedEntity.clearAnonymousDestination();
-		userAnonymizedEntity = userAnonymizedService.updateUserAnonymized(userAnonymizedEntity);
+		userAnonymizedService.updateUserAnonymized(userAnonymizedEntity);
```
aristotelos: The `userAnonymizedEntity` is later on in this method saved again. Should this method not return the updated entity to be used in the next save?

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -490,10 +490,10 @@ void sendBuddyConnectResponseMessage(UUID senderUserId, UUID senderUserAnonymize
...
 	private void disconnectBuddyIfConnected(UserAnonymizedDto buddyUserAnonymized, UUID userAnonymizedId)
 	{
-		Optional<BuddyAnonymized> buddyAnonymized = buddyUserAnonymized.getBuddyAnonymized(userAnonymizedId);
-		buddyAnonymized.ifPresent(ba -> {
-			ba.setDisconnected();
-			BuddyAnonymized.getRepository().save(ba);
+		Optional<BuddyAnonymizedDto> buddyAnonymized = buddyUserAnonymized.getBuddyAnonymized(userAnonymizedId);
+		buddyAnonymized.map(ba -> buddyAnonymizedRepository.findOne(ba.getId())).ifPresent(bae -> {
+			bae.setDisconnected();
+			userAnonymizedService.updateUserAnonymized(buddyUserAnonymized.getId());
```
aristotelos: A disadvantage of the caching introduced in this PR is that we have to be very aware that any change to `BuddyAnonymized` has to result in a cache update. Shouldn't this be reflected in a comment on `BuddyAnonymized`?

core/src/main/java/nu/yona/server/subscriptions/service/UserAnonymizedDto.java
Bert-R: Done

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
Bert-R: It doesn't matter, as the entity itself isn't updated. If it would matter, we would need to do a ``getOne`` to reload it from Hibernate. We can't return the entity from the ``UserAnonymizedService``, as that returns cacheable DTOs only

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
Bert-R: There is nothing wrong with that, but the question is whether it'll be read when the knowledge is required. I'll add a class comment explaining this.

Bert-R:
Pushed an update

PR #399, opened 2017-09-13
YD-294 Interval end is now exclusive

Bert-R:
Intervals normally include one end only (so end - start == duration). Updated our code to align with that.

### DIFF
--- a/core/src/main/java/nu/yona/server/analysis/entities/DayActivityRepository.java
+++ b/core/src/main/java/nu/yona/server/analysis/entities/DayActivityRepository.java
@@ -32,7 +32,7 @@ public interface DayActivityRepository extends CrudRepository<DayActivity, Long>
 			@Param("goalId") UUID goalId);
 
 	@Query("select a from DayActivity a where a.userAnonymized.id = :userAnonymizedId and a.goal.id in :goalIds and a.startDate >= :dateFrom and a.startDate < :dateUntil order by a.startDate desc")
-	List<DayActivity> findAllActivitiesForUserInInterval(@Param("userAnonymizedId") UUID userAnonymizedId,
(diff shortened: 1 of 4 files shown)


core/src/main/java/nu/yona/server/analysis/entities/DayActivityRepository.java
```diff
@@ -31,13 +31,8 @@
...
 			@Param("goalId") UUID goalId);
 
-	@Query("select a from DayActivity a where a.userAnonymized.id = :userAnonymizedId and a.goal.id in :relevantGoalIds and a.startDate >= :dateFrom and a.startDate <= :dateUntil order by a.startDate desc")
-	List<DayActivity> findAllActivitiesForUserInIntervalEndIncluded(@Param("userAnonymizedId") UUID userAnonymizedId,
-			@Param("relevantGoalIds") Set<UUID> relevantGoalIds, @Param("dateFrom") LocalDate dateFrom,
-			@Param("dateUntil") LocalDate dateUntil);
-
 	@Query("select a from DayActivity a where a.userAnonymized.id = :userAnonymizedId and a.goal.id in :goalIds and a.startDate >= :dateFrom and a.startDate < :dateUntil order by a.startDate desc")
-	List<DayActivity> findActivitiesForUserAndGoalsInIntervalEndExcluded(@Param("userAnonymizedId") UUID userAnonymizedId,
+	List<DayActivity> findAllActivitiesForUserInInterval(@Param("userAnonymizedId") UUID userAnonymizedId,
```
aristotelos: Why not just `findAll` as in `WeekActivityRepository`?

Bert-R: Because I didn't think of that one  :)
I'll do it

Bert-R:
Pushed an update

PR #400, opened 2017-09-19
YD-100 Remove expired new device requests

Bert-R:
Along with this, renamed ActivityAggregationBatchJobResultDto into BatchJobResultDto, as the type is generic, fitting any batch job.

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/entities/NewDeviceRequestRepository.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/entities/NewDeviceRequestRepository.java
@@ -1,17 +1,20 @@
 /*******************************************************************************
- * Copyright (c) 2016 Stichting Yona Foundation
- *
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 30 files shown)


core/src/main/java/nu/yona/server/subscriptions/entities/NewDeviceRequestRepository.java
```diff
@@ -1,17 +1,20 @@
 /*******************************************************************************
- * Copyright (c) 2016 Stichting Yona Foundation
- *
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ * Copyright (c) 2016 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License, v.
```
aristotelos: `, 2017`

Bert-R: I've updated all copyrights.

Bert-R:
Pushed an update

PR #403, opened 2017-09-26
YD-486 Fixed cleanup of expired new device requests

Bert-R:


core/src/main/java/nu/yona/server/subscriptions/entities/UserRepository.java
```diff
@@ -16,4 +22,11 @@
 	int countByAppLastOpenedDateBetween(LocalDate startDate, LocalDate endDate);
 
 	int countByAppLastOpenedDateIsNull();
+
+	@Query("select a from WeekActivity a where a.userAnonymized.id = :userAnonymizedId and a.startDate >= :dateFrom and a.startDate < :dateUntil order by a.startDate desc")
+	Set<WeekActivity> findAll(@Param("userAnonymizedId") UUID userAnonymizedId, @Param("dateFrom") LocalDate dateFrom,
```
aristotelos: This does not seem to belong in a `UserRepository`? Copy-paste error :)?

Bert-R: I copied it as an example and forgot to remove it. Removed it now.

PR #408, opened 2017-10-18
YD-492 Guarded DB stats through test

Bert-R:
Created a test that guards the database access statistics of the analysis engine service. This service is very performance critical, so it is important that the database access is very limited.

The HibernateStatisticsController is now moved to Core, so each Yona service has the URL path to manage the statistics.


core/src/testUtils/groovy/nu/yona/server/test/AnalysisService.groovy
```diff
@@ -40,8 +39,38 @@ class AnalysisService extends Service
 					"url":"$url"
 				}""")
 	}
+
 	def postToAnalysisEngine(String vpnLoginId, jsonString)
 	{
 		yonaServer.postJson(USER_ANONYMIZED_PATH + vpnLoginId + "/networkActivity/", jsonString)
 	}
+	void setEnableStatistics(def enable)
```
aristotelos: This part is duplicate code with `AppService`. Can't we merge this to `YonaServer`?

Bert-R: Makes sense. Moved it.

Bert-R:
Pushed an update

Bert-R:
Don't merge this yet. Something is wrong in my last change set.

aristotelos:
I meant some more methods, `resetStatistics`, `clearCaches`, `getStatistics` as well. And do we need a wrapper method `setEnableStatistics` around `yonaServer.enableStatistics`.

Bert-R:
Oops, I overlooked these.

Bert-R:
Pushed an update.

PR #409, opened 2017-10-20
YD-488 Appropriate error code on invalid number

Bert-R:
When the response from Plivo contains "is not a valid phone number", we will now return an appropriate error message to the user instead of a technical one.


aristotelos:
Can't we improve our validation of the number when it is submitted by the user? That would solve the root cause of the problem.

Bert-R:
I don't think we can. For Dutch numbers, we could, but not internationally.

aristotelos:
I see, our regex will allow numbers that do not exist (currently).

Bert-R:
> I see, our regex will allow numbers that do not exist (currently).

I've spoken to Plivo today. They'll send me information about number checking, which I'll share once I have it.

PR #413, opened 2017-10-27
YD-494 Extended the dashboard with user counts

Bert-R:
Added the following structure:
* Users that confirmed their mobile number: 1
* Users that did not confirm their mobile number: 8
** Users invited on buddy request: 6
** Other users: 2


aristotelos:
Looks great, will merge when the build is done

PR #418, opened 2017-11-21
YD-495 add device notion

Bert-R:
Every user now has a device. If the device name and type are not provided explicitly, a default device is registered.


yonabuild:

yonabuild:
Build started sha1 is merged.

yonabuild:
Build finished.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -112,6 +114,122 @@ class CreateUserOnBuddyRequestTest extends AbstractAppServiceIntegrationTest
...
+	{
+		given:
+		def richard = addRichard()
+		def mobileNumberBob = makeMobileNumber(timestamp)
+		def inviteUrl = buildInviteUrl(sendBuddyRequestForBob(richard, mobileNumberBob))
+		def bob = appService.getUser(appService.&assertUserGetResponseDetailsWithPrivateDataCreatedOnBuddyRequest, inviteUrl, true, null)
+
+		when:
+		def newNickname = "Bobby"
+		def updatedBobJson = bob.convertToJson()
```
aristotelos: Maybe a `def bobToBeUpdated = bob.clone()` would be easier to read:
```
def bobToBeUpdated = bob.clone()
bobToBeUpdated.nickname = newNickname
bobToBeUpdated.deviceName = "My S8"
...
```

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -112,6 +114,122 @@ class CreateUserOnBuddyRequestTest extends AbstractAppServiceIntegrationTest
...
+		updatedBobJson.nickname = newNickname
+		User bobToBeUpdated = new User(updatedBobJson)
+		bobToBeUpdated.deviceName = "My S8"
+		bobToBeUpdated.deviceOperatingSystem = "ANDROID"
+		User updatedBob = appService.updateUserCreatedOnBuddyRequest(appService.&assertUserUpdateResponseDetails, bobToBeUpdated, inviteUrl)
+
+		then:
+		updatedBob.devices.size() == 0 // Mobile number not confirmed yet
```
aristotelos: Wouldn't it be more logical to return the devices here, even if the mobile number is not confirmed yet? Because the user is also allowed to save them.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -112,6 +114,122 @@ class CreateUserOnBuddyRequestTest extends AbstractAppServiceIntegrationTest
...
+		bobWithConfirmedNumber.devices.size() == 1
+		bobWithConfirmedNumber.devices[0].name == "My S8"
+		bobWithConfirmedNumber.devices[0].operatingSystem == "ANDROID"
+		
+		cleanup:
+		appService.deleteUser(richard)
+		appService.deleteUser(updatedBob)
+	}
+
+	def 'Bob adjusts data, including his iOS device'()
```
aristotelos: I don't think this test makes sense, because it is exactly the same as the previous one, except for some data.

appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
```diff
@@ -77,11 +78,87 @@ class UserTest extends AbstractAppServiceIntegrationTest
...
+	def 'Create John Doe with an Android device'()
+	{
+		given:
+		def ts = timestamp
+
+		when:
+		def johnAsCreated = createJohnDoe(ts, "My S8", "ANDROID")
+
+		then:
+		def johnAfterNumberConfirmation = appService.confirmMobileNumber(
```
aristotelos: The variable `johnAfterNumberConfirmation` is not used

appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
```diff
@@ -77,11 +78,87 @@ class UserTest extends AbstractAppServiceIntegrationTest
...
+					assert it.responseData._embedded."yona:devices"._embedded."yona:devices".size == 1
+					assert it.responseData._embedded."yona:devices"._embedded."yona:devices"[0].name == "My S8"
+					assert it.responseData._embedded."yona:devices"._embedded."yona:devices"[0].operatingSystem == "ANDROID"
+				}, johnAsCreated)
+
+		cleanup:
+		appService.deleteUser(johnAsCreated)
+	}
+
+	def 'Create John Doe with an iOS device'()
```
aristotelos: This test and the following 3 are specific for the devices feature. Wouldn't it be more logical to put them in a separate test class? The `UserTest` class is already big.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
```diff
@@ -112,6 +114,122 @@ class CreateUserOnBuddyRequestTest extends AbstractAppServiceIntegrationTest
...
+		bobWithConfirmedNumber.devices.size() == 1
+		bobWithConfirmedNumber.devices[0].name == "My iPhone X"
+		bobWithConfirmedNumber.devices[0].operatingSystem == "IOS"
+		
+		cleanup:
+		appService.deleteUser(richard)
+		appService.deleteUser(updatedBob)
+	}
+
+	def 'Try: Bob adjusts data, including his device with an unsupported operating system'()
```
aristotelos: This and the next test I would leave out, as they are already asserted in `UserTest`. If device validation works in one scenario, I think we can trust it works in other scenarios too.

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
```diff
@@ -295,14 +313,75 @@ class AppService extends Service
...
+	{
+		assert dateTimeString ==~ /[0-9]{4}-[0-9]{2}-[0-9]{2}/
+	}
+
+	static void assertResponseStatusCreated(def response)
+	{
+		assertResponseStatus(response, 201)
+	}
+
+	static void assertResponseStatus(def response, int status)
```
aristotelos: Personally, I am not in favor of these methods. I prefer `response.status == 200` over `appService.assertResponseStatusSuccess(response)` and prefer `response.status == 301` over `appService.assertResponseStatus(response, 301)`.

core/src/test/java/nu/yona/server/subscriptions/service/UserServiceTest.java
```diff
@@ -133,20 +127,10 @@ public void postOpenAppEvent_appLastOpenedDateOnSameDay_notUpdated()
 		verify(mockUserRepository, times(0)).save(any(User.class));
 	}
 
-	@Test(expected = InvalidDataException.class)
```
aristotelos: Why did you remove this test?

appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
```diff
@@ -77,11 +78,87 @@ class UserTest extends AbstractAppServiceIntegrationTest
...
+				{
+					AppService.assertResponseStatus(it, 400)
+					assert it.responseData.code == "error.device.unknown.operating.system"
+				}, "John", "Doe", "JD",
+				makeMobileNumber(ts), "First device", "UNKNOWN")
+
+		then:
+		assert johnAsCreated == null // Creation failed
+	}
+
```
aristotelos: Some other test suggestions: removing a device, adding a second device.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
Bert-R: Not really. They can create the user entity with a device but can't do anything with devices. We are till now leaving out everything till the number is confirmed. I don't think there would be any benefit if we would make an exception for the device.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
Bert-R: I wanted to verify that both supported operating systems indeed work.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
Bert-R: OK, will do.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
Bert-R: With the argument that we already test the types of devices in ``UserTest`` (see your next comment), I will leave this test out too.

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
Bert-R: That won't work as the nickname is a final field. I'd rather leave it that way. I favor immutability over the simplicity of cloning.

appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
Bert-R: Will change.

appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
Bert-R: True. I've put them here because this test has code to create an arbitrary user. It's easier to fit these tests in here then in the more "rigid" Richard and Bob structure. Once we implement the device maintenance and device change buddy messaging, we need a dedicated device test. I was thinking that we should move these tests there at that time.

appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
Bert-R: Yes, but that's part of another ticket: [YD-497](http://jira.yona.nu/browse/YD-497)

core/src/test/java/nu/yona/server/subscriptions/service/UserServiceTest.java
Bert-R: Because it's now impossible to create such invalid data: the ``PostPutUser`` does not contain goals anymore.

yonabuild:

yonabuild:
Build started sha1 is merged.

yonabuild:
Build finished.

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
Bert-R: I understand, but see what these assert methods do: they show the error response of the server in the log message. That makes analysis of a failing test a lot simpler:

A regular ``assert response.status == 200`` gives this:
```
Condition not satisfied:

it.status == 200
|  |      |
|  400    false
groovyx.net.http.HttpResponseDecorator@5137927

	at nu.yona.server.UserTest.Try to create John Doe with an unsupported device operating system_closure1(UserTest.groovy:134)
	at nu.yona.server.test.AppService.addUser(AppService.groovy:68)
	at nu.yona.server.UserTest.Try to create John Doe with an unsupported device operating system(UserTest.groovy:137)
```

With ``AppService.assertResponseStatus(response, 200)``, you get this:
```
java.lang.AssertionError: Invalid status: 400 (expecting 200). Response: [message:Operating system 'RASPBIAN' is not supported., code:error.device.unknown.operating.system]. Expression: (response.status == status). Values: status = 200
	at nu.yona.server.test.AppService.assertResponseStatus(AppService.groovy:379)
	at nu.yona.server.UserTest.Try to create John Doe with an unsupported device operating system_closure1(UserTest.groovy:134)
	at nu.yona.server.test.AppService.addUser(AppService.groovy:68)
	at nu.yona.server.UserTest.Try to create John Doe with an unsupported device operating system(UserTest.groovy:137)
```

This makes it a lot quicker to find out what's going wrong.

Bert-R:
Pushed an update for the integration tests.

yonabuild:

yonabuild:
Build started sha1 is merged.

yonabuild:
Build finished.

yonabuild:

yonabuild:
Build started sha1 is merged.

yonabuild:
Build finished.

yonabuild:

yonabuild:
Build started sha1 is merged.

yonabuild:
Build finished.

Bert-R:
This needs to be developed a little further before it can be merged: A buddy request/response should exchange the devices between the buddies.

That will make the upgrade step simpler: all users that have a device already (i.e. those that signed up after the current PR was merged) have already shared their device with their buddies. So the only thing we need to do for migration is add a device to users that don't have one yet, and communicate that to their buddies.

yonabuild:

yonabuild:
Build started sha1 is merged.

Bert-R:
Pushed an update that communicates the initial set of devices. It can now be merged once the review is OK.

appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
Bert-R: It was actually pretty simple, so I've moved them to DeviceTest.groovy

yonabuild:
Build finished.

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
aristotelos: OK, that helps indeed. Have you replaced all response status assertions with those variants? Have you also created a `assertResponseStatusOK` that asserts on 200 instead of 2xx? I think that more precise asserts are preferable.

core/src/test/java/nu/yona/server/subscriptions/service/UserServiceTest.java
aristotelos: OK

appservice/src/intTest/groovy/nu/yona/server/CreateUserOnBuddyRequestTest.groovy
aristotelos: OK

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -501,15 +503,35 @@ public PostPutUserDto(@JsonProperty("firstName") String firstName, @JsonProperty
 		}
 	}
 
+	enum Purpose
```
aristotelos: I think the purpose of this enum is to get an overview of the possible states a user resource can have. So I think `PresentationState` is maybe a better name.

core/src/main/resources/messages/messages_nl.properties
```diff
@@ -101,6 +101,8 @@ error.no.device.request.present=Er bestaat geen ''new device request'' voor de g
 error.device.request.expired=Het ''new device request'' voor gebruiker met ID ''{0}'' is verlopen
 error.device.request.invalid.password=Het tijdelijke wachtwoord voor het ''new device request'' is ongeldig. Is het verzoek misschien vervangen door een ander?
 error.device.unknown.operating.system=Besturingssysteem ''{0}'' wordt niet ondersteund.
+error.device.invalid.device.name=De apparaatnaam ``{0}`` is niet toegestaan. Een geldige apparaatnaam is niet langer dan {1} tekens en bevat geen ''{2}''.
```
aristotelos: The quotes are wrong (`` instead of '').

core/src/main/resources/messages/messages.properties
```diff
@@ -101,6 +101,7 @@ error.no.device.request.present=No new device request present for mobile number
 error.device.request.expired=New device request expired for mobile number ''{0}''
 error.device.request.invalid.password=Invalid temporary password to use new device request for mobile number ''{0}''. Maybe the request was replaced with a different secret?
 error.device.unknown.operating.system=Operating system ''{0}'' is not supported.
+error.device.invalid.device.name=Device name ``{0}`` is invalid. A valid device name is not longer than {1} characters and does not contain a ''{2}''.
```
aristotelos: The quotes are wrong here too.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -203,24 +208,47 @@ private void assertValidBuddy(UserDto requestingUser, BuddyDto buddy)
...
+	{
+		List<String> deviceNames = connectMessageEntity.getDeviceNames();
+		List<UUID> deviceAnonymizedIds = connectMessageEntity.getDeviceAnonymizedIds();
+		List<Boolean> deviceVpnConnectionStatuses = connectMessageEntity.getDeviceVpnConnectionStatuses();
+		int numDevices = deviceNames.size();
+		assert deviceAnonymizedIds.size() == numDevices;
+		assert deviceVpnConnectionStatuses.size() == numDevices;
+
+		Set<BuddyDevice> devices = new HashSet<>();
+		for (int i = 0; (i < numDevices); i++)
```
aristotelos: Shouldn't we use streams instead of for loops?

core/src/main/java/nu/yona/server/subscriptions/service/BuddyConnectResponseMessageDto.java
```diff
@@ -136,10 +136,7 @@ MessageActionDto handleAction_Process(UserDto actingUser, BuddyConnectResponseMe
 			}
 			else
 			{
-				buddyService.setBuddyAcceptedWithSecretUserInfo(actingUser.getOwnPrivateData().getUserAnonymizedId(),
-						connectResponseMessageEntity.getBuddyId(),
-						connectResponseMessageEntity.getRelatedUserAnonymizedId().get(),
-						connectResponseMessageEntity.getSenderNickname());
+				buddyService.setBuddyAcceptedWithSecretUserInfo(actingUser, connectResponseMessageEntity);
```
aristotelos: Isn't it a bit weird to pass an entity to a service? I know that we do that more often, but I would rather pass a DTO.

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectMessage.java
```diff
@@ -4,49 +4,122 @@
...
 	}
 
+	public List<String> getDeviceNames()
+	{
+		return Arrays.stream(deviceNames.split(DeviceBase.DEVICE_NAMES_SEPARATOR)).collect(Collectors.toList());
+	}
+
+	public List<UUID> getDeviceAnonymizedIds()
+	{
+		return Arrays.stream(deviceAnonymizedIds.split(DeviceBase.DEVICE_NAMES_SEPARATOR)).map(UUID::fromString)
```
aristotelos: It is a bit strange to reuse the `DEVICE_NAMES_SEPARATOR` here. It is cleaner to introduce a new constant.

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectMessage.java
```diff
@@ -4,49 +4,122 @@
...
+				.collect(Collectors.toList());
+	}
+
+	public List<Boolean> getDeviceVpnConnectionStatuses()
+	{
+		return Arrays.stream(deviceVpnConnectionStatuses.split(DeviceBase.DEVICE_NAMES_SEPARATOR)).map(Boolean::parseBoolean)
```
aristotelos: It is a bit strange to reuse the `DEVICE_NAMES_SEPARATOR` here. It is cleaner to introduce a new constant.

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectMessage.java
```diff
@@ -4,49 +4,122 @@
...
 		super();
 	}
 
-	protected BuddyConnectMessage(BuddyInfoParameters buddyInfoParameters, String message, UUID buddyId)
+	protected BuddyConnectMessage(BuddyInfoParameters buddyInfoParameters, String message, UUID buddyId, Set<UserDevice> devices)
 	{
 		super(buddyInfoParameters, message);
 		this.buddyId = buddyId;
+		List<UserDevice> sortedDevices = devices.stream()
+				.sorted((d1, d2) -> d1.getDeviceAnonymizedId().compareTo(d2.getDeviceAnonymizedId()))
```
aristotelos: Isn't the sorting a bit superfluous here? Why not just convert it to a list?

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectMessage.java
```diff
@@ -4,49 +4,122 @@
...
+import nu.yona.server.device.entities.UserDevice;
 import nu.yona.server.messaging.entities.BuddyMessage;
 
 @Entity
 public abstract class BuddyConnectMessage extends BuddyMessage
 {
+	private static final int UUID_LENGTH = 36;
+	private static final int BOOLEAN_LENGTH = 5;
+	private static final int MAX_NUM_OF_DEVICES = 10;
+	private static final double ENCRYPTION_OVERHEAD_FACTOR = 1.2; // This is a guess
```
aristotelos: This factor should move in a different class.

appservice/src/intTest/groovy/nu/yona/server/DeviceTest.groovy
```diff
@@ -0,0 +1,145 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Stichting Yona Foundation
```
aristotelos: 2017

appservice/src/intTest/groovy/nu/yona/server/UpdateBuddyUserInfoTest.groovy
```diff
@@ -40,6 +40,7 @@ class UpdateBuddyUserInfoTest extends AbstractAppServiceIntegrationTest
 
 		User richardAfterProcess = appService.reloadUser(richard)
 		richardAfterProcess.buddies[0].nickname == "Bobby"
```
aristotelos: Please create a task to remove the nickname from the buddy DTO after the app has been updated.

appservice/src/main/java/nu/yona/server/subscriptions/rest/BuddyController.java
```diff
@@ -321,12 +320,6 @@ private void addEditLink(ControllerLinkBuilder selfLinkBuilder, BuddyResource bu
 			buddyResource.add(selfLinkBuilder.withRel(JsonRootRelProvider.EDIT_REL));
 		}
 
-		private void addUserPhotoLink(BuddyResource buddyResource)
```
aristotelos: Nice reduction!

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -405,15 +472,66 @@ private void setSslRootCertificateCn()
...
+		}
+
+		Optional<UserDeviceDto> getDevice()
+		{
+			return (deviceName == null) ? Optional.empty()
+					: Optional.of(UserDeviceDto.createPostPutInstance(deviceName, deviceOperatingSystem));
+		}
+	}
+
+	enum Purpose
```
aristotelos: This enum was introduced to have an overview of all states of a user DTO, isn't it? I think a better name for it may be `UserResourcePresentationState`.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1706,6 +1706,48 @@ definitions:
...
+    enum:
+    - ANDROID
+    - IOS
+    - UNKNOWN
+  DeviceBase:
+    type: object
+    properties:
+      name:
+        type: string
+        description: Name of the device running the Yona app
```
aristotelos: Should not contain a colon

core/src/main/java/nu/yona/server/subscriptions/service/BuddyUserPrivateDataDto.java
```diff
@@ -0,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2015, 2017 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License,
```
aristotelos: 2017

core/src/main/java/nu/yona/server/subscriptions/service/UserPrivateDataBaseDto.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+import nu.yona.server.device.service.DeviceBaseDto;
+import nu.yona.server.goals.service.GoalDto;
+
+public class UserPrivateDataBaseDto
```
aristotelos: This has no JSON annotation anymore. Are the `@JsonIgnore` properties in this class needed?

core/src/main/resources/messages/messages.properties
```diff
@@ -66,6 +67,7 @@ error.encoding.public.key=Error encoding the public key
 error.decoding.public.key=Error decoding the public key
 
 error.buddy.not.found=Buddy with ID ''{0}'' not found
+error.buddy.not.found.for.user=User with ID ''{0}'' is not a buddy of the user user with ID ''{1}''
```
aristotelos: user user duplication

core/src/main/resources/messages/messages.properties
```diff
@@ -98,6 +100,8 @@ error.goal.update.time.before.original=Creation time of updated goal ({0}) is be
 error.no.device.request.present=No new device request present for mobile number ''{0}''
 error.device.request.expired=New device request expired for mobile number ''{0}''
 error.device.request.invalid.password=Invalid temporary password to use new device request for mobile number ''{0}''. Maybe the request was replaced with a different secret?
+error.device.unknown.operating.system=Operating system ''{0}'' is not supported.
+error.device.invalid.device.name=Device name ``{0}`` is invalid. A valid device name is not longer than {1} characters and does not contain a ''{2}''.
```
aristotelos: Wrong quotes (`` instead of '')

core/src/main/resources/messages/messages_nl.properties
```diff
@@ -98,6 +100,9 @@ error.goal.update.time.before.original=Creatietijd van aangepast doel ({0}) ligt
 error.no.device.request.present=Er bestaat geen ''new device request'' voor de gebruiker met het ID ''{0}''
 error.device.request.expired=Het ''new device request'' voor gebruiker met ID ''{0}'' is verlopen
 error.device.request.invalid.password=Het tijdelijke wachtwoord voor het ''new device request'' is ongeldig. Is het verzoek misschien vervangen door een ander?
+error.device.unknown.operating.system=Besturingssysteem ''{0}'' wordt niet ondersteund.
+error.device.invalid.device.name=De apparaatnaam ``{0}`` is niet toegestaan. Een geldige apparaatnaam is niet langer dan {1} tekens en bevat geen ''{2}''.
```
aristotelos: Wrong quotes

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -277,6 +286,33 @@ private User createUserEntity(UserDto user, UserSignUp signUp)
 		return userEntity;
 	}
 
+	private void addDevicesToEntity(UserDto userDto, User userEntity)
+	{
+		userDto.getOwnPrivateData().getDevices().stream().map(d -> (UserDeviceDto) d).map(d -> createUserDeviceEntity(d, 0))
```
aristotelos: Will the device ID always be 0?

yonabuild:

yonabuild:
Build started sha1 is merged.

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
Bert-R: Indeed, that's preferable. At the moment, it's done with the statement you see above. An ``assertResponseStatusOk`` would make it a little shorter. I propose to do that in another PR, as this is big enough already.

yonabuild:
Build finished.

appservice/src/intTest/groovy/nu/yona/server/DeviceTest.groovy
Bert-R: Updated

appservice/src/intTest/groovy/nu/yona/server/UpdateBuddyUserInfoTest.groovy
Bert-R: I have a note on my desk to do that, including the embedded goals

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
Bert-R: The enum was created inside the {{UserResourceAssembler}}. There {{Purpose}} was quite OK, but due to a compiler bug, I had to move it out of that class, making the name less meaningful. I'll rename it into ``UserResourceRepresentation``. Hope that's fine.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
Bert-R: Done

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectMessage.java
Bert-R: Currently, it's an estimate only. If we move it elsewhere, it might be used more widely and might prove to be wrong. I favor keeping at the place we use it.

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectMessage.java
Bert-R: You're right. The order doesn't matter, as long as it's fixed. I'l update it.

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectMessage.java
Bert-R: I'll define a constant in this class:

```
String SEPARATOR = DeviceBase.DEVICE_NAMES_SEPARATOR;
```

Then use that constant throughout this class. That way, we always use ``DEVICE_NAMES_SEPARATOR``, but it's less prominent. OK?

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectMessage.java
Bert-R: See above.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
Bert-R: I woud like to, but I don't know how we could do that. In Groovy, you could do ``eachWithIndex``, but Java doesn't have that.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyUserPrivateDataDto.java
Bert-R: Done

core/src/main/java/nu/yona/server/subscriptions/service/UserPrivateDataBaseDto.java
Bert-R: Yes, as the subtypes do JSON serialization. That includes the inherited properties.

core/src/main/resources/messages/messages.properties
Bert-R: There were a few more of these. Corrected them all.

core/src/main/resources/messages/messages.properties
Bert-R: There were a few more of these. Corrected them all.

core/src/main/resources/messages/messages_nl.properties
Bert-R: There were a few more of these. Corrected them all.

core/src/main/resources/messages/messages.properties
Bert-R: Done

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
Bert-R: The initial one. When a device is added, it will get a higher sequence number.

core/src/main/resources/messages/messages_nl.properties
Bert-R: There were a few more of these. Corrected them all.

yonabuild:

yonabuild:
Build started sha1 is merged.

Bert-R:
Pushed an update addressing the review comments so far.

yonabuild:
Build finished.

core/src/testUtils/groovy/nu/yona/server/test/AppService.groovy
Bert-R: I've created [YD-506](http://jira.yona.nu/browse/YD-506) for this and implemented it already, based on the change set that's in this PR.

appservice/src/intTest/groovy/nu/yona/server/UpdateBuddyUserInfoTest.groovy
Bert-R: Done: [YD-505](http://jira.yona.nu/browse/YD-505).

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
aristotelos: I don't see an increment of sequence number anywhere. Maybe good to add a test around it?
Will the user anonymized ID stay the same for current users?

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
aristotelos: Could maybe be done using `IntStream.range(0, numDevices).foreach(...)`? See [here](https://stackoverflow.com/questions/18552005/is-there-a-concise-way-to-iterate-over-a-stream-with-indices-in-java-8).

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectMessage.java
aristotelos: Perfect

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
aristotelos: Perfect

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
Bert-R: I have seen that, but that doesn't make it more readable.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
Bert-R: Correct. That'll come with [YD-497](http://jira.yona.nu/browse/YD-497).

aristotelos: Aha, I see. With that, I can approve this PR.

PR #420, opened 2017-11-30
YD-506 Use assertResponseStatus...

Bert-R:
It's now used in all applicable places. To make the assertions easier to reuse, they are all moved to the class CommonAssertions.

Along with this fixed a mistake in the test OverwriteUserTest.'Overwrite Richard while being a buddy of Bob and verify Richard does not receive Bob\'s goal conflicts anymore'. That tried to fetch the messages of the overwritten user, but that user does not exist anymore, so it results in a 400. Not sure why this ever passed.

### DIFF
--- a/core/src/testUtils/groovy/nu/yona/server/test/CommonAssertions.groovy
+++ b/core/src/testUtils/groovy/nu/yona/server/test/CommonAssertions.groovy
@@ -1,0 +1,272 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Stichting Yona Foundation
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v.2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at https://mozilla.org/MPL/2.0/.
+ *******************************************************************************/
+package nu.yona.server.test
+
+import java.time.Duration
+import java.time.LocalDate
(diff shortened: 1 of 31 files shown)


yonabuild:

yonabuild:
Build started sha1 is merged.

core/src/testUtils/groovy/nu/yona/server/test/CommonAssertions.groovy
```diff
@@ -0,0 +1,272 @@
...
+package nu.yona.server.test
+
+import java.time.Duration
+import java.time.LocalDate
+import java.time.ZonedDateTime
+
+import groovy.json.*
+import nu.yona.server.YonaServer
+
+class CommonAssertions extends Service
```
aristotelos: Shouldn't this class be marked as `static`?

Bert-R: Marking it ``static`` doesn't make a difference. If it were Java, I would add a private constructor with a comment, but that feels as overkill for a scripting language. What do you think?

aristotelos: I agree, I now read that Java has no such concept.

PR #427, opened 2018-01-03
YD-523 Ignore network activity if app not opened

Bert-R:
If the user has an "old account", without device and they did not open the Yona app yet after the migration step to add a device was introduced, they still do not have a device. In such a scenario, we now ignore the app activities till they opened the app and with that got a device registered.

Along with this addressed an issue with the default value of the device index in the NetworkActivityDto.


yonabuild:

yonabuild:
Build started sha1 is merged.

yonabuild:
Build finished.

PR #428, opened 2018-01-06
YD-524 Correct user entity locking and remove superfluous devices

Bert-R:
All updates to the User entity are now done while owning a pessimistic lock database lock. This lead to some deadlocks where updates happened in different transactions on the same thread. These issues have been resolved through changes in the exception-based rollback.

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
@@ -254,7 +254,7 @@ public class BuddyService
 	@Transactional
 	public void removeBuddyAfterConnectRejection(UUID idOfRequestingUser, UUID buddyId)
 	{
-		userService.getValidatedUserById(idOfRequestingUser);
+		userService.assertValidatedUser(userService.getUserEntityById(idOfRequestingUser));
 		Buddy buddy = buddyRepository.findOne(buddyId);
 
 		if (buddy != null)
--- a/core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
@@ -254,25 +254,26 @@ public class BuddyService
 	@Transactional
 	public void removeBuddyAfterConnectRejection(UUID idOfRequestingUser, UUID buddyId)
 	{
-		User user = userService.getValidatedUserById(idOfRequestingUser);
+		userService.getValidatedUserById(idOfRequestingUser);
 		Buddy buddy = buddyRepository.findOne(buddyId);
 
 		if (buddy != null)
 		{
-			removeBuddy(user, buddy);
+			removeBuddy(idOfRequestingUser, buddy);
 		}
 		// else: buddy already removed, probably in response to removing the user
 	}
 
-	private void removeBuddy(User user, Buddy buddy)
+	private void removeBuddy(UUID userId, Buddy buddy)
 	{
-		user.removeBuddy(buddy);
-		User.getRepository().save(user);
-		buddyRepository.delete(buddy);
+		userService.updateUser(userId, user -> {
+			user.removeBuddy(buddy);
+			buddyRepository.delete(buddy);
 
-		UserAnonymized userAnonymizedEntity = user.getAnonymized();
-		userAnonymizedEntity.removeBuddyAnonymized(buddy.getBuddyAnonymized());
-		userAnonymizedService.updateUserAnonymized(userAnonymizedEntity);
+			UserAnonymized userAnonymizedEntity = user.getAnonymized();
+			userAnonymizedEntity.removeBuddyAnonymized(buddy.getBuddyAnonymized());
+			userAnonymizedService.updateUserAnonymized(userAnonymizedEntity);
+		});
 	}
 
 	@Transactional
@@ -290,7 +291,7 @@ public class BuddyService
 		removeMessagesSentByBuddy(user, buddy);
 		removeBuddyInfoForBuddy(user, buddy, message, DropBuddyReason.USER_REMOVED_BUDDY);
 
-		removeBuddy(user, buddy);
+		removeBuddy(user.getId(), buddy);
 
 		User buddyUser = buddy.getUser();
 		if (buddyUser == null)
@@ -395,7 +396,7 @@ public class BuddyService
 					user.getUserAnonymizedId(), buddy.getId(), user.getDevices(), Status.REJECTED,
 					getDropBuddyMessage(DropBuddyReason.USER_ACCOUNT_DELETED, Optional.empty()));
 		}
-		removeBuddy(user, buddy);
+		removeBuddy(user.getId(), buddy);
 	}
 
 	private UUID getUserAnonymizedIdForBuddy(Buddy buddy)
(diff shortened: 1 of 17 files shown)


yonabuild:

yonabuild:
Build started sha1 is merged.

Bert-R:
I've created a pull request for the first part of this fix. If I'm quick enough with writing the correction logic for the duplicate devices, we can wait with merging this PR till that is complete. Otherwise, we should merge this on itself, as it takes away a significant risk of data corruption during migration steps.

yonabuild:
Build finished.

yonabuild:

yonabuild:
Build started sha1 is merged.

Bert-R:
The added commit completes this PR.

yonabuild:
Build finished.

core/src/test/java/nu/yona/server/device/service/DeviceServiceTest.java
```diff
@@ -515,6 +532,78 @@ public void getDeviceAnonymizedId_byId_tryGetNonExistingId_exception()
...
+		removeDuplicateDefaultDevicesFirstOrSecond(0);
+	}
+
+	@Test
+	public void removeDuplicateDefaultDevices_twoDevicesUseSecond_duplicateRemovedActivitiesMoved()
+	{
+		removeDuplicateDefaultDevicesFirstOrSecond(1);
+	}
+
+	private void removeDuplicateDefaultDevicesFirstOrSecond(int indexToRetain)
```
aristotelos: I'd rather use `@Parameters` here.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -479,45 +474,52 @@ User addUserCreatedOnBuddyRequest(UserDto buddyUserResource)
...
+			logger.info("Updated user with mobile number '{}' and ID '{}'", userDto.getMobileNumber(), userDto.getId());
+			buddyService.broadcastUserInfoChangeToBuddies(userEntity, originalUser);
+		});
+		return createUserDtoWithPrivateData(updatedUserEntity);
+	}
+
+	@Transactional
+	public User updateUser(UUID id, Consumer<User> updateAction)
+	{
+		User user = getUserEntityByIdWithUpdateLock(id);
```
aristotelos: Please add a comment here that we do this locking because read and updates on the `User` are done in the same request for special purposes (upgrade mechanism).

core/src/main/java/nu/yona/server/subscriptions/service/NewDeviceRequestService.java
```diff
@@ -40,16 +40,17 @@
 	@Transactional
 	public NewDeviceRequestDto setNewDeviceRequestForUser(UUID userId, String yonaPassword, String newDeviceRequestPassword)
 	{
-		User userEntity = userService.getValidatedUserById(userId);
+		userService.updateUser(userId, ue -> {
+			User userEntity = userService.getValidatedUserById(ue.getId());
```
aristotelos: Isn't this a bit strange to get the `userEntity` instead of using `ue`?

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -254,25 +254,26 @@ public BuddyDto addBuddyToAcceptingUser(UserDto acceptingUser, BuddyConnectReque
 	@Transactional
 	public void removeBuddyAfterConnectRejection(UUID idOfRequestingUser, UUID buddyId)
 	{
-		User user = userService.getValidatedUserById(idOfRequestingUser);
+		userService.getValidatedUserById(idOfRequestingUser);
```
aristotelos: It would be more speaking to just call a `validate` method instead of a `get`.

core/src/main/java/nu/yona/server/subscriptions/service/PinResetRequestService.java
```diff
@@ -73,7 +73,7 @@ public void sendPinResetConfirmationCode(UUID userId)
 		sendConfirmationCodeTextMessage(user, pinResetConfirmationCode);
 	}
 
-	@Transactional
+	@Transactional(dontRollbackOn = PinResetRequestConfirmationException.class)
```
aristotelos: Does this fix a bug?

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -406,60 +400,61 @@ private void deleteExistingUserToOverwriteIt(String mobileNumber, String userPro
...
-		ConfirmationCode confirmationCode = userEntity.getMobileNumberConfirmationCode();
+		User updatedUserEntity = updateUser(userId, userEntity -> {
+			ConfirmationCode confirmationCode = userEntity.getMobileNumberConfirmationCode();
 
-		assertValidConfirmationCode(userEntity, confirmationCode, userProvidedConfirmationCode,
-				() -> MobileNumberConfirmationException.confirmationCodeNotSet(userEntity.getMobileNumber()),
-				r -> MobileNumberConfirmationException.confirmationCodeMismatch(userEntity.getMobileNumber(),
-						userProvidedConfirmationCode, r),
-				() -> MobileNumberConfirmationException.tooManyAttempts(userEntity.getMobileNumber()));
+			assertValidConfirmationCode(userEntity, confirmationCode, userProvidedConfirmationCode,
```
aristotelos: This method name could be made more speaking to reflect that it does an update underneath.

core/src/test/java/nu/yona/server/device/service/DeviceServiceTest.java
Bert-R: That can't be combined easily with ``SpringJUnit4ClassRunner``. There are tricks for it, but that won't make the test more readable.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
Bert-R: Done.

core/src/main/java/nu/yona/server/subscriptions/service/NewDeviceRequestService.java
Bert-R: Done

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
Bert-R: Done.

core/src/main/java/nu/yona/server/subscriptions/service/PinResetRequestService.java
Bert-R: Without this, the number of attempts for failed confirmations wouldn't be saved.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
Bert-R: I'm not disagreeing, but I doubt whether the code becomes more readable if we would change this into something like ``assertValidConfirmationCodeAndRegisterAttempt``.

yonabuild:

yonabuild:
Build started sha1 is merged.

yonabuild:
Build finished.

PR #431, opened 2018-01-18
YD-529 Fix "No devices" error

Bert-R:
The result of the data migration was not stored in the user anonymized cache.
Now every update to a user also updates the user anonymized cache, provided the update is done in a context where the user private data can be accessed.

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
@@ -508,6 +508,10 @@ public class UserService
 	{
 		User user = getUserEntityByIdWithUpdateLock(id);
 		updateAction.accept(user);
+		if (user.canAccessPrivateData())
+		{
+			userAnonymizedService.updateUserAnonymized(user.getAnonymized()); // Save and update cache
+		}
 		return userRepository.save(user);
 	}
 
--- a/core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
@@ -510,8 +510,8 @@ public class UserService
 		updateAction.accept(user);
 		if (user.canAccessPrivateData())
 		{
-			// The private data is accessible and might be updated
-			// Let the UserAnonymizedService save it to the repository and cache it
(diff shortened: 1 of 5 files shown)


yonabuild:

yonabuild:
Build started sha1 is merged.

yonabuild:
Build finished.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -508,6 +508,10 @@ public User updateUser(UUID id, Consumer<User> updateAction)
 	{
 		User user = getUserEntityByIdWithUpdateLock(id);
 		updateAction.accept(user);
+		if (user.canAccessPrivateData())
```
aristotelos: The assumption is that if the user anonymized is updated via the `updateAction`, this can only happen when the private data can be accessed (because the relation is encrypted). This I think is valid (it could be documented here).

Bert-R: Like this?
```
			// The private data is accessible and might be updated
			// Let the UserAnonymizedService save it to the repository and cache it
```

yonabuild:

yonabuild:
Build started sha1 is merged.

yonabuild:

yonabuild:
Build finished.

yonabuild:
Build started sha1 is merged.

yonabuild:
Build finished.

core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -510,8 +510,8 @@ public User updateUser(UUID id, Consumer<User> updateAction)
 		updateAction.accept(user);
 		if (user.canAccessPrivateData())
 		{
-			// The private data is accessible and might be updated
-			// Let the UserAnonymizedService save it to the repository and cache it
+			// The private data is accessible and might be updated, including the UserAnonymized
+			// Let the UserAnonymizedService save save that to the repository and cache it
```
aristotelos: save save

yonabuild:
Build finished.

PR #435, opened 2018-02-05
 YD-517 Corrected GoalDto.equals

Bert-R:
And added a unit test for it.

### DIFF
--- a/core/src/main/java/nu/yona/server/goals/service/GoalDto.java
+++ b/core/src/main/java/nu/yona/server/goals/service/GoalDto.java
@@ -29,6 +29,11 @@ import nu.yona.server.goals.entities.TimeZoneGoal;
 import nu.yona.server.rest.PolymorphicDto;
 import nu.yona.server.util.TimeUtil;
 
+/**
+ * The base goal DTO class.<br/>
+ * NOTE: The ID is always available except when a new goal is POSTed. To keep the API simple, the DTO always has an ID. In case of
+ * a freshly posted goal, it is an all-zeros UUID.
+ */
 @JsonRootName("goal")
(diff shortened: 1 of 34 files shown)


core/src/main/java/nu/yona/server/goals/service/GoalDto.java
```diff
@@ -45,7 +45,7 @@
 	protected GoalDto(UUID id, Optional<LocalDateTime> creationTime, Optional<LocalDateTime> endTime, UUID activityCategoryId,
 			boolean mandatory)
 	{
-		this.id = id;
+		this.id = (id == null) ? new UUID(0, 0) : id;
```
aristotelos: Isn't it possible to make `id` an `Optional<UUID>`?

Bert-R: I played with that. It gives a slightly more elegant solution, but cannot be done because ``Optional`` is not serializable. And ``GoalDto`` needs to be serializable, as it's cached in Hazelcast.

aristotelos: OK, we could add a comment about that to explain the choice. But can't we make the public `getGoalId` an `Optional<UUID>` and keep the serialization details private?

Bert-R: It would formally be right to make it optional when the zero UUID is set, but practically, it's a lot of clutter for no value. The scenario where we do not have a UUID is limited to the case where a new goal DTO is serialized to us. That scenario does not retrieve the ID from the DTO but generates a new one. All scenarios that take the ID from the DTO can rely on the availability of the ID.

Given that, I'd rather explain this story in a comment in ``GoalDto``.

PR #438, opened 2018-02-10
YD-540 New Smootwall places urlcategory differently

Bert-R:
Earlier versions of Smoothwall placed the urlcategory under requesttags. The current version puts it under tagset. Updated the Perl script accordingly.


kti-richard:
Looks good

PR #439, opened 2018-02-16
YD-500 Create VPN account per device

Bert-R:
This includes the following:
* UserDevice got a legacy account marker, indicating this VPN account doesn't use the <deviceIndex> suffix
* User creation logic updated: creation of VPN account becomes part of the initial device creation
* User deletion logic updated: deletion of VPN account is part of device deletion
* VPN profile resource added to UserDeviceDto
* Apple mobile config became device-specific rather than user-specific
* The VPN profile resource embedded in User now deprecated (removed from the Swagger spec)
* The Perl script updated so that it passes the device index in the network activity event.
* The property deviceIndex made mandatory in NetworkActivityDto.
* Upgrade created: moves the VPN password to the device and sets the legacy account marker on Device.


appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -712,9 +660,8 @@ private void addUserPhotoLinkIfPhotoPresent(UserResource userResource)
 
 		private void addAppleMobileConfigLink(UserResource userResource)
```
aristotelos: Is there a task to remove this as well?

appservice/src/main/java/nu/yona/server/device/rest/DeviceController.java
```diff
@@ -136,6 +155,39 @@
...
+		HttpHeaders headers = new HttpHeaders();
+		headers.setContentType(new MediaType("application", "x-apple-aspen-config"));
+		try (CryptoSession cryptoSession = CryptoSession.start(password, () -> userService.canAccessPrivateData(userId)))
+		{
+			return new ResponseEntity<>(getUserSpecificAppleMobileConfig(deviceService.getDevice(deviceId)), headers,
+					HttpStatus.OK);
+		}
+	}
+
+	private byte[] getUserSpecificAppleMobileConfig(UserDeviceDto device)
```
aristotelos: Rename to `getDeviceSpecificAppleMobileConfig`

core/src/main/java/nu/yona/server/device/entities/UserDevice.java
```diff
@@ -16,18 +16,31 @@
...
+	enum VpnAccountType
+	{
+		LEGACY, STANDARD
+	}
+
 	@Type(type = "uuid-char")
 	@Column(name = "user_private_id")
 	private UUID userPrivateId;
 
+	@Convert(converter = StringFieldEncryptor.class)
```
aristotelos: Why is it important to encrypt the VPN account type?

core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
```diff
@@ -146,15 +144,15 @@ public UUID getNamedMessageSourceId()
...
 	}
 
-	public UUID getVpnLoginId()
-	{
-		// these are the same for performance
-		return userAnonymizedId;
-	}
-
-	public String getVpnPassword()
+	public Optional<String> takeVpnPassword()
```
aristotelos: I think `getAndClearVpnPassword` is a bit clearer.

core/src/main/java/nu/yona/server/subscriptions/service/UserDto.java
```diff
@@ -228,10 +229,21 @@ public static UserDto createInstanceWithPrivateData(User userEntity, Set<BuddyDt
...
 
+	// YD-541: Remove this method
+	private static Optional<VPNProfileDto> createVpnProfileDto(User userEntity)
+	{
+		return determineDefaultDevice(userEntity).map(VPNProfileDto::createInstance);
+	}
+
+	private static Optional<UserDevice> determineDefaultDevice(User userEntity)
+	{
+		return userEntity.getDevices().stream().filter(d -> d.getDeviceAnonymized().getDeviceIndex() == 0).findAny();
```
aristotelos: Isn't this a bit duplicate with `DeviceService.getDefaultDeviceId`?

core/src/main/java/nu/yona/server/subscriptions/service/migration/MoveVpnPasswordToDevice.java
```diff
@@ -0,0 +1,40 @@
...
+
+@Component
+@Order(2) // Never change the order or remove a migration step. If the step is obsolete, make it empty
+public class MoveVpnPasswordToDevice implements MigrationStep
+{
+	@Override
+	public void upgrade(User user)
+	{
+		Optional<String> vpnPassword = user.takeVpnPassword();
+		vpnPassword.ifPresent(p -> setPassworOnDefaultDevice(user, p));
```
aristotelos: `setPasswor...` -> `setPassword`

core/src/main/java/nu/yona/server/subscriptions/service/migration/MoveVpnPasswordToDevice.java
```diff
@@ -0,0 +1,40 @@
...
+
+	private void setPassworOnDefaultDevice(User user, String password)
+	{
+		UserDevice defaultDevice = getDefaultDevice(user);
+		defaultDevice.setLegacyVpnAccountPassword(password);
+	}
+
+	private UserDevice getDefaultDevice(User user)
+	{
+		return user.getDevices().stream().sorted(
```
aristotelos: Is this logic not duplicated with `DeviceService.getDefaultDeviceId`?

core/src/test/java/nu/yona/server/device/service/DeviceServiceTest.java
```diff
@@ -169,6 +176,8 @@ public void addDeviceToUser_addFirstDevice_userHasOneDevice()
 
 		UserDevice device = devices.iterator().next();
 		assertDevice(device, startTime, deviceName, operatingSystem, 0);
+
+		assertVpnAccountCreated(device);
```
aristotelos: We could move those new asserts to separate tests, so that the tests test one thing and stay easy to read.

core/src/test/java/nu/yona/server/subscriptions/service/PrivateUserDataMigrationServiceIntegration_stepOrderTest.java
```diff
@@ -0,0 +1,84 @@
...
+	private MoveVpnPasswordToDevice step3;
+
+	@Override
+	protected Map<Class<?>, Repository<?, ?>> getRepositories()
+	{
+		return Collections.emptyMap();
+	}
+
+	@Test
+	public void getPrivateUser_twoVersionsBehind_migratesToCurrentVersion()
```
aristotelos: Please correct the name of this test method.

core/src/test/java/nu/yona/server/subscriptions/service/migration/MoveVpnPasswordToDeviceTest.java
```diff
@@ -0,0 +1,155 @@
...
+	public void upgrade_movePasswordTwice_idemPotent() throws IllegalArgumentException, IllegalAccessException
+	{
+		String vpnPassword = "upgrade_movePasswordTwice_idemPotent";
+		vpnPasswordField.set(userPrivateField.get(richard), vpnPassword);
+
+		UserDevice device = createDevice();
+		richard.addDevice(device);
+
+		migrationStep.upgrade(richard);
+		assertThat(device.getVpnPassword(), equalTo(vpnPassword));
```
aristotelos: I would prefer to have a simple act:
```
migrationStep.upgrade(richard);
migrationStep.upgrade(richard);
```
The first two asserts are superfluous in this test. Could we also use the convention to separate arrange/act/assert by newlines?

core/src/test/java/nu/yona/server/subscriptions/service/migration/MoveVpnPasswordToDeviceTest.java
```diff
@@ -0,0 +1,155 @@
...
+		repositoriesMap.put(DeviceAnonymized.class, deviceAnonymizedRepository);
+		return repositoriesMap;
+	}
+
+	@Test
+	public void upgrade_movePassword_passwordMoved() throws IllegalArgumentException, IllegalAccessException
+	{
+		String vpnPassword = "upgrade_movePassword_passwordMoved";
+		vpnPasswordField.set(userPrivateField.get(richard), vpnPassword);
+
```
aristotelos: I would prefer to remove this newline and add one after the call to `upgrade` to separate arrange/act/assert.

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
Bert-R: No. It isn't strictly necessary to take it out here, as it's based on the requesting device. So, we could stay backward compatible and keep it here. Alternatively, we could add the link to the device and ask the app to take it from there.

appservice/src/main/java/nu/yona/server/device/rest/DeviceController.java
Bert-R: Done

core/src/main/java/nu/yona/server/device/entities/UserDevice.java
Bert-R: It allows grouping users as old and new ones, which would be helpful when trying to break anonymity.

core/src/main/java/nu/yona/server/subscriptions/entities/UserPrivate.java
Bert-R: Yes, I didn't like it very much either, but couldn't find this simple alternative.

core/src/main/java/nu/yona/server/subscriptions/service/UserDto.java
Bert-R: Yes. We could refactor the code now and move the logic to determine the default from ``DeviceService`` to ``UserDto``, but as soon as we address YD-541, we would need to move it back again. That's why I duplicated it here for a moment.

core/src/main/java/nu/yona/server/subscriptions/service/migration/MoveVpnPasswordToDevice.java
Bert-R: Done

core/src/main/java/nu/yona/server/subscriptions/service/migration/MoveVpnPasswordToDevice.java
Bert-R: Yes, this time on entities, rather than DTOs.

core/src/test/java/nu/yona/server/device/service/DeviceServiceTest.java
Bert-R: I don't favor that. It's an assertion that belongs to this test, not a new test.

core/src/test/java/nu/yona/server/subscriptions/service/PrivateUserDataMigrationServiceIntegration_stepOrderTest.java
Bert-R: I read it three times here and felt it was correctly spelled :)
Then looked at the code and found that it doesn't quite match the test. Done.

core/src/test/java/nu/yona/server/subscriptions/service/migration/MoveVpnPasswordToDeviceTest.java
Bert-R: I regularly add assertions to verify the test setup was done right. That's why I added the assertions after the first upgrade. But I'm fine removing them.

core/src/test/java/nu/yona/server/subscriptions/service/migration/MoveVpnPasswordToDeviceTest.java
Bert-R: OK

Bert-R:
Pushed an update

appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
Bert-R: See https://jira.yona.nu/browse/YD-544

core/src/main/java/nu/yona/server/device/entities/UserDevice.java
aristotelos: OK, it adds a bit of clutter (converting enum to string, adding random bytes, removing random bytes when matching etc.) which is better to avoid. I think the risk is not that big.

core/src/main/java/nu/yona/server/subscriptions/service/UserDto.java
aristotelos: OK, so this code will be removed later on.

core/src/main/java/nu/yona/server/subscriptions/service/migration/MoveVpnPasswordToDevice.java
aristotelos: As this migration step will be removed later we can keep it now.

core/src/test/java/nu/yona/server/device/service/DeviceServiceTest.java
aristotelos: Actually, the name of the test should then change to `_userHasOneDeviceWithVpnAccountCreated`

Bert-R: What about ``_deviceProperlyAdded``? Encoding all assertions in that text seems overkill to me.

Bert-R:
Pushed an update that stores the VPN account type as a plain boolean

PR #444, opened 2018-03-03
YD-546 Added device-specific links

Bert-R:
Several device-specific links that were already present on the User resource are now added to the Device resource. Once the apps are updated to take the link from the device, the links should be removed from the user (see YD-544).

This includes the following links:
* ``appleMobileConfig``
* ``postOpenAppEvent``
* ``appActivity``
* ``sslRootCert``

The same applies to one property:
* ``sslRootCertCN``

Along with it:
* Made the usage message of ``HandleDansGuardianLog.pl`` complete by adding ``--verbose``
* Made the name of the second device in ``ActivityTest.'Richard works on two devices'()`` less confusing


appservice/src/main/java/nu/yona/server/device/rest/DeviceController.java
```diff
@@ -397,18 +460,30 @@ public DeviceResource toResource(DeviceBaseDto device)
 			ControllerLinkBuilder selfLinkBuilder = getSelfLinkBuilder(device.getId());
 			addSelfLink(selfLinkBuilder, deviceResource);
 			addEditLink(selfLinkBuilder, deviceResource);
+			if (isRequestingDevice(device))
+			{
+				addPostOpenAppEventLink(deviceResource);
+				addAppActivityLink(deviceResource);
+				addSslRootCertificateLink(deviceResource);
```
aristotelos: ❓  This link is actually not different per device, is it? So why move it here?

appservice/src/main/java/nu/yona/server/device/rest/DeviceController.java
```diff
@@ -350,9 +385,35 @@ public static ControllerLinkBuilder getRegisterDeviceLinkBuilder(UUID userId)
...
+		}
+
+		public boolean isRequestingDevice()
+		{
+			return isRequestingDevice;
+		}
+
+		@JsonProperty("sslRootCertCN")
+		@JsonInclude(Include.NON_EMPTY)
+		public Optional<String> getSslRootCertCn()
```
aristotelos: ❓  This is not different per device too, is it?

appservice/src/main/java/nu/yona/server/device/rest/DeviceController.java
```diff
@@ -397,18 +460,30 @@ public DeviceResource toResource(DeviceBaseDto device)
 			ControllerLinkBuilder selfLinkBuilder = getSelfLinkBuilder(device.getId());
 			addSelfLink(selfLinkBuilder, deviceResource);
 			addEditLink(selfLinkBuilder, deviceResource);
+			if (isRequestingDevice(device))
```
aristotelos: ❓ [Initially](https://wiki.yona.nu/display/DEV/Adding+device+notion?focusedCommentId=18284704#comment-18284704), you were doubting if we should move the links at all. What is the reason that you think it is better to move them after all? It introduces some complexity to requesting the devices collection and marking the requesting device.

appservice/src/main/java/nu/yona/server/device/rest/DeviceController.java
Bert-R: Though it's not specific for the device, it's really device information. I want to group all such information on the device resource.

appservice/src/main/java/nu/yona/server/device/rest/DeviceController.java
Bert-R: See previous comment.

appservice/src/main/java/nu/yona/server/device/rest/DeviceController.java
Bert-R: It's a cleaner design. The information is really device specific, so I feel it's better to put the information in the context it belongs. It's more clear to the app builders.

appservice/src/intTest/groovy/nu/yona/server/AddDeviceTest.groovy
```diff
@@ -86,18 +86,18 @@ class AddDeviceTest extends AbstractAppServiceIntegrationTest
...
-		assertDefaultOwnDevice(defaultDevice)
+		assertDefaultOwnDevice(defaultDevice, false)
 		assert newDevice.name == newDeviceName
 		assert newDevice.operatingSystem == newDeviceOs
 		assertDateTimeFormat(newDevice.registrationTime)
 		assertDateFormat(newDevice.appLastOpenedDate)
 
 		// User self-link uses the new device as "requestingDeviceId"
-		def idOfNewDevice = newDevice._links.self.href - ~/.*\//
+		def idOfNewDevice = newDevice._links.self.href - ~/.*\// - ~/\?.*/
```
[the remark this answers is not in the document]
Bert-R: OK, will do.

Bert-R:
Answered the questions and pushed an update.

PR #445, opened 2018-03-06
YD-547 Fixed disappearing user photo link

Bert-R:
The UserDto class accidentally updated the user photo ID. This was wrong, as updating the user photo link is done differently.

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/UserPhotoTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/UserPhotoTest.groovy
@@ -393,6 +393,36 @@ class UserPhotoTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(richard)
 	}
 
+	def 'Richard\'s photo link is still available after he updated his nickname'()
+	{
+		given:
+		User richard = addRichard()
+		def multipartEntity = MultipartEntityBuilder.create()
+				.addPart("file", new InputStreamBody(new ByteArrayInputStream(Base64.getDecoder().decode(EXAMPLE_PNG_DATA_BASE64)), "image/png", "MyPhoto.png"))
+				.build()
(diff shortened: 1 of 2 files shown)


appservice/src/intTest/groovy/nu/yona/server/UserPhotoTest.groovy
```diff
@@ -393,6 +393,36 @@ class UserPhotoTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(richard)
 	}
 
+	def 'Richard\'s photo link is still available after he updated his nickname'()
+	{
+		given:
+		User richard = addRichard()
+		def multipartEntity = MultipartEntityBuilder.create()
```
aristotelos: 🔧  Please use `uploadUserPhoto(richard)` below

core/src/main/java/nu/yona/server/subscriptions/service/UserDto.java
```diff
@@ -189,7 +189,6 @@ void updateUser(User originalUserEntity)
 		originalUserEntity.setLastName(lastName);
 		originalUserEntity.setMobileNumber(mobileNumber);
 		originalUserEntity.setNickname(privateData.getNickname());
-		originalUserEntity.setUserPhotoId(privateData.getUserPhotoId());
```
aristotelos: Oops... I think this is a left-over from some earlier ideas on how to do user photo updates.

appservice/src/intTest/groovy/nu/yona/server/UserPhotoTest.groovy
Bert-R: Much better indeed. I didn't notice that helper method.

Bert-R:
Pushed an update

PR #446, opened 2018-03-13
Kti helm healthchecks

kti-richard:
- Added some new support files (helm rbac, jenkins slave example)
- Added new profile to wait-for-services script called k8snew.  This should work well with the new health checks and init containers in the pod specs.
- Containers are health checked against /health/ for the springboot containers.
- Containers are prefixed with an init container for DB and LDAP as needed (ldap only on liquibase for now)
- Changed the default storage handling to work faster and with less bootstrap code when using Minikube, and to use more common naming convention for other clusters.  Will still need to see how this impacts dev-build (might need to upgrade k8s there, and add a proper provisioner)
- Most dramatic change is the new values.yaml setup to break the ldap and db out from the Java URL references to be used by the init containers.  New values will need to be setup for the environments if not using the defaults.
- Increased the ldap package version required (new ldap helm version already deployed)

### DIFF
--- a/k8s/02_storage.yaml
+++ b/k8s/02_storage.yaml
@@ -7,13 +7,13 @@ metadata:
     type: local
 spec:
   storageClassName: default
-  persistentVolumeReclaimPolicy: Recycle
+  persistentVolumeReclaimPolicy: Delete
   capacity:
     storage: 3Gi
   accessModes:
     - ReadWriteOnce
   hostPath:
-    path: "/data/yona-persistentvolumes/pv0001"
+    path: "/tmp/.yona-persistentvolumes/pv0001"
 ---
 kind: PersistentVolume
 apiVersion: v1
@@ -23,13 +23,13 @@ metadata:
     type: local
 spec:
   storageClassName: default
-  persistentVolumeReclaimPolicy: Recycle
+  persistentVolumeReclaimPolicy: Delete
   capacity:
     storage: 3Gi
   accessModes:
     - ReadWriteOnce
   hostPath:
-    path: "/data/yona-persistentvolumes/pv0002"
+    path: "/tmp/.yona-persistentvolumes/pv0002"
 ---
 kind: PersistentVolume
 apiVersion: v1
@@ -39,10 +39,10 @@ metadata:
     type: local
 spec:
   storageClassName: default
-  persistentVolumeReclaimPolicy: Recycle
+  persistentVolumeReclaimPolicy: Delete
   capacity:
     storage: 3Gi
   accessModes:
     - ReadWriteOnce
   hostPath:
-    path: "/data/yona-persistentvolumes/pv0003"
+    path: "/tmp/.yona-persistentvolumes/pv0003"
--- a/k8s/02_storage.yaml
+++ b/k8s/02_storage.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
     type: local
 spec:
-  storageClassName: yona-default
+  storageClassName: default
   persistentVolumeReclaimPolicy: Recycle
   capacity:
     storage: 3Gi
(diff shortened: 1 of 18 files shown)


kti-richard:
I have reviewed and fixed the values.yaml on dev-build, mobiquity, beta, and production to support the init containers.  Once deployed, and working fine, we should be able to remove the url_override values if desired.

k8s/02_storage.yaml
```diff
@@ -22,14 +22,14 @@ metadata:
...
-  persistentVolumeReclaimPolicy: Recycle
+  storageClassName: default
+  persistentVolumeReclaimPolicy: Delete
   capacity:
     storage: 3Gi
   accessModes:
     - ReadWriteOnce
   hostPath:
-    path: "/data/yona-persistentvolumes/pv0002"
+    path: "/tmp/.yona-persistentvolumes/pv0002"
```
Bert-R: This requires a change on the hosts, right?

k8s/helm/yona/templates/03_job-liquibase.yaml
```diff
@@ -12,8 +12,37 @@ metadata:
...
+              value: {{ .Values.mariadb.mariadbUser | default "develop" | quote }}
+            - name: DBPASSWORD
+              value: {{ .Values.mariadb.mariadbPassword | default "develop" | quote }}
+            - name: DBHOST
+              value: {{ .Values.mariadb.mariadbHostname | quote }}
+            - name: DBNAME
+              value: {{ .Values.mariadb.mariadbDatabase | quote }}
+            - name: TIMEOUT
+              value: "10s"
+        - name: validateldap
```
Bert-R: It's a bit strange to validate LDAP here, as this job has no LDAP dependency. Only the app service depends on LDAP.

k8s/helm/yona/templates/10_deployment_admin.yaml
```diff
@@ -49,6 +64,14 @@ spec:
...
             - name: config-volume
               mountPath: /opt/app/config
+          livenessProbe:
+            httpGet:
+              path: /health/
+              port: 9080
+            initialDelaySeconds: 60
+            periodSeconds: 15
+            timeoutSeconds: 3
+            failureThreshold: 3
```
Bert-R: This implies the service needs to start in 105 seconds, right? 60 + 15 * 3. Isn't that a but a narrow window? Or am I misunderstanding it?

Bert-R:
> Will still need to see how this impacts dev-build (might need to upgrade k8s there, and add a proper provisioner)
Given that the Mobiquity test server is set up the same way, I expect that would need a similar change. Right?

k8s/02_storage.yaml
kti-richard: This only gets used on the dev-build as best I understand.  The mobiquity does not use the 'infrastructure' components.   /tmp should work OK for the build server with space being the only issue (7G free currently, and db/ldap will not take much in the bootstrap mode).

k8s/helm/yona/templates/03_job-liquibase.yaml
kti-richard: The liquibase container also populates the ldap at bootstrap so it is a dependency.

k8s/helm/yona/templates/10_deployment_admin.yaml
kti-richard: The admin container will not be checked at all for 60 seconds (enough time for springboot to turn up and get the /health endpoint working in my tests).   
After that, the K8S will check /health every 15 seconds, and if it does not respond in 3 seconds, 3 times in a row, K8S will shoot the instance in the head, and start a new one.

kti-richard: We might need to tune these further.

kti-richard:
The Mobiquity does not use the 'infrastructure' components IIRC, so it should not need any persistent volume support.   I have upgraded the config.yaml there for the other requirements though.

k8s/02_storage.yaml
Bert-R: Alternatively, I can merge it and push it through the pipeline at a moment you're available to help resolve any issues that might occur.

PR #448, opened 2018-03-16
K8S API Changes do not allow for mutable labels on containers

kti-richard:
Due to changes that seem to affect us now at the beta 1.9.4 level of k8s, combined with new helm, we can no longer use the dynamic manifest labels because they are immutable now.  It makes logical sense, but it make for tracking a specific pod release when used with rolling upgrades harder.  We have to rely on the rolling upgrade system to keep track of that.
This restriction is not on the Job containers, and since that leads our process, we can still watch for that first in the wait-for-service checks.

### DIFF
--- a/k8s/helm/yona/templates/10_deployment_admin.yaml
+++ b/k8s/helm/yona/templates/10_deployment_admin.yaml
@@ -20,7 +20,6 @@ spec:
     metadata:
       labels:
         app: admin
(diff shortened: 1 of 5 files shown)


Bert-R:
They wait logic looks indeed a bit strange now, but the services look more normal, so it's a bit of give-and-take. I'll merge it, so we can see it on the build server. Next stages, I'll do this evening.

PR #449, opened 2018-03-28
YD-551 Workaround for double encoding of invite URL

Bert-R:
Due to issue spring-projects/spring-hateoas#703, the URL was encoded three times. A workaround is implemented now.


Bert-R:
Going to merge this myself, as it's a small workaround that people are waiting for.

PR #451, opened 2018-04-03
 YD-514 Implement VPN status end point

Bert-R:
The VPN status is captured by means of a sequence of status change events. The current status is simply taken from the last change event.

Along with this:
* Removed the ``isVpnConnected`` property from ``DeviceBase``, as it's functionality is taken over by the last connection status change event
* Simplified the integration tests by implementing the ``postOpenAppEvent`` method on ``Device``


appservice/src/intTest/groovy/nu/yona/server/VpnConnectionStatusEventTest.groovy
```diff
@@ -0,0 +1,163 @@
...
+
+		when:
+		def response = richard.devices[0].postVpnStatus(appService, true)
+
+		then:
+		assertResponseStatusNoContent(response)
+
+		User richardAfterReload = appService.reloadUser(richard)
+		User bobAfterReload = appService.reloadUser(bob)
+		!richard.devices[0].vpnConnected
```
aristotelos: We could move this to a simple test "Initially when Richard is created, VPN status is disconnected" and leave it out of these tests.

appservice/src/intTest/groovy/nu/yona/server/VpnConnectionStatusEventTest.groovy
```diff
@@ -0,0 +1,163 @@
...
+		!bob.buddies[0].user.devices[0].vpnConnected
+		bobAfterReload.buddies[0].user.devices[0].vpnConnected
+
+		def bobMessagesAfterUpdate = appService.getMessages(bob)
+		assertResponseStatusOk(bobMessagesAfterUpdate)
+		def vpnConnectionStatusChangeMessages = bobMessagesAfterUpdate.responseData._embedded?."yona:messages".findAll{ it."@type" == "BuddyVpnConnectionStatusChangeMessage"}
+
+		vpnConnectionStatusChangeMessages.size() == 1
+		vpnConnectionStatusChangeMessages[0]._links.self != null
+		vpnConnectionStatusChangeMessages[0]._links."yona:user".href == bob.buddies[0].user.url
```
aristotelos: Maybe a link to the device would be helpful too?

appservice/src/intTest/groovy/nu/yona/server/VpnConnectionStatusEventTest.groovy
```diff
@@ -0,0 +1,163 @@
...
+ * obtain one at https://mozilla.org/MPL/2.0/.
+ *******************************************************************************/
+package nu.yona.server
+
+import static nu.yona.server.test.CommonAssertions.*
+
+import groovy.json.*
+import nu.yona.server.test.User
+
+class VpnConnectionStatusEventTest extends AbstractAppServiceIntegrationTest
```
aristotelos: There are no tests with multiple devices, which may be good to verify

appservice/src/intTest/groovy/nu/yona/server/VpnConnectionStatusEventTest.groovy
```diff
@@ -0,0 +1,163 @@
...
+		def bobMessagesAfterUpdate = appService.getMessages(bob)
+		assertResponseStatusOk(bobMessagesAfterUpdate)
+		def vpnConnectionStatusChangeMessages = bobMessagesAfterUpdate.responseData._embedded?."yona:messages".findAll{ it."@type" == "BuddyVpnConnectionStatusChangeMessage"}
+
+		vpnConnectionStatusChangeMessages.size() == 2
+		vpnConnectionStatusChangeMessages[0]._links.self != null
+		vpnConnectionStatusChangeMessages[0]._links."yona:user".href == bob.buddies[0].user.url
+		vpnConnectionStatusChangeMessages[0].message == "User disconnected VPN on device '${richard.devices[0].name}'"
+
+		vpnConnectionStatusChangeMessages[1]._links.self != null
```
aristotelos: These asserts on the older message are less relevant for the test

core/src/main/java/nu/yona/server/device/entities/DeviceAnonymized.java
```diff
@@ -37,6 +46,14 @@
 
 	private String appVersion;
 
+	@OneToMany(mappedBy = "deviceAnonymized", cascade = CascadeType.ALL, orphanRemoval = true, fetch = FetchType.LAZY)
+	private Set<VpnStatusChangeEvent> vpnStatusChangeEvents;
+
+	@ManyToOne
+	@JoinColumnsOrFormulas({
+			@JoinColumnOrFormula(formula = @JoinFormula(value = "(SELECT e.id FROM vpn_status_change_events e WHERE e.device_anonymized_id = id ORDER BY e.id DESC LIMIT 1)", referencedColumnName = "id")) })
```
aristotelos: Isn't it more common to order by `eventTime`?

core/src/main/java/nu/yona/server/device/entities/DeviceBase.java
```diff
@@ -60,12 +55,8 @@ public void setName(String name)
 
 	public boolean isVpnConnected()
 	{
-		return isVpnConnected;
-	}
-
-	public void setVpnConnected(boolean isVpnConnected)
-	{
-		this.isVpnConnected = isVpnConnected;
+		return getDeviceAnonymizedIfExisting().flatMap(DeviceAnonymized::getLastVpnStatusChangeEvent)
```
aristotelos: I guess this is defensive programming that if the user is deleted, the buddy will not have load issues? Or is there a real situation in which the device anonymized can be `null`?

core/src/main/java/nu/yona/server/device/service/DeviceService.java
```diff
@@ -412,4 +414,32 @@ private Version parseVersionString(String versionStr)
 			throw DeviceServiceException.invalidVersionString(versionStr);
 		}
 	}
+
+	public void registerVpnStatusChangeEvent(UUID userId, UUID deviceId, boolean isVpnConnected)
```
aristotelos: I first read this as "register to the event". Maybe `recordVpnStatusChangeEvent` is a name that can be less easy misread.

aristotelos:
How will the current app implementation react to a new, unknown message type? Will it ignore it or display it in the message list?

Bert-R:
> Will it ignore it or display it in the message list?

That one we'll have to see. I believe the apps show it with a strange icon. They should actually implement an approach to handle new/unknown messages (e.g. just show a default icon and the text). If the behavior is really bad, we'll have to put this feature behind a feature toggle.

core/src/main/java/nu/yona/server/device/entities/DeviceAnonymized.java
Bert-R: I've chosen for ID as always increments and is indexed already, while the resolution of the time field might be so low that duplicate values occur for events that come in quick succession.

core/src/main/java/nu/yona/server/device/entities/DeviceBase.java
Bert-R: Added this comment:

```
		// Call ..IfExisting, as we might be dealing with a buddy device here where the shared DeviceAnonymized has been deleted
		// by the buddy.
```

appservice/src/intTest/groovy/nu/yona/server/VpnConnectionStatusEventTest.groovy
Bert-R: To me, it helps understanding the test: see there are two such message, and what they contain.

core/src/main/java/nu/yona/server/device/entities/DeviceAnonymized.java
Bert-R: The resolution of the date/time might be too low to tell which one came first for two messages that came in close succession. Since the ID is auto-increment, such issues don't exist, and it's indexed already, so it's cheapest too.

appservice/src/intTest/groovy/nu/yona/server/VpnConnectionStatusEventTest.groovy
Bert-R: The main purpose of these assertions is not to guard the initial status but to guard the test itself. Does the test indeed see a change from "disconnected" to "connected"?

appservice/src/intTest/groovy/nu/yona/server/VpnConnectionStatusEventTest.groovy
Bert-R: It makes sense doing so, but isn't trivial and won't make it consistent. See [YD-556](https://jira.yona.nu/browse/YD-556).

appservice/src/intTest/groovy/nu/yona/server/VpnConnectionStatusEventTest.groovy
Bert-R: Yes, indeed. Added it.

core/src/main/java/nu/yona/server/device/service/DeviceService.java
Bert-R: I've tried reading it in an alternative way, but couldn't. I see "Register VPN status change event" and that's what it does. Also, I think that changing the verb from "register" into "record" won't help anybody.

Am I missing something?

PR #453, opened 2018-04-21
YD-513 Encrypt first and last name

Bert-R:
The first and last name are now stored in ``UserPrivate`` (encrypted) and the fields in ``User`` are left empty. They can be removed once all users are migrated.


core/src/main/java/nu/yona/server/messaging/entities/BuddyMessage.java
```diff
@@ -33,6 +33,14 @@
 	private Optional<UUID> senderUserPhotoId;
 	private byte[] senderUserPhotoIdCiphertext;
 
+	@Transient
+	private String firstName;
```
Bert-R: Strictly speaking, we don't need first and last name on all subtypes of ``BuddyMessage``, but only on ``BuddyConnectMessage`` (actually its subtypes) and ``BuddyDisconnectMessage``.

It could be addressed by introducing another message base type, e.g. ``BuddyRelationshipMessage``, as a base type for ``BuddyConnectMessage`` and ``BuddyDisconnectMessage``, and put first and last name on that.

What do you think of that?

core/src/main/java/nu/yona/server/subscriptions/service/migration/EncryptFirstAndLastName.java
```diff
@@ -0,0 +1,32 @@
...
+{
+	@Autowired
+	private BuddyService buddyService;
+
+	@Override
+	public void upgrade(User user)
+	{
+		UserDto originalUser = UserDto.createInstanceWithPrivateData(user, Collections.emptySet());
+		user.moveFirstAndLastNameToPrivate();
+		buddyService.broadcastUserInfoChangeToBuddies(user, originalUser);
```
aristotelos: Users of the app will see a change message in the log. That might seem a bit strange.

core/src/test/java/nu/yona/server/subscriptions/service/migration/EncryptFirstAndLastNameTest.java
```diff
@@ -0,0 +1,166 @@
...
+		assertThat(message, instanceOf(BuddyInfoChangeMessage.class));
+		BuddyInfoChangeMessage buddyInfoChangeMessage = (BuddyInfoChangeMessage) message;
+		assertThat(buddyInfoChangeMessage.getNewFirstName(), equalTo(firstName));
+		assertThat(buddyInfoChangeMessage.getNewLastName(), equalTo(lastName));
+		assertThat(buddyInfoChangeMessage.getNewNickname(), equalTo(richard.getNickname()));
+		assertThat(buddyInfoChangeMessage.getMessage(), equalTo("User changed personal info"));
+	}
+
+	@Test
+	public void upgrade_moveFirstAndLastNameTwice_idempotent() throws IllegalArgumentException, IllegalAccessException
```
aristotelos: How could it be the case that `upgrade` is called twice?

Bert-R: It's a safety measure. IMHO, upgrade steps need to be idempotent, just to be sure.

PR #454, opened 2018-04-26
 YD-557 Pushed first and last name down

Bert-R:
Introduced a new message type ``BuddyConnectionChangeMessage`` and moved the first and last name there.

Along with this, provided default for missing first or last name. Messages that are created before the first and last name were moved to the private data will not have the first and last name in the messages. To ensure such messages still contain at least something reasonable, a default value is returned.

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectionChangeMessage.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectionChangeMessage.java
@@ -53,16 +53,16 @@ public abstract class BuddyConnectionChangeMessage extends BuddyMessage
 
 	public String getFirstName()
 	{
-		return (firstName == null) ? getPlaceHolderString("message.alternative.first.name") : firstName;
+		return (firstName == null) ? getTranslation("message.alternative.first.name", getSenderNickname()) : firstName;
 	}
 
 	public String getLastName()
 	{
-		return (lastName == null) ? getPlaceHolderString("message.alternative.last.name") : lastName;
+		return (lastName == null) ? getTranslation("message.alternative.last.name", getSenderNickname()) : lastName;
 	}
 
-	private String getPlaceHolderString(String messageId)
+	private String getTranslation(String messageId, String nickname)
 	{
-		return Translator.getInstance().getLocalizedMessage(messageId, getSenderNickname());
+		return Translator.getInstance().getLocalizedMessage(messageId, nickname);
 	}
 }
--- a/core/src/main/resources/messages/messages_nl.properties
+++ b/core/src/main/resources/messages/messages_nl.properties
@@ -18,8 +18,8 @@ message.buddy.device.vpn.disconnect=De gebruiker heeft de VPN verbinding verbrok
 message.temp.nickname={0} {1}
 message.self.nickname={0} (ik)
 message.system.nickname=Yona
-message.alternative.first.name=<{}>
-message.alternative.last.name=<{}>
+message.alternative.first.name=<{0}>
+message.alternative.last.name=<{0}>
 default.device.name=Eerste apparaat
 
 error.invalid.request=Fout request


core/src/main/java/nu/yona/server/exceptions/ResourceBasedException.java
```diff
@@ -100,11 +80,6 @@ protected ResourceBasedException(HttpStatus statusCode, Throwable t, String mess
 		this.statusCode = statusCode;
 	}
 
-	public static void setTranslator(Translator translator)
```
aristotelos: Actually, this public method is a better injection than having an inline `Translator.getInstance()`. For unit testing, it is otherwise impossible to change the dependency.

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectionChangeMessage.java
```diff
@@ -0,0 +1,68 @@
...
+	public void decrypt()
+	{
+		super.decrypt();
+		firstName = SecretKeyUtil.decryptString(firstNameCiphertext);
+		lastName = SecretKeyUtil.decryptString(lastNameCiphertext);
+	}
+
+	public String getFirstName()
+	{
+		return (firstName == null) ? getPlaceHolderString("message.alternative.first.name") : firstName;
```
aristotelos: `getPlaceHolderString` -> `getPlaceholderString` (see [here](https://www.merriam-webster.com/dictionary/placeholder))
But I would prefer `getTranslation("message.alternative.first.name", getSenderNickname())` because it now looks like a string with placeholders is returned.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyDto.java
```diff
@@ -104,16 +104,16 @@ public UserDto getUser()
 		return user;
 	}
 
-	Buddy createBuddyEntity(Translator translator)
```
aristotelos: I prefer the old method for testability. Same for `determineTempNickname`.

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectionChangeMessage.java
```diff
@@ -0,0 +1,68 @@
...
+
+import javax.persistence.Entity;
+import javax.persistence.Transient;
+
+import nu.yona.server.Translator;
+import nu.yona.server.crypto.seckey.SecretKeyUtil;
+import nu.yona.server.messaging.entities.BuddyMessage;
+
+@Entity
+public abstract class BuddyConnectionChangeMessage extends BuddyMessage
```
aristotelos: Can we add a unit test that a `BuddyConnectionChangeMessage` works as expected when `firstName` and `lastName` are `null`? So for instance:
- `getFirstName_firstNameNull_returnsAlternativeContainingNickname`
- `encrypt_firstAndLastNameNull_doesNotThrow`
- `decrypt_firstAndLastNameNull_doesNotThrow`

core/src/main/resources/messages/messages_nl.properties
```diff
@@ -18,6 +18,8 @@ message.buddy.device.vpn.disconnect=De gebruiker heeft de VPN verbinding verbrok
 message.temp.nickname={0} {1}
 message.self.nickname={0} (ik)
 message.system.nickname=Yona
+message.alternative.first.name=<{}>
+message.alternative.last.name=<{}>
```
aristotelos: Why not `<{0}>` as in English?

core/src/main/resources/messages/messages.properties
```diff
@@ -18,6 +18,8 @@ message.buddy.device.vpn.disconnect=User disconnected VPN on device ''{0}''
 message.temp.nickname={0} {1}
 message.self.nickname={0} (me)
 message.system.nickname=Yona
+message.alternative.first.name=<{0}>
```
aristotelos: Does this show `<Nickname>` literally? I hope the app did their html escaping right...

core/src/main/java/nu/yona/server/exceptions/ResourceBasedException.java
Bert-R: We currently don't have tests that depend on this, but if we want to do it, we can use a similar setter in ``Translator``. The code to get the instance in a static context needs to be in ``Translator``, as that's the only one to logically depend upon.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyDto.java
Bert-R: For testability, we can easily set a mock instance on the static member. That way, we can keep the code clean.

core/src/main/resources/messages/messages.properties
Bert-R: Indeed, it shows something like ``<BR>``. The app doesn't do HTML, so that's not an issue.

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectionChangeMessage.java
aristotelos: I still would prefer `getTranslation("message.alternative.first.name", getSenderNickname())` and `getTranslation("message.alternative.last.name", getSenderNickname())`.

Bert-R: OK

core/src/main/java/nu/yona/server/subscriptions/entities/BuddyConnectionChangeMessage.java
Bert-R: Created [YD-558](https://jira.yona.nu/browse/YD-558).

Bert-R:
Pushed an update

PR #455, opened 2018-04-27
YD-559 Fixed invalid lock mode error

Bert-R:
This is fixed by making the UserLockChecker resilient for intermediate flushes.

Along with this, improved the logic in User and BuddyUserPrivateDataDto to fetch the user name from the user entity when that is not migrated yet.


Bert-R:
I'm going to merge this without review, to get it into production before the apps go live in the app stores.

PR #458, opened 2018-05-01
YD-558 Add unit test for BuddyUserPrivateDataDtoTest

Bert-R:
The method to determine the name given some alternatives moved from BuddyConnectionChangeMessage to BuddyUserPrivateDataDtoTest, so a unit test was added for that.

Along with this:
* Moved the common properties of UserPrivate and Buddy to PrivateUserProperties
* Added a logging statement to PrivateUserDataMigrationService, so we can see when the migrations happened. If concurrent migrations happen (which I actually suspect), we get proof through the logging.

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/entities/PrivateUserProperties.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/entities/PrivateUserProperties.java
@@ -1,0 +1,86 @@
+/*******************************************************************************
+ * Copyright (c) 2018 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License, v.
+ * 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *******************************************************************************/
+package nu.yona.server.subscriptions.entities;
+
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+
+import javax.persistence.Convert;
+import javax.persistence.MappedSuperclass;
+
+import nu.yona.server.crypto.seckey.StringFieldEncryptor;
+import nu.yona.server.crypto.seckey.UUIDFieldEncryptor;
+
+@MappedSuperclass
+public abstract class PrivateUserProperties extends EntityWithUuidAndTouchVersion
+{
+
+	@Convert(converter = StringFieldEncryptor.class)
+	private String firstName;
+
+	@Convert(converter = StringFieldEncryptor.class)
+	private String lastName;
+
+	@Convert(converter = StringFieldEncryptor.class)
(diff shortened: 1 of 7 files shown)


core/src/main/java/nu/yona/server/subscriptions/entities/PrivateUserProperties.java
```diff
@@ -0,0 +1,86 @@
...
+	{
+		super(null);
+	}
+
+	public PrivateUserProperties(UUID id, String firstName, String lastName, String nickname)
+	{
+		super(id);
+		this.firstName = Objects.requireNonNull(firstName);
+		this.lastName = Objects.requireNonNull(lastName);
+		this.nickname = nickname;
```
aristotelos: Can `nickname` really be `null`? I think it shouldn't be, or should be `Optional`.

core/src/test/java/nu/yona/server/subscriptions/service/BuddyUserPrivateDataDtoTest.java
```diff
@@ -0,0 +1,147 @@
...
+		user.setPrivateDataMigrationVersion(9999);
+
+		String name = BuddyUserPrivateDataDto.determineName(() -> null, Optional.of(user), User::getFirstName, messageId,
+				nickname);
+
+		assertThat(name, equalTo(messageId + ":" + nickname));
+	}
+
+	@Test
+	public void determineName_userNameAvailable_userName()
```
aristotelos: `userNameAvailable_userName` -> `firstNameAvailable_firstName`

Bert-R: Will do.

core/src/main/java/nu/yona/server/subscriptions/entities/PrivateUserProperties.java
Bert-R: It can indeed be null, in case the user is created on buddy request. Making it ``Optional`` makes sense, but it's not very usual in our entities.

I'll see how big the impact is.

Bert-R: This would imply a major extension of this PR, well beyond its scope.
If we want this, we should do a structural refactoring to use ``Optional`` more widely.

Bert-R:
Pushed an update with the renamed test

core/src/main/java/nu/yona/server/subscriptions/entities/PrivateUserProperties.java
aristotelos: OK, no problem to skip it and do it later.

PR #461, opened 2018-05-08
KTI_Hazelcast_Breakout

kti-richard:
PR for the separation of Hazelcast out of the springboot and into a dedicated cluster.
To support local development a cluster will also be deployed via the Yona helm chart just like with MariaDB and LDAP if the infrastructure flag is enabled.

### DIFF
--- a/k8s/helm/yona/templates/10_deployment_analysis.yaml
+++ b/k8s/helm/yona/templates/10_deployment_analysis.yaml
@@ -20,6 +20,7 @@ spec:
     metadata:
       labels:
         app: analysis
+        build: {{ .Chart.AppVersion | quote }}
         stage: {{ .Values.global.stage | default "develop" }}
     spec:
       initContainers:
@@ -38,8 +39,8 @@ spec:
               value: "10s"
       containers:
         - name: analysis
-          image: 'yonadev/analysisservice:build-{{ .Chart.AppVersion }}'
-          imagePullPolicy: Always
+          image: '{{ .Values.global.imagebase | default "yonadev" }}/analysisservice:build-{{ .Chart.AppVersion }}'
+          #imagePullPolicy: Always
           env:
             - name: YONA_DB_USER_NAME
               value: {{ .Values.mariadb.mariadbUser | default "develop" | quote }}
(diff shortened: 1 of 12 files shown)


kti-richard:
Not sure how this will effect stages that do not use infrastructure extras (ldap/mariadb) as they will not have a hazelcast to work with.  We could consider deactivating the cache handling (though I did not have much luck with that earlier), deploying a persistent hazelcast to those stages, or making a separate helm value to deploy just a hazelcast separate from the mariadb and ldap.

k8s/helm/yona/templates/10_deployment_analysis.yaml
```diff
@@ -38,8 +39,8 @@ spec:
               value: "10s"
       containers:
         - name: analysis
-          image: 'yonadev/analysisservice:build-{{ .Chart.AppVersion }}'
-          imagePullPolicy: Always
+          image: '{{ .Values.global.imagebase | default "yonadev" }}/analysisservice:build-{{ .Chart.AppVersion }}'
+          #imagePullPolicy: Always
```
Bert-R: Is this intentionally commented?

k8s/helm/yona/templates/10_deployment_batch.yaml
```diff
@@ -38,8 +39,8 @@ spec:
               value: "10s"
       containers:
         - name: batch
-          image: 'yonadev/batchservice:build-{{ .Chart.AppVersion }}'
-          imagePullPolicy: Always
+          image: '{{ .Values.global.imagebase | default "yonadev" }}/batchservice:build-{{ .Chart.AppVersion }}'
+          #imagePullPolicy: Always
```
Bert-R: Same here?

kti-richard:
Tested with helm k8s infrastructure deploy, and it works.

k8s/helm/yona/templates/10_deployment_analysis.yaml
kti-richard: Yes - some times I want it, so it is there to not have to look up the variable name.   It could be removed.

k8s/helm/yona/templates/10_deployment_batch.yaml
kti-richard: Same there.

k8s/helm/yona/templates/10_deployment_analysis.yaml
Bert-R: OK

PR #462, opened 2018-05-16
Spec is immutable

seedickcode:
Need to revert a change that snuck in with the hazelcast work.  The labels are immutable, so helm can't deploy with 'upgrade' if they change.


yonabuild:

Bert-R:
add to whitelist

kti-richard:
Created PR with my personal account

PR #464, opened 2018-05-16
YD-554 Improve entity loading

Bert-R:
Made a considerable improvement

I feel we need to do more, especially related to loading the user profile and to message loading.

For message loading, we still need 28 queries in this break down:

| Method | Query count |
|---|-------|
| canAccessPrivateData | 2 |
| getPrivateValidatedUserEntity | 1 |
| tryTransferDirectMessagesToAnonymousDestination | 2 |
| tryProcessUnprocessedMessages | 2 |
| getPrivateValidatedUser | 14 |
| getReceivedMessages | 7 |

The first candidate seems to be to work with a partially loaded user DTO (``getPrivateValidatedUser ``).


analysisservice/src/test/java/nu/yona/server/analysis/service/ActivityUpdateServiceTest.java
```diff
@@ -91,6 +94,8 @@
 	@Mock
 	private ActivityCacheService mockAnalysisEngineCacheService;
 	@Mock
+	private GoalRepository mockGoalRepository;
```
aristotelos: The `ActivityUpdateService` doesn't have `GoalRepository` dependency. So why is this necessary?

analysisservice/src/test/java/nu/yona/server/analysis/service/AnalysisEngineServiceTest.java
```diff
@@ -104,6 +107,8 @@
 	@Mock
 	private ActivityCacheService mockAnalysisEngineCacheService;
 	@Mock
+	private GoalRepository mockGoalRepository;
```
aristotelos: Same comment here, this is not a dependency of `AnalysisEngineService`...

appservice/src/intTest/groovy/nu/yona/server/BuddyValidationTest.groovy
```diff
@@ -166,7 +166,7 @@ class BuddyValidationTest extends AbstractAppServiceIntegrationTest
 		User richard = addRichard()
 		User bob = addBob()
 		bob.emailAddress = "bob@dunn.com"
-		appService.sendBuddyConnectRequest(richard, bob)
+		assertResponseStatusCreated(appService.sendBuddyConnectRequest(richard, bob))
```
aristotelos: Isn't it more logical to place this assertion inside the `sendBuddyConnectRequest` method?

core/src/main/java/nu/yona/server/messaging/service/MessageService.java
```diff
@@ -189,14 +197,24 @@ private void deleteMessages(Collection<Message> messages)
...
+	}
+
+	private MessageSource getAnonymousMessageSource(User user)
+	{
+		return getMessageSource(user.getAnonymousMessageSourceId());
+	}
+
+	private MessageSource getMessageSource(UUID id)
+	{
+		return MessageSource.getRepository().findOne(id);
```
aristotelos: It is better to add the `MessageSourceRepository` as a dependency.

core/src/main/java/nu/yona/server/subscriptions/entities/UserAnonymized.java
```diff
@@ -151,4 +155,13 @@ public void removeDeviceAnonymized(DeviceAnonymized deviceAnonimized)
 	{
 		return devicesAnonymized;
 	}
+
+	public void preloadGoals()
+	{
+		if (!isGoalPreloadDone)
+		{
+			Goal.getRepository().findByUserAnonymizedId(getId()); // Preload in one go
```
aristotelos: Can't we inject the `GoalRepository` dependency as a parameter in the `preloadGoals` method and pass it from the `UserAnonymizedService`?

core/src/main/java/nu/yona/server/util/HibernateStatisticsService.java
```diff
@@ -318,5 +341,74 @@ public long getTransactionCount()
...
 			return transactionCount;
 		}
+
+		public List<String> getSqlStatements()
+		{
+			return sqlStatements;
+		}
+	}
+
+	private static class LogAppender extends AppenderBase<ILoggingEvent>
```
aristotelos: Shouldn't we call this `SqlStatementLogAppender`?

analysisservice/src/test/java/nu/yona/server/analysis/service/ActivityUpdateServiceTest.java
Bert-R: The method ``UserAnonymized.preloadGoals`` uses it and that's being called from the ``ActivityUpdateService``.

analysisservice/src/test/java/nu/yona/server/analysis/service/AnalysisEngineServiceTest.java
Bert-R: See reply above.

appservice/src/intTest/groovy/nu/yona/server/BuddyValidationTest.groovy
Bert-R: :anguished: It's already part of it. The last parameter is a boolean named ``assertSuccess``, with ``true`` as default. I'll rever this change.

core/src/main/java/nu/yona/server/messaging/service/MessageService.java
Bert-R: Agreed.

core/src/main/java/nu/yona/server/util/HibernateStatisticsService.java
Bert-R: If it were a top-level class, I would agree, but given the local context I don't see added value in a longer name.

PR #468, opened 2018-05-26
YD-567 Fixed message processing concurrency issue 

Bert-R:
The actions inside ``prepareMessageCollection`` are now processed within a user-level lock, thus preventing concurrency issues.

The fix brings the side effect that the number of prepared statements for message loading drops 1, from 28 to 27.

This change set is based on the work in PR #466, to prevent merge conflicts. It should be merged once that PR is merged.


appservice/src/intTest/HibernateStatsTest.md
```diff
@@ -4,18 +4,18 @@
 | ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---| ---
 | GetUserWithoutPrivateDataFirst| 0| 0| 0| 0| 0| 0| 1| 0| 0| 0| 1| 0| 1| 0| 1| 0| 0| 0| 0| 0| 0| 0| 0| 3| 3| 1| 1
 | GetUserWithoutPrivateDataSecond| 0| 0| 0| 0| 0| 0| 1| 0| 0| 0| 1| 0| 1| 0| 1| 0| 0| 0| 0| 0| 0| 0| 0| 2| 2| 1| 1
-| GetUserWithPrivateDataFirst| 0| 7| 15| 0| 0| 0| 1| 0| 6| 0| 65| 0| 3| 0| 22| 0| 0| 0| 7| 3| 0| 0| 0| 2| 2| 3| 3
-| GetUserWithPrivateDataSecond| 0| 1| 9| 0| 0| 0| 1| 0| 3| 0| 59| 0| 3| 0| 16| 0| 0| 0| 7| 4| 0| 0| 0| 2| 2| 3| 3
-| GetMessagesFirst| 0| 1| 9| 0| 0| 0| 1| 0| 8| 0| 128| 0| 3| 0| 28| 0| 0| 0| 12| 35| 0| 0| 0| 2| 2| 5| 5
-| GetMessagesSecond| 0| 1| 9| 0| 0| 0| 1| 0| 8| 0| 128| 0| 3| 0| 28| 0| 0| 0| 12| 31| 0| 0| 0| 2| 2| 5| 5
+| GetUserWithPrivateDataFirst| 0| 7| 15| 0| 0| 0| 1| 0| 6| 0| 65| 0| 3| 0| 22| 0| 0| 0| 7| 2| 0| 0| 0| 2| 2| 3| 3
+| GetUserWithPrivateDataSecond| 0| 1| 9| 0| 0| 0| 1| 0| 3| 0| 59| 0| 3| 0| 16| 0| 0| 0| 7| 5| 0| 0| 0| 2| 2| 3| 3
+| GetMessagesFirst| 0| 1| 9| 0| 0| 0| 1| 0| 8| 0| 128| 0| 3| 0| 27| 0| 0| 0| 11| 28| 0| 0| 0| 2| 2| 67| 67
```
aristotelos: Does this mean the number of transactions is much higher now?

core/src/main/java/nu/yona/server/messaging/service/MessageService.java
```diff
@@ -85,41 +87,47 @@
 		return messageSource.getReceivedMessages(pageable, earliestDateTime);
 	}
 
-	// handle in a separate transaction to limit exceptions caused by concurrent calls to this method
-	@Transactional
+	// This is intentionally not marked with @Transactional, as the transaction is explicitly started within the lock inside
+	// analyze(ActivityPayload, Set<ActivityCategoryDto>)
```
aristotelos: `the lock inside <correct method>`

Bert-R: Corrected it.

appservice/src/intTest/HibernateStatsTest.md
Bert-R: Sorry for not noticing that earlier. That number (67) is absurd and at the moment, I have now idea why it is that high. I need to take a closer look, as it could be a performance issue. I wouldn't be concerned if it doubled, but 13 times is wrong.

Bert-R: It's actually not wrong. Each Hibernate call done outside a Spring transaction is done in its own Hibernate transaction. That makes run up this quickly. We could add explicit transaction demarcation around the ``mustXxx`` calls and thus bring the transaction count down, but whether that makes it faster, I doubt and we currently don't have a performance regression test to confirm the impact.

aristotelos: OK, shouldn't we investigate this number still and how to make it lower? This is very high if every get messages results in those high numbers. Are you sure this has no performance impact? Or should we first run a regression test on this branch? I'll merge it so that it can be deployed to the test server.

PR #469, opened 2018-05-31
Changing error settings to include correlation.id

kti-nicole:
Changing server config helm chart to include changing logging pattern

Set logging pattern to add yona.correlation.id


yonabuild:

Bert-R:
Add to whitelist

PR #476, opened 2018-06-13
YD-575 Postpone user anonymized creation

Bert-R:
# Note
This PR is for global feedback only, not for detailed feedback (yet). The question is: do we want to go down this path and make it nice and good (and more complex) or do we accept that the system gets polluted with ``UserAnonymized`` entities that are cleaned up late?

The PR proves that it is possible, as everything works as before, but it also proves that there is quite an added complexity.

# Description
By creating the user anonymized and its related entities till the moment the user confirmed their mobile number, we can simplify the user cleanup mechanism.

User entities are created when the user themselves signs up or when a user invites a new one as buddy. In many such cases, the actual user sign up never happens, either because they initially entered a wrong phone number and don't get past the confirmation process or because the invited buddy does not respond.

For such cases, we need to run a clean-up job with rules like:
* User didn't confirm mobile number in 7 days
* User didn't accept invitation in 30 days

Such a clean-up currently can be done only partially: we can remove the user entity with what belongs to it (UserPrivate, UserDevice, etc.), but we cannot remove the UserAnonymized, the goals, etc. as we cannot correlate them to the user being removed. So for UserAnonymized and related entities, we need another clean-up process, based on inactivity (e.g. the user didn't do anything for the past 6 months). This inevitably is much slower-paced, as we don't want to force users to be active "frequently".

This could be simplified by deferring the creation of the UserAnonymized and related entities till the moment the user confirmed their mobile number. Then we know we have a serious user.

### DIFF
--- a/analysisservice/src/test/java/nu/yona/server/analysis/service/ActivityUpdateServiceTest.java
+++ b/analysisservice/src/test/java/nu/yona/server/analysis/service/ActivityUpdateServiceTest.java
@@ -30,7 +30,6 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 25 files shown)


Bert-R:
I agree and will close this PR.

PR #480, opened 2018-06-19
Update Smoothwall log parser to tail without fifo

kti-richard:
* Uses File::Tail and `cpan File::Tail` on Smoothwall works
Do not know how long it will persist though (through updates).
http://search.cpan.org/~mgrabnar/File-Tail-1.3/Tail.pm
This seems fairly robust instead of trying to develop our own logic

* Since stdin and file tail have different enough setup, split
the core into separate function that the file and stdin will call

* Mode set to 755 because that's where it was during the commit


kti-richard:
This is active on Beta currently

PR #481, opened 2018-06-20
YD-580 Message action handling now transactional

Bert-R:
Till now, DtoManager.handleAction was not marked with @transactional. Due to that, the services calls from implementations (also in subtypes) were individually starting a transaction, so the handling wasn't atomic. This is fixed now.

Along with this corrected the logging of buddy connect requests. The values "new" and "existing" were interchanged, so existing users where logged as new and vice versa.

### DIFF
--- a/core/src/main/java/nu/yona/server/messaging/service/MessageDto.java
+++ b/core/src/main/java/nu/yona/server/messaging/service/MessageDto.java
@@ -13,6 +13,7 @@ import java.util.UUID;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.Transactional;
 
 import com.fasterxml.jackson.annotation.JsonFormat;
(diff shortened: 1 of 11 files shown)


core/src/main/java/nu/yona/server/messaging/service/MessageDto.java
```diff
@@ -165,6 +166,7 @@ public boolean isRead()
 		private SenderInfo.Factory senderInfoFactory;
 
 		@Override
+		@Transactional
```
aristotelos: Can't this be put in `MessageService.TheDtoManager.handleAction`? It seems strange to put it in a DTO.

Bert-R: It's not the DTO but the manager of the DTO. That's a ``Component`` (and always has been). That component needs to be transactionally safe, which is why the annotation belongs here. The ``MessageService.TheDtoManager.handleAction`` is already annotated with ``@Transactional``, but Spring basically suspends that transaction when calling a nontransactional component.

PR #482, opened 2018-06-20
YD-581 Return remote failure status code to caller

Bert-R:
If a remote call fails with e.g. a 503, then the caller will receive that status code.


aristotelos:
Isn't it more logical that if the upstream exception is a 4xx error, that the statuscode is 500? Because the server is wrong in this case, not the client, is it?

Bert-R:
OK, makes sense. I'll update the patch.

Bert-R:
After taking a closer look and running the tests, I don't think it's the right thing to do. App activity reports are passed on to the analysis engine service. Wrong data (e.g. end-before-start) leads to a 400 error, and that's the right thing to do.

PR #484, opened 2018-06-28
Reorganization of the Smoothwall parser code and packaging

kti-richard:
* Layout as per Smoothwall OEM module will allow for easier install
* Removed material no longer needed (fifo creation tools)
* Setup a makefile to bundle a tarball for easy deploy to Smoothwall
* Broke out the generic 'options' variable into specific variables
for the analysys url and log file to parse.
* Added an exclude to gitignore for the smoothwall bin directory


scripts/smoothwall/README.md
```diff
@@ -0,0 +1,38 @@
+# Yona Smoothwall Log Parser
+
+This program will watch the output of the danguardian access logs for the json data with the categorization.  It will cross reference this against a list of categories to watch (via API request) and will forward the payloads onto the Analysis server.
```
Bert-R: Spelling error: Should be DansGuardian instead of danguardian

scripts/smoothwall/README.md
```diff
@@ -0,0 +1,38 @@
...
+### Install on Smoothwall Server
+
+Create a tarball with 
+```
+make
+```
+in this directory, or download from
+
+```
+http://repo.ops.yona.nu/yona_smoothwall_parser_latest.tar.gz
```
Bert-R: We can include this in the server build and automatically upload the binary to that repo, so we're sure the binary is always up to date with the source.

Makes sense?

scripts/smoothwall/README.md
kti-richard: Fixed

scripts/smoothwall/README.md
kti-richard: Yes, it might make sense to add eventually.   I would need to setup a secure method to get it from build to repo server or leverage a similar handling to the helm publish and didn't want to tackle that now.

kti-richard: I think we would also need to do integration testing on it on beta smoothwall so we could then promote it to 'latest' and push to repo ?   Right now that is a 'me' validation, and I would manually will push to latest.

Bert-R: Yes, indeed, that's what we would be able to do then. Let's keep it in mind.
The PR now looks good to me.

PR #485, opened 2018-06-28
Convert hard coded IPs in Jenkinsfile to DNS

kti-richard:
Integration tests are hard coded IPs to the old build server.
Switched to DNS.   (DNS entry still points to old server as of this message for external queries, but is a localhost alias for the new build server)


Bert-R:
Sorry, I didn't think of this one.

PR #486, opened 2018-07-02
YD-562 Use pessimistic update lock

Bert-R:
* Reverted the locking strategy back to SQL pessimistic locking (SELECT ... FOR UPDATE)
* Implemented a multithreaded integration test to test concurrent message processing. The test never failed, but logging showed that the next thread started working before the previous thread committed, thus causing concurrency issues. The new doPreparationsAndCheckCanAccessPrivateData approach allong with the pessimistic locking we used earlier resolves these issues.

Along with this added a defensive check to UserPrivate.

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
@@ -19,6 +19,7 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import javax.annotation.PostConstruct;
+import javax.persistence.LockModeType;
 import javax.transaction.Transactional;
 
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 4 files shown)


core/src/main/java/nu/yona/server/subscriptions/service/UserService.java
```diff
@@ -751,10 +753,40 @@ public static String getPassword(Optional<String> password)
...
+	/**
+	 * This method returns a user entity. The passed on Id is checked whether or not it is set. it also checks that the return
+	 * value is always the user entity. If not an exception is thrown. A pessimistic database write lock is claimed when fetching
+	 * the entity.
+	 * 
+	 * @param id the ID of the user
+	 * @return The user entity (never null)
+	 */
+	@Transactional
+	private User getUserEntityByIdWithUpdateLock(UUID id)
```
aristotelos: Does the `@Transactional` annotation make sense on a private method?

Bert-R: It doesn't, so I've removed it. The method is taken from an old version, where it was public. I made it private but forgot to remove the annotation.

PR #487, opened 2018-07-03
YD-549 Support Firebase instance ID

aristotelos:
- Support to pass the Firebase instance ID when registering a new device
- Support to set or update the Firebase instance ID of an existing device
- Returning the Firebase instance ID with the own user devices

### DIFF
--- a/analysisservice/src/test/java/nu/yona/server/analysis/service/ActivityUpdateServiceTest.java
+++ b/analysisservice/src/test/java/nu/yona/server/analysis/service/ActivityUpdateServiceTest.java
@@ -159,7 +159,7 @@ public class ActivityUpdateServiceTest
 		MessageDestination anonMessageDestinationEntity = MessageDestination
 				.createInstance(PublicKeyUtil.generateKeyPair().getPublic());
 		Set<Goal> goals = new HashSet<>(Arrays.asList(gamblingGoal, gamingGoal, socialGoal, shoppingGoal));
-		deviceAnonEntity = DeviceAnonymized.createInstance(0, OperatingSystem.IOS, "Unknown");
+		deviceAnonEntity = DeviceAnonymized.createInstance(0, OperatingSystem.IOS, "Unknown", Optional.empty());
 		deviceAnonId = deviceAnonEntity.getId();
 		userAnonEntity = UserAnonymized.createInstance(anonMessageDestinationEntity, goals);
 		userAnonEntity.addDeviceAnonymized(deviceAnonEntity);
--- a/appservice/src/intTest/groovy/nu/yona/server/DeviceManagementTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/DeviceManagementTest.groovy
@@ -141,7 +141,7 @@ class DeviceManagementTest extends AbstractAppServiceIntegrationTest
 		when:
 		def newDeviceName = "My iPhone"
 		def newDeviceOs = "ANDROID"
-		def response = appService.registerNewDevice(registerUrl, "Wrong password", newDeviceName, newDeviceOs, "0.1", null)
+		def response = appService.registerNewDevice(registerUrl, "Wrong password", newDeviceName, newDeviceOs, "0.1")
 
 		then:
 		assertResponseStatus(response, 400)
@@ -160,7 +160,7 @@ class DeviceManagementTest extends AbstractAppServiceIntegrationTest
 		when:
 		def newDeviceName = "My iPhone"
 		def newDeviceOs = "ANDROID"
-		def response = appService.registerNewDevice(registerUrl, "Wrong password", newDeviceName, newDeviceOs, "0.1", null)
+		def response = appService.registerNewDevice(registerUrl, "Wrong password", newDeviceName, newDeviceOs, "0.1")
 
 		then:
 		assertResponseStatus(response, 400)
--- a/appservice/src/intTest/groovy/nu/yona/server/DeviceManagementTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/DeviceManagementTest.groovy
@@ -13,7 +13,6 @@ import groovy.json.*
 import nu.yona.server.test.CommonAssertions
 import nu.yona.server.test.Device
 import nu.yona.server.test.User
-import spock.lang.IgnoreRest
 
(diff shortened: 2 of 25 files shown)


analysisservice/src/test/java/nu/yona/server/analysis/service/ActivityUpdateServiceTest.java
```diff
@@ -159,7 +159,7 @@ public void setUp()
 		MessageDestination anonMessageDestinationEntity = MessageDestination
 				.createInstance(PublicKeyUtil.generateKeyPair().getPublic());
 		Set<Goal> goals = new HashSet<>(Arrays.asList(gamblingGoal, gamingGoal, socialGoal, shoppingGoal));
-		deviceAnonEntity = DeviceAnonymized.createInstance(0, OperatingSystem.IOS, "Unknown");
+		deviceAnonEntity = DeviceAnonymized.createInstance(0, OperatingSystem.IOS, "Unknown", Optional.empty());
```
Bert-R: Only one out of the 11 calls to ``DeviceAnonymized.createInstance`` passes a nonempty value for the Firebase ID, so please add an overload that does not have the Firebase ID and make that pass the empty value to the current implementation.

appservice/src/intTest/groovy/nu/yona/server/DeviceManagementTest.groovy
```diff
@@ -139,7 +141,7 @@ class DeviceManagementTest extends AbstractAppServiceIntegrationTest
 		when:
 		def newDeviceName = "My iPhone"
 		def newDeviceOs = "ANDROID"
-		def response = appService.registerNewDevice(registerUrl, "Wrong password", newDeviceName, newDeviceOs, "0.1")
+		def response = appService.registerNewDevice(registerUrl, "Wrong password", newDeviceName, newDeviceOs, "0.1", null)
```
Bert-R: Make the Firebase device ID an optional parameter, so not everyone needs to bother.

core/src/main/java/nu/yona/server/device/entities/DeviceAnonymized.java
```diff
@@ -60,17 +62,20 @@ protected DeviceAnonymized()
...
 
-	private DeviceAnonymized(UUID id, int deviceIndex, OperatingSystem operatingSystem, String appVersion)
+	private DeviceAnonymized(UUID id, int deviceIndex, OperatingSystem operatingSystem, String appVersion,
+			Optional<String> firebaseInstanceId)
 	{
 		super(id);
 		this.deviceIndex = deviceIndex;
 		this.operatingSystem = operatingSystem;
 		this.appVersion = Objects.requireNonNull(appVersion);
+		this.firebaseInstanceId = Objects.requireNonNull(firebaseInstanceId).orElse(null);
```
Bert-R: The call to ``Objects.requireNonNull`` doesn't add value here. If ``firebaseInstanceId`` is null, ``Objects.requireNonNull`` would throw an NPE, but that would also occur because of the call to ``orElse``.

core/src/main/java/nu/yona/server/device/entities/DeviceAnonymized.java
```diff
@@ -148,4 +153,14 @@ public void addVpnStatusChangeEvent(VpnStatusChangeEvent event)
 	{
 		return Optional.ofNullable(lastVpnStatusChangeEvent);
 	}
+
+	public Optional<String> getFirebaseInstanceId()
+	{
+		return Optional.ofNullable(firebaseInstanceId);
+	}
+
+	public void setFirebaseInstanceId(Optional<String> firebaseInstanceId)
```
Bert-R: Should we allow clearing it? I don't think so.

analysisservice/src/test/java/nu/yona/server/analysis/service/ActivityUpdateServiceTest.java
aristotelos: All those 10 calls that do not pass it are from test code. Adding a new method for test code only seems strange.

appservice/src/intTest/groovy/nu/yona/server/DeviceManagementTest.groovy
aristotelos: OK, good point

core/src/main/java/nu/yona/server/device/entities/DeviceAnonymized.java
aristotelos: Simplified

core/src/main/java/nu/yona/server/device/entities/DeviceAnonymized.java
aristotelos: I think it is possible. See [onTokenRefresh documentation](https://firebase.google.com/docs/reference/android/com/google/firebase/iid/FirebaseInstanceIdService.html#onTokenRefresh()): some of those could apply, isn't it?

aristotelos: Correction: [onNewToken](https://firebase.google.com/docs/reference/android/com/google/firebase/messaging/FirebaseMessagingService#onNewToken(java.lang.String)) since June 28. So maybe Google now thinks it is never cleared. I'm not entirely sure if there could be a situation in which the app would like to revoke the token.

analysisservice/src/test/java/nu/yona/server/analysis/service/ActivityUpdateServiceTest.java
Bert-R: That was my first thought, but there is one from non-test code. That made me switch and propose to add an overload.

core/src/main/java/nu/yona/server/device/entities/DeviceAnonymized.java
Bert-R: I favor going for simple at first. If we run into NPEs, we can enhance it easily.

analysisservice/src/test/java/nu/yona/server/analysis/service/ActivityUpdateServiceTest.java
Bert-R: Nor can I. I was mistaken.

PR #488, opened 2018-07-03
YD-582 Include correlation ID in error response

Bert-R:
This enables the app to include that in error reports that the users send to support, which in turn enables us to find the related log messages.

### DIFF
--- a/core/src/main/java/nu/yona/server/rest/ErrorResponseDto.java
+++ b/core/src/main/java/nu/yona/server/rest/ErrorResponseDto.java
@@ -1,43 +1,78 @@
 /*******************************************************************************
- * Copyright (c) 2016 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License, v.
- * 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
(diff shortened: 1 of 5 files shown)


core/src/main/java/nu/yona/server/rest/ErrorResponseDto.java
```diff
@@ -1,76 +1,101 @@
...
- * 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ * Copyright (c) 2016, 2018 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
  *******************************************************************************/
 package nu.yona.server.rest;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 /**
```
Bert-R: This class has plenty comments that don't add value. Shall we strip them?

aristotelos: No problem to me

Bert-R: Done

PR #489, opened 2018-07-27
Kti istio

kti-nicole:
Adding options to help chart to support istio, and changing the arguments used to connect to mariadb so the connection will work out-of-the-box.


k8s/helm/yona/templates/03_job-liquibase.yaml
```diff
@@ -26,7 +29,7 @@ spec:
             - name: DBPASSWORD
               value: {{ .Values.mariadb.mariadbPassword | default "develop" | quote }}
             - name: DBHOST
-              value: {{ .Values.mariadb.mariadbHostname | quote }}
+              value: {{ .Release.Name }}-mariadb.{{ .Release.Namespace }}.svc.cluster.local
```
kti-richard: In stacks with an external DB, the mariadbHostname is set to an pure DNS or IP Addresss (different from url_override, which is a jdbc url).    Since the url_override is also set in these situations, just use a conditional on that to apply the mariadb.mariadbHostname here.   

Perhaps a better solution would be instead to just remove mairadbHostname from the default values, and then use that as a conditional, but if we're doing that, lets add 'override' to the key as well.

k8s/helm/yona/templates/10_deployment_admin.yaml
```diff
@@ -31,7 +38,7 @@ spec:
             - name: DBPASSWORD
               value: {{ .Values.mariadb.mariadbPassword | default "develop" | quote }}
             - name: DBHOST
-              value: {{ .Values.mariadb.mariadbHostname | quote }}
+              value: {{ .Release.Name }}-mariadb.{{ .Release.Namespace }}.svc.cluster.local
```
kti-richard: Same as jump - need to wrap this in a conditional for external DB

PR #490, opened 2018-08-04
YD-583 Fix Cinderella bug: test failures after midnight

Bert-R:
Fixed some tests and disabled others in the window just after midnight. The window is specific per test and varies between 5 and 30 minutes.

Along with this removed an unused method from AnalysisEngineServiceTest.

### DIFF
--- a/core/src/test/java/nu/yona/server/test/util/JUnitUtil.java
+++ b/core/src/test/java/nu/yona/server/test/util/JUnitUtil.java
@@ -4,16 +4,19 @@
  *******************************************************************************/
 package nu.yona.server.test.util;
 
+import static org.junit.Assume.assumeThat;
 import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.when;
 
 import java.io.Serializable;
 import java.lang.reflect.Field;
+import java.time.ZonedDateTime;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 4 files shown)


kti-richard:
@Bert-R Pipeline should be good for this to go through now

core/src/test/java/nu/yona/server/test/util/JUnitUtil.java
```diff
@@ -121,4 +124,9 @@ public static Field getAccessibleField(Class<?> clazz, String fieldName)
 		field.setAccessible(true);
 		return field;
 	}
-}
+
+	public static void skipBefore(String description, ZonedDateTime now, int hour, int minute)
+	{
+		assumeThat(description, now, ZonedDateTimeMatchers.after(now.withHour(hour).withMinute(minute).withSecond(0)));
```
aristotelos: I think it would be better to make the tests independent of "now" instead of skipping them at midnight. There could be interesting cases that fail at midnight that should not fail. I'll add a task on the board.

Bert-R: That won't be trivial to do. To revive the tests under all circumstances and to also test such things around midnight, it would be better to mock the clock. The you have complete freedom about the scenario.

PR #491, opened 2018-08-11
YD-584 Add app version code

Bert-R:
This coexists with the app version. The app version is a free-text string that is user readable, the app version code is an integer that is incremented for each and every new build of the app, across releases.

The legacy test for the open app event in DeviceTest.groovy has been updated to also create the user in the legacy way.

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/DeviceTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/DeviceTest.groovy
@@ -253,11 +253,13 @@ class DeviceTest extends AbstractAppServiceIntegrationTest
 		def richard = addRichard()
 
 		when:
-		def response = richard.devices[0].postOpenAppEvent(appService, richard.devices[0].operatingSystem, "9.9.9", 2)
+		def appVersion = "9.9.9"
+		def response = richard.devices[0].postOpenAppEvent(appService, richard.devices[0].operatingSystem, appVersion, 2)
 
 		then:
 		assertResponseStatus(response, 400)
 		assert response.responseData.code == "error.device.app.version.not.supported"
+		assert response.responseData.message == "Yona app is out of date and must be updated. Actual version is '$appVersion' but oldest supported version for 'IOS' is '1.0.1'"
 
 		cleanup:
 		appService.deleteUser(richard)
--- a/appservice/src/intTest/groovy/nu/yona/server/DeviceTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/DeviceTest.groovy
@@ -117,7 +117,7 @@ class DeviceTest extends AbstractAppServiceIntegrationTest
 					assertResponseStatus(it, 400)
 					assert it.responseData.code == "error.device.unknown.operating.system"
 				}, "John", "Doe", "JD",
-				makeMobileNumber(ts), "My Raspberry", "RASPBIAN", Device.SUPPORTED_APP_VERSION)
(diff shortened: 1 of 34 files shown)


appservice/src/intTest/groovy/nu/yona/server/DeviceTest.groovy
```diff
@@ -251,27 +253,29 @@ class DeviceTest extends AbstractAppServiceIntegrationTest
 		def richard = addRichard()
 
 		when:
-		def response = richard.devices[0].postOpenAppEvent(appService, richard.devices[0].operatingSystem, "0.0.1")
+		def appVersion = "9.9.9"
```
aristotelos: I guess this app version is not interpreted at all. This could be reflected in the name of the test: "Try to post app opened event for a too old app version code". A comment could be added here that this app version is ignored.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
```diff
@@ -1892,28 +1880,32 @@ definitions:
...
     - IOS
     - UNKNOWN
+    example: ANDROID
+  DeviceAppVersion:
+    type: string
+    description: The version of the Yona app running on this device. This is the marketing version, a free text string.
+    example: 1.1 build 83
+  DeviceAppVersionCode:
+    type: integer
+    description: The version code of the Yona app running on this device. This is positive a monotonic increassing number, across releases
```
aristotelos: `increassing` -> `increasing`

aristotelos: `positive a monotonic increasing number` -> `a positive, monotonic increasing number`

core/src/main/java/nu/yona/server/device/service/DeviceService.java
```diff
@@ -48,7 +44,12 @@
 @Service
 public class DeviceService
 {
-	private static final String ANDROID_MIN_APP_VERSION = "1.0.0";
+	// When updating the minimum version, be sure to update the version code and the version.
+	// Minimal technical version codes, used to verify the minimum
+	private static final int ANDROID_MIN_APP_VERSION_CODE = 5;
```
aristotelos: Are you sure nobody is using a lower version?

appservice/src/intTest/groovy/nu/yona/server/DeviceTest.groovy
Bert-R: The version code is the deciding factor, but we don't communicate that to the user. That's why this is put in a variable and then used once in the request and then to assert the error message.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
Bert-R: Done.

Bert-R: Done

core/src/main/java/nu/yona/server/device/service/DeviceService.java
Bert-R: Yes. The current released version is 39, so we can be confident that nobody uses 5 or earlier. In case someone is, then it would be right to given an error.

However... even version code 39 does not send the open app event. That's only recently added to the Android app, as part of 1.1.

appservice/src/main/resources/static/swagger/swagger-spec.yaml
Bert-R: Corrected this line in #494

PR #495, opened 2018-08-28
Templating nodeport

kti-nicole:
To make multiple installs on the same cluster possible


Bert-R:
The build failed on it, see [this.](https://yonadev.ci.cloudbees.com/job/build-and-test/job/master/850/console).

I tried resolving it through this commit: 1b52bac

Bert-R:
That doesn't fix it, see [this](https://yonadev.ci.cloudbees.com/job/build-and-test/job/master/851/console).

PR #503, opened 2018-09-21
YD-593 Throw specific exception when user anon not found

Bert-R:
Currently, it just performs a get on the Optional, causing an error with no user anonymized ID. The current patch addresses that.

My suspicion of the issue is that a user deleted their account, causing the user anonymized to be deleted from the database but not from the cache. With improved error handling, we should be able to see that.


Bert-R:
I'm going to merge this simple error logging improvement myself, so we get insight in the problem when it occurs again.

PR #507, opened 2018-10-01
YD-595 Assert device info included POST of user

Bert-R:
If the device name is given, all other data must be provided too. If the device name is not provided, the other information cannot be provided either.

### DIFF
--- a/appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
+++ b/appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
@@ -242,10 +242,10 @@ public class UserController extends ControllerBase
 		{
 			String hint = "If the device name is provided, the other properties must be present too";
 			Require.isNonNull(postPutUser.deviceOperatingSystemStr,
-					() -> InvalidDataException.extraProperty("deviceOperatingSystem", hint));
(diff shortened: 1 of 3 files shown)


appservice/src/main/java/nu/yona/server/subscriptions/rest/UserController.java
```diff
@@ -226,10 +227,35 @@ private String determineRequestingUserId(UUID userId, String requestingUserIdStr
...
+					() -> InvalidDataException.extraProperty("deviceOperatingSystem", hint));
+			Require.isNull(postPutUser.deviceAppVersion, () -> InvalidDataException.extraProperty("deviceAppVersion", hint));
+			Require.isNull(postPutUser.deviceAppVersionCode,
+					() -> InvalidDataException.extraProperty("deviceAppVersionCode", hint));
+		}
+		else
+		{
+			String hint = "If the device name is provided, the other properties must be present too";
+			Require.isNonNull(postPutUser.deviceOperatingSystemStr,
+					() -> InvalidDataException.extraProperty("deviceOperatingSystem", hint));
```
aristotelos: Shouldn't this be `missingProperty` instead of `extraProperty`?

Bert-R: You're totally right. It's a copy/paste error. I've pushed an update to correct this.

PR #511, opened 2018-10-18
YD-596 Include load test in pipeline

Bert-R:
The load test is now an extra (optional) stage.


Jenkinsfile
```diff
@@ -115,6 +115,45 @@ pipeline {
...
+				environment name: 'RUN_LOAD_TEST', value: 'yes'
+			}
+			environment {
+				JM_PATH_IN_CONT = "/mnt/jmeter"
+				JM_LOCAL_PATH = "jmeter"
+				JM_THREADS = "100"
+				JM_LOAD_DURATION = "600"
+			}
+			steps {
+				checkout scm
```
aristotelos: Doesn't make this the step slower than needed?

Jenkinsfile
```diff
@@ -115,6 +115,45 @@ pipeline {
...
+			steps {
+				slackSend color: 'good', channel: '#devops', message: "Server build ${env.BUILD_NUMBER} ready to start load test"
+				script {
+					env.RUN_LOAD_TEST = input message: 'User input required',
+					submitter: 'authenticated',
+					parameters: [choice(name: 'Run load test', choices: 'no\nyes', description: 'Choose "yes" if you want to run the load test')]
+				}
+			}
+		}
+		stage('Run load test') {
```
aristotelos: So for my understanding, this step doesn't run with the load test server created in the stage before, but it runs on a 'yona' build agent?

PR #514, opened 2018-10-23
YD-599 Load top-100 apps

Bert-R:
Load the activity categories as provided by Cornelis:
* Gambling not longer a mandatory goal
* Some name and description changes
* A lot more applications categorized


dbinit/data/productionActivityCategories.json
```diff
@@ -155,7 +594,7 @@
       "id": "27395d17-7022-4f71-9daf-f431ff4f11e8",
       "localizableName": {
          "en-US": "Social",
-         "nl-NL": "Social media"
+         "nl-NL": "Sociale Netwerken"
```
aristotelos: Should this be with two capital letters?

Bert-R: I had the same question but didn't raise it to Cornelis. Now I did and he agreed to make it lowercase.

PR #516, opened 2018-10-30
YD-600 Spread index calculation considers DST change

Bert-R:
On the date a zone enters or leaves daylight saving time, the day has 23 or 25 hours. This can give issues in the spread calculation. When leaving DST (i.e. on a 25-hour day), the calculated number of minutes for a time stamp after the transition needs to be corrected by subtracting 60 minutes. Right after transitioning into DST, we subtract -60 minutes.

### DIFF
--- a/core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
+++ b/core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
@@ -193,6 +193,11 @@ public class DayActivity extends IntervalActivity
 	private static int determineDstCorrectionMinutes(ZonedDateTime time)
 	{
 		ZoneOffsetTransition previousTransition = time.getZone().getRules().previousTransition(time.toInstant());
+		if (previousTransition == null)
+		{
+			// Never did a transition in this zone.
+			return 0;
+		}
 		ZonedDateTime transitionDateTime = previousTransition.getInstant().atZone(time.getZone());
 		if (transitionDateTime.toLocalDate().equals(time.toLocalDate()) && transitionDateTime.isBefore(time))
 		{
--- a/core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
+++ b/core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
@@ -186,18 +186,20 @@ public class DayActivity extends IntervalActivity
 
 	private int getSpreadIndex(ZonedDateTime atTime)
 	{
-		int dstCorrection = determineDstCorrection(atTime);
+		int dstCorrection = determineDstCorrectionMinutes(atTime);
 		return (int) ((Duration.between(getStartTime(), atTime).toMinutes() - dstCorrection) / 15);
 	}
 
-	private static int determineDstCorrection(ZonedDateTime time)
+	private static int determineDstCorrectionMinutes(ZonedDateTime time)
 	{
 		ZoneOffsetTransition previousTransition = time.getZone().getRules().previousTransition(time.toInstant());
 		ZonedDateTime transitionDateTime = previousTransition.getInstant().atZone(time.getZone());
 		if (transitionDateTime.toLocalDate().equals(time.toLocalDate()) && transitionDateTime.isBefore(time))
 		{
 			// Transition happened earlier today
-			return previousTransition.isOverlap() ? 60 : -60;
+			ZoneOffset offsetBefore = previousTransition.getOffsetBefore();
+			ZoneOffset offsetAfter = previousTransition.getOffsetAfter();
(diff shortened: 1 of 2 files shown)


aristotelos:
Isn't it easier to just round down if the spread index is greater than 96 and round up if the spread index is less than 0? In theory, there could also be days with 26 hours.

Bert-R:
You're right that it's not smart to assume that the time difference is always one hour. At the other hand, it's not a matter of ignoring "excess minutes": the activities are also wrongly positioned in the spread. That's where the unit test failed on.

I've pushed an update that simply derives the correction from the offset differences.

core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
```diff
@@ -185,7 +186,22 @@ else if (spreadItemIndex == spreadEndIndex)
...
 	{
-		return (int) (Duration.between(getStartTime(), atTime).toMinutes() / 15);
+		int dstCorrection = determineDstCorrectionMinutes(atTime);
+		return (int) ((Duration.between(getStartTime(), atTime).toMinutes() - dstCorrection) / 15);
+	}
+
+	private static int determineDstCorrectionMinutes(ZonedDateTime time)
+	{
+		ZoneOffsetTransition previousTransition = time.getZone().getRules().previousTransition(time.toInstant());
+		ZonedDateTime transitionDateTime = previousTransition.getInstant().atZone(time.getZone());
```
aristotelos: Is `previousTransition` never `null`?

core/src/test/java/nu/yona/server/analysis/entities/DayActivityTest.java
```diff
@@ -46,6 +55,52 @@ private DayActivity createDayActivity(WeekActivity w, int plusDays)
...
+	public void getSpread_afterTransitionIntoDst_negativeCorrectionApplied()
+	{
+		DayActivity d = createDayActivity(ZonedDateTime.of(2018, 3, 25, 0, 0, 0, 0, testZone));
+		addActivity(d, "03:05:00", "03:07:00");
+
+		assertSpreadItems(d.getSpread(), "11=0,12=2,13=0");
+	}
+
+	@Test
+	public void getSpread_beforeTransitionOutOfDst_noCorrectionApplied()
```
aristotelos: How about adding a test `getSpread_duringTransitionOutOfDst_...`?

Bert-R: That cannot occur. Transit out of DST causes an overlap, not a gap.

Bert-R:
Pushed an update

core/src/test/java/nu/yona/server/analysis/entities/DayActivityTest.java
aristotelos: What I meant is during the overlapping period.

Bert-R: The users might be doing activities between 2:00 and 3:00 twice, potentially causing overlapping activities. This is not special for us and we have a test already (``getSpreadGetTotalActivityDurationMinutes_multipleActivitiesOverlappingUnsorted_countsOverlappingActivitiesOnce``).

core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
```diff
@@ -185,7 +186,27 @@ else if (spreadItemIndex == spreadEndIndex)
 
 	private int getSpreadIndex(ZonedDateTime atTime)
 	{
-		return (int) (Duration.between(getStartTime(), atTime).toMinutes() / 15);
+		int dstCorrection = determineDstCorrectionMinutes(atTime);
```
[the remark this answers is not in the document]
Bert-R: Yes, that works, see this:
```
		ZonedDateTime dateTime = ZonedDateTime.of(2018, 10, 28, 0, 0, 0, 0, ZoneId.of("Europe/Amsterdam"));

		for (int i = 0; (i < 25); i++)
		{
			System.out.println(dateTime.plusHours(i).toLocalTime().getHour());
		}
```
This outputs:

```
0
1
2
2
3
4
5
6
7
8
9
10
11
12
13
14
15
16
17
18
19
20
21
22
23
```
Thanks! I'll update the code.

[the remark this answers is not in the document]
Bert-R: We need to convert to ``LocalDateTime`` rather than ``LocalTime``, because otherwise we can't see the difference anymore between start of the current day and the next one. But with that it works and is easier to understand than my earlier solution.

Bert-R:
Pushed an update

PR #519, opened 2018-11-05
YD-590 Register caches explicitly

Bert-R:
The Hazelcast caches don't update metrics unless explicitly registered.

### DIFF
--- a/core/src/main/java/nu/yona/server/CacheConfiguration.java
+++ b/core/src/main/java/nu/yona/server/CacheConfiguration.java
@@ -1,0 +1,80 @@
+package nu.yona.server;
+/*******************************************************************************
+ * Copyright (c) 2018 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License, v.
+ * 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *******************************************************************************/
+
+import java.io.IOException;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.actuate.metrics.cache.CacheMetricsRegistrar;
+import org.springframework.cache.Cache;
+import org.springframework.cache.CacheManager;
+import org.springframework.cache.annotation.CachingConfigurerSupport;
+import org.springframework.cache.concurrent.ConcurrentMapCacheManager;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import com.hazelcast.client.HazelcastClient;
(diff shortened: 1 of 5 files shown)


core/src/main/java/nu/yona/server/CacheConfiguration.java
```diff
@@ -0,0 +1,80 @@
+package nu.yona.server;
```
aristotelos: We usually put the copyright comment above the `package` line

core/src/main/java/nu/yona/server/goals/service/ActivityCategoryService.java
```diff
@@ -63,19 +65,35 @@
...
 	@Autowired
 	private ActivityCategoryRepository repository;
 
+	@Autowired(required = false)
+	private CacheConfiguration cacheConfiguration;
+
 	private static final Logger logger = LoggerFactory.getLogger(ActivityCategoryService.class);
 
+	@PostConstruct
+	public void register()
```
aristotelos: Suggestion: `registerCacheForMetrics`

core/src/main/java/nu/yona/server/goals/service/ActivityCategoryService.java
```diff
@@ -63,19 +65,35 @@
...
+	}
+
 	@Transactional
 	public ActivityCategoryDto getActivityCategory(UUID id)
 	{
 		ActivityCategory activityCategoryEntity = getEntityById(id);
 		return ActivityCategoryDto.createInstance(activityCategoryEntity);
 	}
 
-	@Cacheable(value = "activityCategorySet", key = "'instance'")
```
aristotelos: This fixes a bug I assume?

core/src/main/java/nu/yona/server/CacheConfiguration.java
Bert-R: True. I now corrected ``CoreConfiguration`` too.

core/src/main/java/nu/yona/server/goals/service/ActivityCategoryService.java
Bert-R: Makes sense

core/src/main/java/nu/yona/server/goals/service/ActivityCategoryService.java
Bert-R: I'm not sure whether I would call it a bug, but the ``@CacheConfig`` annotation named the cache ``activityCategories`` while this annotation here named it ``activityCategorySet``, and the last one won. Now the cache is named ``activityCategories``.

Bert-R:
Pushed an update

PR #523, opened 2018-11-20
YD-605 Send SMS with direct message

aristotelos:
Sending an SMS with direct messages unless the user was created on buddy request.

### DIFF
--- a/core/src/test/java/nu/yona/server/messaging/service/MessageServiceTest.java
+++ b/core/src/test/java/nu/yona/server/messaging/service/MessageServiceTest.java
@@ -50,6 +50,8 @@ import nu.yona.server.messaging.entities.MessageDestinationRepository;
 import nu.yona.server.messaging.entities.MessageRepository;
 import nu.yona.server.messaging.entities.MessageSource;
 import nu.yona.server.messaging.entities.MessageSourceRepository;
+import nu.yona.server.sms.SmsService;
+import nu.yona.server.sms.SmsTemplate;
 import nu.yona.server.subscriptions.entities.User;
 import nu.yona.server.subscriptions.entities.UserAnonymized;
 import nu.yona.server.subscriptions.service.UserAnonymizedDto;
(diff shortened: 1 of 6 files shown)


core/src/test/java/nu/yona/server/messaging/service/MessageServiceTest.java
```diff
@@ -167,18 +173,34 @@ public void sendMessage_anonymousDestination_sendsFirebaseMessage()
...
 		MessageDestination namedMessageDestination = Mockito.mock(MessageDestination.class);
+		String mobileNumber = "+316123456789";
+		User user = Mockito.mock(User.class);
+		when(user.getNamedMessageDestination()).thenReturn(namedMessageDestination);
+		when(user.getMobileNumber()).thenReturn(mobileNumber);
+
+		service.sendDirectMessage(message, user);
 
+		verify(mockFirebaseService, never()).sendMessage(any(), any());
+		verify(mockSmsService, times(1)).send(mobileNumber, SmsTemplate.DIRECT_MESSAGE_NOTIFICATION, new HashMap<>());
```
Bert-R: Use ``Collection.emptyMap()`` here too.

core/src/test/java/nu/yona/server/messaging/service/MessageServiceTest.java
```diff
@@ -167,18 +173,34 @@ public void sendMessage_anonymousDestination_sendsFirebaseMessage()
...
-
 		MessageDestination namedMessageDestination = Mockito.mock(MessageDestination.class);
+		String mobileNumber = "+316123456789";
+		User user = Mockito.mock(User.class);
+		when(user.getNamedMessageDestination()).thenReturn(namedMessageDestination);
+		when(user.getMobileNumber()).thenReturn(mobileNumber);
+
+		service.sendDirectMessage(message, user);
 
+		verify(mockFirebaseService, never()).sendMessage(any(), any());
```
Bert-R: If it's worth verifying that direct messages don't push through Firebase, I think we should also verify no SMS is sent when pushing through Firebase, because these two are mutually exclusive. However, I feel we don't need to verify that. It's simply impossible to make a mistake: for anonymous messages, we only have the device ID and no mobile number. For direct messages, we only have the mobile number, no device ID.

aristotelos: OK, removed this assertion

PR #532, opened 2019-01-24
YD-617 Warn when less than 100 users can still join

Bert-R:
Instead of "less than 10% of the current maximum number of users"


Bert-R:
Merging this myself. It's a very small change and I'd like to see whether Jenkins still is fine after the big plug-in upgrade.

PR #534, opened 2019-01-29
YD-619 Simplify Buddy DTO

Bert-R:
* A buddy always has a user
* The BuddyDto constructor now asserts that
* The BuddyResource now relies on that

The Swagger spec already defined it as mandatory.


Bert-R:
The reason to make this explicit in the code: the Android app was having similar checks as the ``BuddyResource``

PR #537, opened 2019-01-31
YD-621 Added extra assert to verify proxy settings

Bert-R:


appservice/src/intTest/groovy/nu/yona/server/UserTest.groovy
```diff
@@ -29,6 +29,9 @@ class UserTest extends AbstractAppServiceIntegrationTest
 		then:
 		testUser(john, true, false, ts)
 		def baseUserUrl = YonaServer.stripQueryString(john.url)
+		// The below asserts checks returned URLs match the request URL. This is important when going through a proxy
```
aristotelos: Wouldn't it be more logical to write a separate test case for this?

Bert-R: I personally feel this is the right place. This test validates a couple of URLs and this is one of them. It's not our primary objective to test this kind of functionality anywhere, as it's implemented in Spring and I expect them to test this. I've added this as a safeguard.

PR #538, opened 2019-02-13
YD-618 Upgraded to Java 11

Bert-R:
With everything that's required for it:
* Upgrade Gradle
* Upgrade Docker plug-in in Gradle
* Upgrade Groovy and Spock
* Upgrade equalsverifier
* Upgrade Liquibase
* Make some small code tweaks
* Update the Dockerfiles
* Upgrade the build.gradle files for the Docker plug-in changes


core/src/test/java/nu/yona/server/util/TimeUtilTest.java
```diff
@@ -19,24 +19,24 @@
 	public void getEndOfDay_anyTime_returnsTwelveOClockInGivenZone()
 	{
 		ZonedDateTime result = TimeUtil.getEndOfDay(ZoneId.of("Europe/Amsterdam"),
-				ZonedDateTime.parse("2017-08-22T20:44:11+01:00[Europe/Paris]"));
-		assertThat(result, equalTo(ZonedDateTime.parse("2017-08-23T00:00:00+01:00[Europe/Amsterdam]")));
+				ZonedDateTime.parse("2017-08-22T20:44:11+02:00[Europe/Paris]"));
```
Bert-R: Given this is in the summer, the offset is 2 hours. This was always wrong but never noticed.

dbinit/build.gradle
```diff
@@ -34,46 +32,22 @@ docker {
...
 	}
 }
 
 task pushDockerImage(type: DockerPushImage, dependsOn: buildDockerImage) {
-	imageName = "${project.group}/yona-mariadb-liquibase-update"
+	imageName = dockerImageName
 	tag = project.ext.docker_tag
 }
-
-task pathingJar(type: Jar) {
```
aristotelos: The pathingJar code was added by me because the dbinit project could not start because of too many dependencies in its class path. Is this now handled differently by the new Gradle version or Java version?

Bert-R: It's worth trying. You might want to create a task ticket to see whether it's still necessary.

Bert-R: I misunderstood this comment. I read it as a suggestion to remove it, but actually I removed it already.
It turns out we need to have this. It started failing on me today. I'll push it directly to the master.

PR #541, opened 2019-03-05
YD-607 Use longer interval and ensure proper baseline

aristotelos:
- Use interval of 20 seconds because [Hazelcast batch invalidation frequency is by default 10 seconds](https://docs.hazelcast.org/docs/latest/manual/html-single/index.html#near-cache-invalidation)
- Ensure a proper baseline because the number from app service cannot be trusted (unless we add a `waitForCachePropagation` in the `given` part too)


aristotelos:
I made a quick change here, haven't yet run the integration test locally because I do not have it running on my new laptop yet

PR #543, opened 2019-03-21
YD-624 Upgrade to JUnit 5

Bert-R:


analysisservice/src/test/java/nu/yona/server/analysis/service/AnalysisEngineService_determineRelevantGoalsAndroidTest.java
```diff
@@ -0,0 +1,32 @@
...
+public class AnalysisEngineService_determineRelevantGoalsAndroidTest extends AnalysisEngineService_determineRelevantGoalsTest
+{
+	@Override
+	protected OperatingSystem getOperatingSystem()
+	{
+		return OperatingSystem.ANDROID;
+	}
+
+	@Test
+	public void determineRelevantGoals_networkActivityOneOptionalUserGoal_userGoalOnIOs()
```
aristotelos: Why is the operating system relevant in this test?

By the way, if it is relevant, it should read `determineRelevantGoals_networkActivityOneOptionalUserGoalOnAndroid_returnsEmpty`. Notice that it returned empty in the previous test?

core/src/test/java/nu/yona/server/crypto/seckey/CryptoSessionTest.java
```diff
@@ -83,46 +80,49 @@ public void decrypt_validPassword_returnsDecryptedData(String plaintext)
...
+		});
 	}
 
 	@Test
 	public void start_passwordCheckerSaysOk_doesNotThrow()
 	{
 		try (CryptoSession cryptoSession = CryptoSession.start(Optional.of(PASSWORD1), CryptoSessionTest::passwordIsOk))
 		{
-			assertTrue(true);
+			assertThat("Password checker returned OK", true);
```
aristotelos: This is a really strange use of `assertThat`. I think it should be `org.junit.jupiter.api.Assertions.assertDoesNotThrow;`

core/src/test/java/nu/yona/server/crypto/seckey/CryptoSessionTest.java
```diff
@@ -83,46 +80,49 @@ public void decrypt_validPassword_returnsDecryptedData(String plaintext)
...
 	public void start_passwordCheckerSaysNotOk_throws()
 	{
-		try (CryptoSession cryptoSession = CryptoSession.start(Optional.of(PASSWORD1), CryptoSessionTest::passwordIsNotOk))
-		{
-			assertTrue(false);
-		}
+		assertThrows(CryptoException.class, () -> {
+			try (CryptoSession cryptoSession = CryptoSession.start(Optional.of(PASSWORD1), CryptoSessionTest::passwordIsNotOk))
+			{
+				assertThat("Password checker inadvertently returned OK", false);
```
aristotelos: We should not use `assertThat` here, but `fail("Password checker inadvertently returned OK")` or just leave it away

core/src/test/java/nu/yona/server/subscriptions/service/BuddyDeviceChangeMessageDtoTest.java
```diff
@@ -193,7 +189,7 @@ public void managerHandleAction_deviceAdded_deviceExistsForBuddy()
 		manager.handleAction(UserDto.createInstance(bob), messageEntity, "process", null);
 
 		// Assert success
-		assertTrue(messageEntity.isProcessed());
+		assertThat("Message is not processed", messageEntity.isProcessed());
```
aristotelos: This is not correct, should be `assertTrue(messageEntity.isProcessed(), () -> "Message is not processed")` or `assertThat(messageEntity.isProcessed(), is(true), () -> "Message is not processed")`, same for other places in this file

core/src/test/java/nu/yona/server/test/util/InCryptoSession.java
```diff
@@ -0,0 +1,22 @@
...
+
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@ExtendWith(InCryptoSessionExtension.class)
+public @interface InCryptoSession
+{
+	String value();
```
aristotelos: `value` -> `password`

Bert-R:
> I also stumbled upon the fact that many asserts use `assertThat(x, equalTo(y))` while I think it should be `assertThat(x, is(equalTo(y)))`. See also the [JUnit user guide](https://junit.org/junit5/docs/snapshot/user-guide/#writing-tests-assertions). We can correct that in another PR.

I don't think it's worth the effort. See [the description of ``is``](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/core/Is.html). It's just "allowing tests to be slightly more expressive".

analysisservice/src/test/java/nu/yona/server/analysis/service/AnalysisEngineService_determineRelevantGoalsAndroidTest.java
Bert-R: Good find! The test was wrong. On Android, the set of goals should be empty, as we are not measuring time based on network traffic there.

Even worse:: these tests didn't run at all because ``useJUnitPlatform`` wasn't added for the ``analysisengine`` project.

core/src/test/java/nu/yona/server/crypto/seckey/CryptoSessionTest.java
Bert-R: You're right. I'll improve it.

core/src/test/java/nu/yona/server/crypto/seckey/CryptoSessionTest.java
Bert-R: The assertion is important, so we shouldn't omit it. However, ``fail`` is much better than ``assertThat``.

core/src/test/java/nu/yona/server/test/util/InCryptoSession.java
Bert-R: ``value`` is the name for the default value of an assertion.

core/src/test/java/nu/yona/server/subscriptions/service/BuddyDeviceChangeMessageDtoTest.java
Bert-R: Why is this wrong? It seems in line with [the documentation](http://hamcrest.org/JavaHamcrest/javadoc/2.1/).

Bert-R:
Pushed an update

core/src/test/java/nu/yona/server/crypto/seckey/CryptoSessionTest.java
aristotelos: Why shouldn't we use `assertDoesNotThrow` here?

core/src/test/java/nu/yona/server/subscriptions/service/BuddyDeviceChangeMessageDtoTest.java
aristotelos: I don't see any example like in the code above in the [MatcherAssert doc](http://hamcrest.org/JavaHamcrest/javadoc/2.1/org/hamcrest/MatcherAssert.html). In the [Matchers doc](http://hamcrest.org/JavaHamcrest/javadoc/2.1/org/hamcrest/Matchers.html) and the [JUnit 5 user guide](https://junit.org/junit5/docs/current/user-guide/) there are many examples but not like the code above.

core/src/test/java/nu/yona/server/crypto/seckey/CryptoSessionTest.java
Bert-R: Tests always fail when they throw an exception. ``assertDoesNotThrow`` is meant for special cases only. That's why I opted for explicitly asserting we passed through the try-block. I feel that expresses more explicitly what we want to accomplish.

core/src/test/java/nu/yona/server/subscriptions/service/BuddyDeviceChangeMessageDtoTest.java
Bert-R: I don't see examples indeed, but it's still valid code. See also [Stackoverflow here](https://stackoverflow.com/questions/23398736/custom-error-message-for-assertthat-in-junit).

aristotelos: Let's stick to the manual :). The stackoverflow example has 3 arguments, not 2.

core/src/test/java/nu/yona/server/crypto/seckey/CryptoSessionTest.java
aristotelos: Actually the thing we want to assert is that some code does not throw an exception, that is also what the method name says. `assertDoesNotThrow` is exactly meant for this purpose.

core/src/test/java/nu/yona/server/subscriptions/service/BuddyDeviceChangeMessageDtoTest.java
Bert-R: The Stackoverflow example uses the second of [these methods](http://hamcrest.org/JavaHamcrest/javadoc/2.1/org/hamcrest/MatcherAssert.html#MatcherAssert--) while I'm using the first one. Why is using the first one wrong?

core/src/test/java/nu/yona/server/crypto/seckey/CryptoSessionTest.java
Bert-R: ``assertDoesNotThrow`` is designed to put an assert around a specific piece of a test, not an entire test, see [the documentation](https://junit.org/junit5/docs/5.3.0/api/org/junit/jupiter/api/Assertions.html#assertDoesNotThrow(org.junit.jupiter.api.function.Executable)). I'll remove the assert and add a comment.

core/src/test/java/nu/yona/server/subscriptions/service/BuddyDeviceChangeMessageDtoTest.java
Bert-R: I still don't really get your concern here. The ``assertThat`` API that I'm using here is a valid API, not deprecated and it works properly. To prove that, I've added an exclamation mark in front of ``messageEntity.isProcessed()``, and that results in a proper test failure:
```
java.lang.AssertionError: Message is not processed
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:26)
	at nu.yona.server.subscriptions.service.BuddyDeviceChangeMessageDtoTest.managerHandleAction_deviceAdded_deviceExistsForBuddy(BuddyDeviceChangeMessageDtoTest.java:192)
```

The ``assertThat`` operation is from Hamcrest, and we use it > 300 times. The ``assertTrue`` operation is from JUnit and we are not using it anymore after this refactoring. You are right that the one you refer to is an alternative, and there is even a simpler one. We could write the above statement this way:
```
assertTrue(messageEntity.isProcessed(), "Message is not processed");
```
If I make that fail like above, you get this:
```
org.opentest4j.AssertionFailedError: Message is not processed ==> expected: <true> but was: <false>
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:38)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:40)
	at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:168)
	at nu.yona.server.subscriptions.service.BuddyDeviceChangeMessageDtoTest.managerHandleAction_deviceAdded_deviceExistsForBuddy(BuddyDeviceChangeMessageDtoTest.java:193)
```
This would be our first use of ``assertTrue`` and it would not be any better than the statement I wrote.

aristotelos: OK, you're right it works, I just have never seen `assertThat` used in that way before, and it is probably not that common.

core/src/test/java/nu/yona/server/crypto/seckey/CryptoSessionTest.java
aristotelos: My opinion would be that every test should have an assert and as the test ends with `_doesNotThrow` it is logical to use `assertDoesNotThrow`. Something like:
```
CryptoSession cryptoSession;

assertDoesNotThrow(() -> { cryptoSession = CryptoSession.start(Optional.of(PASSWORD1), CryptoSessionTest::passwordIsOk) })

cryptoSession.close();
```

But as the code works I'll close this conversation.

PR #544, opened 2019-04-05
YD-625 Uncoupled MessageController from specific messages

Bert-R:
The MessageController class used to contain specific logic for basically all message types. This implies that creating a new message or considerably changing the implementation of one implies having to change MessageController too. This coupling was too tight.

Instead, we now have the concept of a decorator for messages. A single message instance can be decorated (with embedded entities or links) by multiple decorators that apply to different classes in the type hierarchy of that instance.

Along with this:
* Explicitly asserted the set of links for each and every message type
* Fixed ActivityCommentTest. It often didn't provide the replied message, even though the message being tested was a reply. The old asserts didn't notice that, but the new ones did
* DeviceChangeMessage: added "yona:buddy" link, as already described in the Swagger spec
* DisclosureRequestMessage and DisclosureResponseMessage: added "yona:buddy" link to entity and to the Swagger spec
* BuddyVpnConnectionStatusChangeMessage: Added "yona:buddy" link and added the type to the Swagger spec


appservice/src/main/java/nu/yona/server/messaging/rest/GoalConflictMessageDecorator.java
```diff
@@ -0,0 +1,50 @@
...
+	@Override
+	public void decorate(MessageResourceAssembler assembler, MessageDto message)
+	{
+		GoalConflictMessageDto goalConflictMessage = (GoalConflictMessageDto) message;
+		addActivityCategoryLink(assembler, goalConflictMessage);
+		addDayActivityDetailLink(assembler, goalConflictMessage);
+
+	}
+
+	private void addActivityCategoryLink(MessageResourceAssembler assembler, GoalConflictMessageDto message)
```
yonabuild: MAJOR Remove this unused method parameter "assembler". [rule](http://10.127.10.107:9000/coding_rules#rule_key=squid%3AS1172)

appservice/src/main/java/nu/yona/server/messaging/rest/DecoratorRegistry.java
```diff
@@ -0,0 +1,82 @@
...
+	{
+		Decorator existingDecorator = decoratorByType.get(decoratedClass);
+		if (existingDecorator != null)
+		{
+			throw new IllegalStateException("Duplicate decorators for class " + decoratedClass.getName() + ": "
+					+ existingDecorator.getClass().getName() + " and " + decoratorToRegister.getClass().getName());
+		}
+	}
+
+	public <T> Set<Decorator> getDecorators(Class<? extends MessageDto> classToDecorate)
```
yonabuild: MAJOR T is not used in the method. [rule](http://10.127.10.107:9000/coding_rules#rule_key=squid%3AS2326)

appservice/src/main/java/nu/yona/server/messaging/rest/ActivityCommentMessageDecorator.java
```diff
@@ -0,0 +1,41 @@
...
+import nu.yona.server.analysis.service.ActivityCommentMessageDto;
+import nu.yona.server.goals.entities.Goal;
+import nu.yona.server.messaging.rest.MessageController.MessageResourceAssembler;
+import nu.yona.server.messaging.service.MessageDto;
+
+@Decorates(ActivityCommentMessageDto.class)
+public class ActivityCommentMessageDecorator implements Decorator
+{
+	@Override
+	public void decorate(MessageResourceAssembler assembler, MessageDto message)
```
yonabuild: MAJOR This method has 21 lines, which is greater than the 20 lines authorized. Split it into smaller methods. [rule](http://10.127.10.107:9000/coding_rules#rule_key=squid%3AS138)

yonabuild:
SonarQube analysis reported 3 issues
* ![MAJOR][MAJOR] 3 major

appservice/src/main/java/nu/yona/server/messaging/rest/Decorator.java
```diff
@@ -0,0 +1,13 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License, v.
+ * 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *******************************************************************************/
+package nu.yona.server.messaging.rest;
+
+import nu.yona.server.messaging.rest.MessageController.MessageResourceAssembler;
+import nu.yona.server.messaging.service.MessageDto;
+
+public interface Decorator
```
aristotelos: I find the namings in this PR confusing. First, we do not really implement the [decorator design pattern](https://www.baeldung.com/java-decorator-pattern). Second, we do use annotations here to decorate classes which are called decorators too. We decorate the decorators...

Proposal:

- Putting the added links of separate message types in separate classes is good for keeping things simple. 
- Using annotations to get the right class looks nice and gives the possibility to implement it for both abstract classes and concrete classes. However, we could also use a type parameter maybe? So use `interface MessageResourceAssembler<T>`?
- Rename `Decorator` -> `MessageResourceAssembler` to avoid confusion about the name
- Rename `DecoratorRegistry` -> `MessageResourceAssemblerRegistry`

What do you think? Maybe it's best to discuss this offline :)

Bert-R: Interesting. I didn't think of the GoF decorators at all.
The word "decorate" has many meanings. I continued to use the term "decorate" in line with what we already did, as expressed in the method ``doDynamicDecoration``. The decorators decorate the resource with extra links and/or embedded resources.

You also use the term in two different meanings:
* The GoF decorator pattern
* Java annotations

The "decorators" aren't message resource assemblers, so that name would be wrong. The resource is partially assembled already and the "decorator" is a helper for that.

My first attempt as to make the decorators typed (using ``public interface Decorator<T extends MessageDto>``), but that does not add any value, as the registration is consulted in a typeless way. After quite a struggle, I concluded I couldn't make it work with a type parameter.

So I'm open to a different name for "decorator", to prevent confusion with the GoF pattern, but we can't use ``MessageResourceAssembler``.

appservice/src/main/java/nu/yona/server/messaging/rest/ActivityCommentMessageDecorator.java
```diff
@@ -0,0 +1,51 @@
...
+	{
+		ActivityCommentMessageDto activityCommentMessageMessage = (ActivityCommentMessageDto) message;
```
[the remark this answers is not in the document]
Bert-R: Updated.

Bert-R:
I've pushed an update with renames. See whether this helps preventing confusion. Feel free to propose a different name: renaming is very little effort.

PR #545, opened 2019-04-23
YD-607 Stabilize random failing system tests

aristotelos:
Isolated the sync between admin service and app service in one test to speed up tests and stabilize them. Removed workarounds.

To get the server scripts running on my machine, I had to change port numbers because 8081 is in use.


adminservice/src/intTest/groovy/nu/yona/server/ActivityCategoriesTest.groovy
```diff
@@ -16,16 +16,65 @@ import spock.lang.Specification
...
-	static final def cachePropagationTimeoutSeconds = 240
+	static final def cachePropagationTimeoutSeconds = 20
 
 	@Shared
 	def AdminService adminService = new AdminService()
 
 	@Shared
 	def AppService appService = new AppService()
 
+	// NOTE: This test assumes that en-US is the client language
```
Bert-R: Can't you [set the default locale](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Locale.html#setDefault(java.util.Locale)) under ``given:`` and then restore [the original](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Locale.html#getDefault()) under ``cleanup:``?

aristotelos: Wouldn't it be better to pass an `Accept-Language` header with all requests in our acceptance test by default?

Bert-R: That works for most of the tests, but some tests (``LocalizationTest`` and ``PinResetRequestTest``) explicitly set that header, so I'd rather only change those tests that care.

aristotelos: I fixed a couple of tests that failed because of the different client locale, but there are still 54 other tests failing on my Dutch system. For example, some are related to the fact that the sender of a message is expected to be "RQ (me)" but is "RQ (ik)" (`ActivityCommentTest`, `AppActivityTest`, `BasicBuddyTest`, `RemoveUserTest`). Others are related to the fact that a different invite email is sent (`CreateUserOnBuddyRequestTest`). Yet others are related to different messages (`DeviceManagementTest`, `RemoveUserTest`, `UpdateBuddyUserInfoTest`, `UserPhotoTest`, `VpnConnectionStatusEventTest`). Yet others are related to the name of the default device (`OverwriteUserTest`).

So I still think we should make it default in the tests that we use `Accept-Language: en-US` and override that default by specific tests. But preferably in another PR.

Bert-R: I didn't know the impact is that large. That should indeed go in a different PR. Setting ``en-US`` as default for the tests seems the right approach indeed.

The comment line in the code that this GitHub comment is attached to can be removed now, right?

aristotelos: Removed the comment

PR #546, opened 2019-04-27
YD-626 Prevent 500 in invalid date in URL

Bert-R:
Added exception handling. Along with this improved the error logging for invalid UUIDs by including the cause exception.

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2018 Stichting Yona Foundation
+ * Copyright (c) 2015, 2019 Stichting Yona Foundation
  * This Source Code Form is subject to the terms of the Mozilla Public License,
  * v.2.0. If a copy of the MPL was not distributed with this file, You can
  * obtain one at https://mozilla.org/MPL/2.0/.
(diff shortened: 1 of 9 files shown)


core/src/main/java/nu/yona/server/analysis/service/DayActivityDto.java
```diff
@@ -78,7 +80,14 @@ public int getTotalMinutesBeyondGoal()
 
 	public static LocalDate parseDate(String iso8601)
 	{
-		return LocalDate.parse(iso8601, ISO8601_DAY_FORMATTER);
+		try
+		{
+			return LocalDate.parse(iso8601, ISO8601_DAY_FORMATTER);
```
aristotelos: Wouldn't it be good to apply the same pattern at `WeekActivityDto.parseDate`?

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -1267,6 +1266,30 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(richard)
 	}
 
+	def 'Try get raw activities with an invalid date'()
```
aristotelos: Good to have an acceptance test. We could also add a unit test for `DayActivityDto.parseDate`.

core/src/main/java/nu/yona/server/analysis/service/DayActivityDto.java
Bert-R: Certainly! Added it.

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
Bert-R: Good point. Added it.

Bert-R:
Pushed an update

PR #548, opened 2019-05-07
YD-627 Refactored the user service to split in 4

Bert-R:
The user service grew to a very large service (nearly 1000 lines). To make it more manageable, it's now split into 4 parts, related to retrieval, addition, deletion and update. The dependencies are as follows:

* UserDeletionService: not used by the others
* UserAdditionService: not used by the others
* UserUpdateService: used by UserAdditionService
* UserRetrievalService: used by UserAdditionService and UserUpdateService

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/service/UserAddService.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/UserAddService.java
@@ -43,14 +43,14 @@ import nu.yona.server.util.Require;
 import nu.yona.server.util.TimeUtil;
 
 @Service
-public class UserAdditionService
+public class UserAddService
 {
 	private enum UserSignUp
 	{
 		INVITED, FREE
 	}
 
-	private static final Logger logger = LoggerFactory.getLogger(UserAdditionService.class);
+	private static final Logger logger = LoggerFactory.getLogger(UserAddService.class);
 
 	@Autowired(required = false)
 	private UserRepository userRepository;
@@ -68,7 +68,7 @@ public class UserAdditionService
 	private WhiteListedNumberService whiteListedNumberService;
 
 	@Autowired(required = false)
-	private UserRetrievalService userRetrievalService;
+	private UserLookupService userLookupService;
 
 	@Autowired(required = false)
 	private UserUpdateService userUpdateService;
@@ -89,7 +89,7 @@ public class UserAdditionService
 	public UserDto addUser(UserDto user, Optional<String> overwriteUserConfirmationCode)
 	{
 		assert user.getPrivateData().getDevices().orElse(Collections.emptySet()).size() == 1;
-		UserRetrievalService.assertValidUserFields(user, UserService.UserPurpose.USER);
+		UserLookupService.assertValidUserFields(user, UserService.UserPurpose.USER);
 
 		handleExistingUserForMobileNumber(user.getMobileNumber(), overwriteUserConfirmationCode);
 
@@ -97,7 +97,7 @@ public class UserAdditionService
 		addDevicesToEntity(user, userEntity);
 		Optional<ConfirmationCode> confirmationCode = createConfirmationCodeIfNeeded(overwriteUserConfirmationCode, userEntity);
 		userEntity = userRepository.save(userEntity);
-		UserDto userDto = userRetrievalService.createUserDtoWithPrivateData(userEntity);
+		UserDto userDto = userLookupService.createUserDtoWithPrivateData(userEntity);
 		if (confirmationCode.isPresent())
 		{
 			userUpdateService.sendConfirmationCodeTextMessage(userEntity.getMobileNumber(), confirmationCode.get(),
@@ -182,7 +182,7 @@ public class UserAdditionService
 		}
 		else
 		{
-			userRetrievalService.assertUserDoesNotExist(mobileNumber);
+			userLookupService.assertUserDoesNotExist(mobileNumber);
 			assertUserIsAllowed(mobileNumber, UserSignUp.FREE);
 		}
 	}
@@ -215,7 +215,7 @@ public class UserAdditionService
 
 	private void deleteExistingUserToOverwriteIt(String mobileNumber, String userProvidedConfirmationCode)
 	{
-		User existingUserEntity = UserRetrievalService.findUserByMobileNumber(mobileNumber);
+		User existingUserEntity = UserLookupService.findUserByMobileNumber(mobileNumber);
 		ConfirmationCode confirmationCode = existingUserEntity.getOverwriteUserConfirmationCode();
 
 		assertValidConfirmationCode(existingUserEntity, confirmationCode, userProvidedConfirmationCode,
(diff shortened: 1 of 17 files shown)


core/src/main/java/nu/yona/server/subscriptions/service/UserAdditionService.java
```diff
@@ -0,0 +1,275 @@
...
+	{
+		if (overwriteUserConfirmationCode.isPresent())
+		{
+			// no need to confirm again
+			userEntity.markMobileNumberConfirmed();
+			return Optional.empty();
+		}
+
+		Optional<ConfirmationCode> confirmationCode = Optional.of(userUpdateService.createConfirmationCode());
+		userEntity.setMobileNumberConfirmationCode(confirmationCode.get());
```
yonabuild: MAJOR Call "confirmationCode.isPresent()" before accessing the value. [rule](http://10.127.10.107:9000/coding_rules#rule_key=squid%3AS3655)

core/src/main/java/nu/yona/server/subscriptions/service/UserAdditionService.java
```diff
@@ -0,0 +1,275 @@
...
+			return;
+		}
+
+		logger.info("Maximum number of users: {}", maxUsers);
+	}
+
+	@Transactional(dontRollbackOn = UserOverwriteConfirmationException.class)
+	public UserDto addUser(UserDto user, Optional<String> overwriteUserConfirmationCode)
+	{
+		assert user.getPrivateData().getDevices().orElse(Collections.emptySet()).size() == 1;
```
yonabuild: MAJOR Replace this assert with a proper check. [rule](http://10.127.10.107:9000/coding_rules#rule_key=squid%3AS4274)

core/src/main/java/nu/yona/server/subscriptions/service/UserUpdateService.java
```diff
@@ -0,0 +1,315 @@
...
+		return userEncryptedData;
+	}
+
+	@Transactional
+	public void addBuddy(UserDto user, BuddyDto buddy)
+	{
+		Require.that(user != null && user.getId() != null, InvalidDataException::emptyUserId);
+		Require.that(buddy != null && buddy.getId() != null, InvalidDataException::emptyBuddyId);
+
+		updateUser(user.getId(), userEntity -> {
```
yonabuild: MAJOR A "NullPointerException" could be thrown; "user" is nullable here. [rule](http://10.127.10.107:9000/coding_rules#rule_key=squid%3AS2259)

yonabuild:
SonarQube analysis reported 3 issues
* ![MAJOR][MAJOR] 3 major

Bert-R:
@aristotelos What do you think of this approach? It makes the user service comprehensible, while not breaking the pattern of having one service responsible for the behavior around an entity. The user service delegates its behavior to 4 helper services.

yonabuild:
SonarQube analysis reported 2 issues
* ![MAJOR][MAJOR] 2 major

aristotelos:
> @aristotelos What do you think of this approach? It makes the user service comprehensible, while not breaking the pattern of having one service responsible for the behavior around an entity. The user service delegates its behavior to 4 helper services.

I like the approach to split up in smaller functional parts, I would suggest choosing a bit more functional names, e.g.:

- `UserSubscribeService`
- `UserUnsubscribeService`
- `UserProfileUpdateService`
- `UserService`

Bert-R:
The reason to keep the ``UserService`` is that we have that for all things (user anonymized, buddy, message, etc.). The fact that the user service became too large to handle as a single one shouldn't matter to its consumers.

As we talk about adding/removing users rather than subscribing/unsubscribing them, I'd like to stay close to that terminology. What about this?
* ``UserLookupService``
* ``UserAddService``
* ``UserDeleteService``
* ``UserUpdateService``

That's a bit less convoluted than the current names.

aristotelos:
> The reason to keep the `UserService` is that we have that for all things (user anonymized, buddy, message, etc.). The fact that the user service became too large to handle as a single one shouldn't matter to its consumers.
> 
> As we talk about adding/removing users rather than subscribing/unsubscribing them, I'd like to stay close to that terminology. What about this?
> 
> * `UserLookupService`
> * `UserAddService`
> * `UserDeleteService`
> * `UserUpdateService`
> 
> That's a bit less convoluted than the current names.

Looks good to me.

Bert-R:
Pushed an update with this rename.

core/src/main/java/nu/yona/server/subscriptions/service/UserAddService.java
```diff
@@ -0,0 +1,275 @@
...
+			return;
+		}
+
+		logger.info("Maximum number of users: {}", maxUsers);
+	}
+
+	@Transactional(dontRollbackOn = UserOverwriteConfirmationException.class)
+	public UserDto addUser(UserDto user, Optional<String> overwriteUserConfirmationCode)
+	{
+		assert user.getPrivateData().getDevices().orElse(Collections.emptySet()).size() == 1;
```
yonabuild: MAJOR Replace this assert with a proper check. [rule](http://10.127.10.107:9000/coding_rules#rule_key=squid%3AS4274)

core/src/main/java/nu/yona/server/subscriptions/service/UserAddService.java
```diff
@@ -89,15 +89,15 @@ private void onStart()
 	public UserDto addUser(UserDto user, Optional<String> overwriteUserConfirmationCode)
 	{
 		assert user.getPrivateData().getDevices().orElse(Collections.emptySet()).size() == 1;
-		UserRetrievalService.assertValidUserFields(user, UserService.UserPurpose.USER);
+		UserLookupService.assertValidUserFields(user, UserService.UserPurpose.USER);
```
aristotelos: I would say this `assertValidUserFields` should be part of `UserService` or `UserAddService`, not of `UserLookupService`

core/src/main/java/nu/yona/server/subscriptions/service/UserUpdateService.java
```diff
@@ -143,7 +143,7 @@ public User updateUser(UUID id, Consumer<User> updateAction)
 	{
 		// We add a lock here to prevent concurrent user updates.
 		// The lock is per user, so concurrency is not an issue.
-		return userRetrievalService.withLockOnUser(id, user -> {
+		return userLookupService.withLockOnUser(id, user -> {
```
aristotelos: Why is `withLockOnUser` part of `UserLookupService`? I would say it should be part of `UserUpdateService`

core/src/main/java/nu/yona/server/subscriptions/service/UserAddService.java
Bert-R: I wanted to keep the assertions together. As everything depends on the ``UserLookupService``, they ended up there. Alternatively, I could extract another service that has these assertions. What do you think of that?

core/src/main/java/nu/yona/server/subscriptions/service/UserUpdateService.java
Bert-R: For reasons of dependencies. The method is called from the ``UserLookupService``, so moving this method to ``UserUpdateService`` would cause a two-way dependency between these two.

Bert-R:
Pushed an update with the ``UserAssertionService``.

PR #549, opened 2019-05-18
YD-630 Explicitly assert time measurement no-go goals

Bert-R:
Looking closer into the reported issue, I believe the current behavior is the best we can do. If a user defines a no-go goal, we would like to give an estimated size of the violation. The current time measurement approach is the best we have.

I've extended a test to explicitly assert the behavior for no-go goals too.

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
@@ -1599,9 +1599,9 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		where:
 		operatingSystem | budget | expectedValue
 		"ANDROID" | 0 | [ 13 : 4]
-		"ANDROID" | 60 | [ : ]
+		"ANDROID" | 60 | [ : ] // We only track no-go goals on Android, as app activity monitoring provides much better measurements
 		"IOS" | 0 | [13 : 4]
-		"IOS" | 60 | [13 : 4]
+		"IOS" | 60 | [13 : 4] // App activity monitoring is not possible on iOS, so we also track budget goals
 	}
 
 	private def getRawActivityData(User user, relativeDate, goal) {
--- a/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
@@ -1558,7 +1558,7 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		reportNetworkActivity(richard, ["social"], "http://www.facebook.com", "W-1 Mon 03:24")
 
 		def expectedGoalAccomplished = budget > 0
-		def expectedMinutesBeyondGoal = expectedGoalAccomplished ? 0 : expectedValue.size() == 0 ? 0 :  expectedValue.entrySet().iterator().next().value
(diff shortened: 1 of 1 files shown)


appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -1544,25 +1544,28 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
...
 		richard = appService.reloadUser(richard, CommonAssertions.&assertUserGetResponseDetailsWithPrivateDataIgnoreDefaultDevice)
-		Goal timeZoneGoalSocialRichard = richard.findActiveGoal(SOCIAL_ACT_CAT_URL)
+		BudgetGoal budgetGoalSocialRichard = richard.findActiveGoal(SOCIAL_ACT_CAT_URL)
 
 		// Activities on default device
 		reportNetworkActivity(richard, ["social"], "http://www.facebook.com", "W-1 Mon 03:20")
+		reportNetworkActivity(richard, ["social"], "http://www.facebook.com", "W-1 Mon 03:24")
 
+		def expectedGoalAccomplished = budget > 0
+		def expectedMinutesBeyondGoal = expectedGoalAccomplished ? 0 : expectedValue.size() == 0 ? 0 :  expectedValue.entrySet().iterator().next().value
```
aristotelos: I would prefer a simple test parameter instead of this computation

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -1579,24 +1582,26 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
...
 		cleanup:
 		appService.deleteUser(richard)
 
 		where:
-		operatingSystem | expectedValue
-		"ANDROID" | [ : ]
-		"IOS" | [13 : 1]
+		operatingSystem | budget | expectedValue
+		"ANDROID" | 0 | [ 13 : 4]
+		"ANDROID" | 60 | [ : ]
```
aristotelos: A comment on why this is empty and not for iOS could help here

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
Bert-R: That would lead to duplication of information, which I'm not in favor of.
How about rewriting the expression? Is this better?
```
def expectedMinutesBeyondGoal = expectedGoalAccomplished || expectedValue.size() == 0 ? 0 :  expectedValue.entrySet().iterator().next().value
```

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
Bert-R: Pushed an update

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
aristotelos: I still prefer the duplication of information in tests above using code to calculate information. But if you prefer it this way, we keep it this way.

Bert-R: OK. I forgot to push the simplification, but now I've done it. A nested ternary operator is not a nice thing.

PR #553, opened 2019-08-05
YD-637 Disconnect activities before delete device

Bert-R:
Activities have an optional reference to the device on which they occurred. If the user deletes the device, the device reference on the related activities is cleared, but the activities are retained.

Along with this fixed the issue that the cached user anonymized wasn't updated when a device was added or removed.

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/DeviceManagementTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/DeviceManagementTest.groovy
@@ -514,6 +514,34 @@ class DeviceManagementTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(richard)
 	}
 
+	def 'Richard deletes a devices on which activities were done'()
+	{
+		given:
+		def ts = timestamp
+		User richardOnFirstDevice = addRichard()
+		setGoalCreationTime(richardOnFirstDevice, NEWS_ACT_CAT_URL, "W-1 Mon 02:18")
+		Device remainingDevice = richardOnFirstDevice.devices[0]
+		def deletedDeviceName = "Second device"
+		User richardOnSecondDevice = appService.addDevice(richardOnFirstDevice, deletedDeviceName, "ANDROID")
+		reportAppActivity(richardOnSecondDevice, "NU.nl", "W-1 Mon 03:15", "W-1 Mon 03:35")
+
+		Device deviceToDelete = richardOnSecondDevice.devices.find{ YonaServer.stripQueryString(it.url) != YonaServer.stripQueryString(remainingDevice.url) }
+
+		when:
+		def response = appService.deleteResourceWithPassword(deviceToDelete.editUrl, richardOnSecondDevice.password)
+
+		then:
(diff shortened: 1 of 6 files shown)


appservice/src/intTest/groovy/nu/yona/server/DeviceManagementTest.groovy
```diff
@@ -514,6 +514,34 @@ class DeviceManagementTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(richard)
 	}
 
+	def 'Richard deletes a devices on which activities were done'()
```
aristotelos: Why do we want to retain activities from the deleted device? If they have any functional use, should this test also not assert that the activities are still there?

core/src/main/java/nu/yona/server/device/service/DeviceService.java
```diff
@@ -105,6 +105,7 @@ public UserDeviceDto addDeviceToUser(User userEntity, UserDeviceDto deviceDto)
 		UserDevice deviceEntity = userDeviceRepository.save(UserDevice.createInstance(userEntity, deviceDto.getName(),
 				deviceAnonymized.getId(), userService.generatePassword()));
 		userEntity.addDevice(deviceEntity);
+		userAnonymizedService.updateUserAnonymized(userEntity.getAnonymized().getId());
```
aristotelos: Does this line fix another bug? If so, should we add another test?

core/src/main/java/nu/yona/server/device/service/DeviceService.java
```diff
@@ -231,6 +232,7 @@ public void deleteDevice(UUID userId, UUID deviceId)
 			UUID deviceAnonymizedId = deviceEntity.getDeviceAnonymized().getId();
 
 			deleteDeviceWithoutBuddyNotification(userEntity, deviceEntity);
+			userAnonymizedService.updateUserAnonymized(userEntity.getAnonymized().getId());
```
aristotelos: Does this line fix another bug? If so, should we add another test?

core/src/main/java/nu/yona/server/device/service/DeviceService.java
Bert-R: Yes, the test initially failed to post an activity because the device anonymized didn't get synchronized to the analysis engine service. An extra test isn't necessary, as that's covered with this test.

core/src/main/java/nu/yona/server/device/service/DeviceService.java
Bert-R: See above.

appservice/src/intTest/groovy/nu/yona/server/DeviceManagementTest.groovy
Bert-R: Asserting the activities is a good one. I'll add that. 

I wanted to retain the activities because they really happened. Otherwise, it would be possible to wipe out a significant part of past activities by dropping a device, which doesn't sound fair to buddies.

Bert-R:
Pushed an update with the extra asserts

PR #556, opened 2019-09-07
YD-641 Upgrade to Liquibase Hibernate5 3.8

Bert-R:
The new version addresses a number of issues (most notably liquibase/liquibase-hibernate#134), resulting in much cleaner output for diffChangeLog.

The new version proposes to basically drop all default values. I've accepted most of them in changelog-0021-yd-641.yml. Some however are unjustified and cause runtime issues if they would be dropped. These will still have to be removed from a future generated change log. I will update the wiki page accordingly.

The new version now properly handles a unique constraint on a type, so I've put that back in IntervalActivity.java

### DIFF
--- a/core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
+++ b/core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2018 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License,
+ * Copyright (c) 2016, 2019 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License,
  * v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
  *******************************************************************************/
 package nu.yona.server.analysis.entities;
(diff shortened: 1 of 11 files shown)


Bert-R:
This makes generation of a change log quite a bit easier, see [this wiki page update](https://wiki.yona.nu/pages/diffpagesbyversion.action?pageId=16023665&selectedPageVersions=12&selectedPageVersions=13).

core/src/main/java/nu/yona/server/analysis/entities/DayActivity.java
```diff
@@ -44,6 +45,7 @@
 							// so you get 4 goals * 1 day = 4 activity collections to be joined
 	private List<Activity> activities;
 
+	@Column(columnDefinition = "bit default false")
```
aristotelos: Why are those lines exactly necessary? Is that a still remaining bug in Liquibase? Because a boolean is false by default in Java, so I would expect that it is saved as false to the database.

Bert-R: I agree it's a bit weird, but I wouldn't call it a bug. In SQL, a column can be null. In some scenarios (because of our mapping strategy), we do not use this column. In such cases, we need to tell what the column value should be, as otherwise it cannot be read back in a boolean.

aristotelos: I still don't get why it shouldn't be simply `null` in the case that we do not use the column because of our mapping strategy. In that case it should also not be necessary to read it back to a boolean, should it?

Bert-R: Indeed, I had similar questions. We only read it if we write it, so a default value shouldn't matter.

PR #558, opened 2019-09-13
YD-643 Up min app version

Bert-R:
This is in preparation to the new API version that will come with YD-622.

The minimum versions for Android is now:
Code: 218
Version: 1.2

For iOS, it does not matter much, as the iOS app is expired in Test Flight, but to be safe there, we demand the latest version on the iOS master branch:
Code: 166
Version: 1.2


Bert-R:
~50 users still have the old app I stalled. It's quite likely they aren't active users. If they are, they will get a message telling them they need to update.

PR #559, opened 2019-09-18
YD-644 set lastMonitoredActivityDate on device

Bert-R:
The property is now set and the integration tests cover it

### DIFF
--- a/analysisservice/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
+++ b/analysisservice/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2018 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License,
+ * Copyright (c) 2015, 2019 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License,
  * v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
  *******************************************************************************/
 package nu.yona.server.analysis.service;
@@ -26,6 +26,7 @@ import nu.yona.server.analysis.entities.Activity;
 import nu.yona.server.analysis.entities.ActivityRepository;
 import nu.yona.server.analysis.entities.DayActivity;
 import nu.yona.server.analysis.entities.DayActivityRepository;
+import nu.yona.server.device.entities.DeviceAnonymized;
 import nu.yona.server.device.entities.DeviceAnonymized.OperatingSystem;
 import nu.yona.server.device.service.DeviceAnonymizedDto;
 import nu.yona.server.device.service.DeviceService;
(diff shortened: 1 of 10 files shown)


analysisservice/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
```diff
@@ -188,14 +189,22 @@ private void analyzeInsideLock(List<ActivityPayload> payloads, UserAnonymizedDto
...
+	{
+		LocalDate lastActivityEndTime = lastActivityPayload.endTime.toLocalDate();
 		Optional<LocalDate> lastMonitoredActivityDate = userAnonymized.getLastMonitoredActivityDate();
-		if (lastActivityEndTime.isPresent()
-				&& lastMonitoredActivityDate.map(d -> d.isBefore(lastActivityEndTime.get())).orElse(true))
+		if (lastMonitoredActivityDate.map(d -> d.isBefore(lastActivityEndTime)).orElse(true))
 		{
-			activityUpdateService.updateLastMonitoredActivityDate(userAnonymizedEntityHolder.getEntity(),
-					lastActivityEndTime.get());
+			DeviceAnonymized deviceAnonymized = deviceService.getDeviceAnonymizedEntity(userAnonymized.getId(),
```
aristotelos: Isn't it cheaper to retrieve the `DeviceAnonymized` entity from the `UserAnonymized` entity?

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -1450,10 +1450,11 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		reportAppActivity(richardIphone, "NU.nl", "W-1 Mon 04:10", "W-1 Mon 04:20") // Partial overlap
 		reportAppActivity(richardIphone, "NU.nl", "W-1 Mon 05:20", "W-1 Mon 05:30") // Full overlap
 		reportAppActivity(richardIphone, "NU.nl", "W-1 Mon 06:15", "W-1 Mon 06:35") // No overlap at all
+		reportAppActivity(richardIphone, "NU.nl", "W-1 Tue 07:53", "W-1 Tue 08:11") // Next day, so different last monitored activity dates for the devices
```
aristotelos: I would prefer a separate set of tests for this feature, instead of combining it with already complex existing tests

analysisservice/src/main/java/nu/yona/server/analysis/service/AnalysisEngineService.java
Bert-R: The ``userAnonymized`` is a DTO, not an entity, so we cannot get the ``DeviceAnonymized`` entity from it.

appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
Bert-R: To me, that wouldn't be right. This test verifies the behavior of a user working on two devices. That test should have asserted the ``lastMonitoredActivityDate`` from the first version, but we forgot to do so. Now I'm correcting the test and the code.

PR #562, opened 2019-10-15
YD-650 Temp passwords are now URL-safe

Bert-R:
Added a test to verify this


Bert-R:
I don't disagree, but looking at the effort of both solution options, I've taken this simple route, as it doesn't harm.

PR #567, opened 2019-10-30
YD-656 Fixed buddy not found by ID issue

Bert-R:
It used to fail in this scenario:
* Richard sends a buddy connect request to Bob
* Bob accepts the request
* Bob overwrites his account before Richard gets the chance to process the buddy connect response message
* Richard now gets a failure on everything because the pending buddy connect response message cannot be processed

The fix is to delete any pending buddy connect response messages of buddies that were deleted while the user was offline.

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
@@ -355,24 +355,44 @@ public class BuddyService
 		if (buddy.getSendingStatus() == Status.ACCEPTED || buddy.getReceivingStatus() == Status.ACCEPTED)
 		{
 			// Buddy request was accepted
-			removeMessagesSentByBuddy(user, buddy);
-			removeMessagesSentByUserToBuddy(user, buddy);
-
-			// Send message to "self", to notify the user about the removed buddy user
-			UUID buddyUserAnonymizedId = getUserAnonymizedIdForBuddy(buddy);
-			sendDropBuddyMessage(BuddyInfoParameters.createInstance(buddy, buddyUserAnonymizedId), Optional.empty(),
-					DropBuddyReason.USER_ACCOUNT_DELETED, user);
+			prepareCleanupRemovedBuddyInAcceptedState(user, buddy);
 		}
 		else if (buddy.getSendingStatus() != Status.REJECTED && buddy.getReceivingStatus() != Status.REJECTED)
 		{
 			// Buddy request was not accepted or rejected yet
-			// Send message to "self", as if the requested user declined the buddy request
-			UUID buddyUserAnonymizedId = buddy.getUserAnonymizedId().orElse(null); //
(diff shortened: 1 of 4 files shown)


core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -355,26 +355,46 @@ void removeBuddyInfoForRemovedUser(User user, Buddy buddy)
...
-					user.getUserAnonymizedId(), buddy.getId(), user.getDevices(), Status.REJECTED,
-					getDropBuddyMessage(DropBuddyReason.USER_ACCOUNT_DELETED, Optional.empty()));
+			prepareCleanupRemovedBuddyInPendingState(user, buddy);
 		}
 		removeBuddy(user.getId(), buddy);
 	}
 
+	private void prepareCleanupRemovedBuddyInPendingState(User user, Buddy buddy)
+	{
+		messageService.deleteUnprocessedMessages(user, m -> isBuddyConnectResponseFromBuddy(m, buddy));
```
aristotelos: It *might* occur that at the same time, the buddy status is changed to accepted, isn't it? So shouldn't we to be safe remove all messages from the buddy, not only the unprocessed ones?

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
```diff
@@ -355,26 +355,46 @@ void removeBuddyInfoForRemovedUser(User user, Buddy buddy)
...
+		// Send message to "self", as if the requested user declined the buddy request
+		UUID buddyUserAnonymizedId = buddy.getUserAnonymizedId().orElse(null); //
```
aristotelos: Empty comment `//`

core/src/main/resources/application.properties
```diff
@@ -12,6 +12,7 @@ spring.jpa.show-sql=false
 spring.jpa.properties.hibernate.session.events.log=false
 # logging.level.org.hibernate.SQL=DEBUG
 # logging.level.org.hibernate=DEBUG
+# logging.level.org.hibernate.type.descriptor.sql=TRACE
```
aristotelos: This should not go to production, should it?

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
Bert-R: No, this cannot occur. The current user would need to switch it to accepted, but that is impossible for a buddy that overwrote their account.

core/src/main/resources/application.properties
Bert-R: It's commented. I left it in for convenience, like the above ones. Every time I needed to Google to find the log category.

core/src/main/java/nu/yona/server/subscriptions/service/BuddyService.java
Bert-R: This was already part of the code that I extracted into a method. Removed it now.

PR #568, opened 2019-11-02
YD-657 Delete messages from buddy in incomplete state

Bert-R:
The related defect is about deleting the unprocessed buddy connect response (acceptance) of a buddy that overwrote their account. That same buddy might also have sent goal conflict messages already. These also need to be removed explicitly.

Along with this:
* Improved error logging in ``MessageDto``
* The ``MessageRepository`` used destination ID often where it could use destination. Refactored this to make the code type-safe and simpler.

### DIFF
--- a/core/src/main/java/nu/yona/server/messaging/entities/MessageSource.java
+++ b/core/src/main/java/nu/yona/server/messaging/entities/MessageSource.java
@@ -129,7 +129,7 @@ public class MessageSource extends EntityWithUuid
 		return message;
 	}
 
-	public List<Message> getMessageFromRelatedUserAnonymizedId(UUID relatedUserAnonymizedId)
+	public List<Message> getMessagesFromRelatedUserAnonymizedId(UUID relatedUserAnonymizedId)
 	{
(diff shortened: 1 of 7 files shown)


core/src/main/java/nu/yona/server/messaging/entities/MessageSource.java
```diff
@@ -128,6 +129,15 @@ public Message getMessage(long idToFetch)
 		return message;
 	}
 
+	public List<Message> getMessageFromRelatedUserAnonymizedId(UUID relatedUserAnonymizedId)
```
aristotelos: `getMessage` -> `getMessages`

Bert-R:
An extra test isn't necessary. I've extended the test "Overwrite Richard with an accepted but unprocessed buddy invitation from Bob" to also include a message.

core/src/main/java/nu/yona/server/messaging/entities/MessageSource.java
Bert-R: Good one! Pushed an update.

PR #569, opened 2019-11-04
YD-658 Various small code improvements

Bert-R:
A number of code improvements:
* BuddyMessage.BuddyInfoParameters holds a relatedUserAnonymizedId that is sometimes null. This possible absence should be made explicit using {{Optional}}.
* We have an old way of explicitly loading an entity. Nowadays, Hibernate.unproxy exists, so we should adopt that
* In ActivityCommentMessage.getMessagesToBeCascadinglyDeleted, we were using a non-loaded Hibernate proxy. This was masked because the default constructor (required for JPA) of BuddyMessage used the wrong super constructor, thus causing the replies collection to be initialized. This is corrected now.
* Various small changes suggested by SonarQube

Along with this:
* Fixed Eclipse project generation: Needed to include the configuration of the integration tests 
* Upgraded Gradle from 5.1.1 to 5.6.4

Remaining issue: Eclips complains on the analysisservice project that it is missing two folders. This needs to be resolved manually by removing these two folders from the build path. The cause of this issue is that the analysis service project depends on the test output of the core project (for the repositories), but now Eclipse wants test folders under all source folders of the core project.


core/src/main/java/nu/yona/server/analysis/entities/ActivityCommentMessage.java
```diff
@@ -159,8 +160,9 @@ public void prepareForDelete()
 		{
 			// This is the sender copy, so do the work
 			Set<Message> messagesToBeCascadinglyDeleted = getMessagesToBeCascadinglyDeletedForJustThisObject();
-			messagesToBeCascadinglyDeleted.addAll(buddyMessage.getMessagesToBeCascadinglyDeletedForJustThisObject());
-			messagesToBeCascadinglyDeleted.add(buddyMessage);
+			ActivityCommentMessage loadedBuddyMessage = EntityUtil.enforceLoading(buddyMessage);
```
aristotelos: Is this a bugfix?

Bert-R: I accidentally created this change set based on another branch, which has this fix. Let's merge the other PRs first and then merge back into this branch. That should make this change go away.

Bert-R: Sorry, no. This is indeed a bug fix. This I mentioned in my first commit and should have mentioned in the PR text:
> * We have an old way of explicitly loading an entity. Nowadays, Hibernate.unproxy exists, so we should adopt that
> * In ActivityCommentMessage.getMessagesToBeCascadinglyDeleted, we were using a non-loaded Hibernate proxy. This was masked because the default constructor (required for JPA) of BuddyMessage used the wrong super constructor, thus causing the replies collection to be initialized. This is corrected now.

Bert-R: Updated the PR description.

PR #571, opened 2019-11-16
YD-660 Fixed detached entity error

Bert-R:
This scenario failed:
* User gets invited as buddy
* Invited user overwrites their account
* From the new account, they request the inviting user to become their buddy
* The inviting buddy now cannot process the messages anymore

This was because the direct messages were removed from the named destination and directly inserted in the anonymous destination.

This is now resolved by duplicating the messages. The original one is deleted from the direct destination and a duplicate of it is attached to the anonymous destination.

Along with this:
* Added extra assert to BasicBuddyTest.'Richard finds Bob\'s buddy connect response', to track down a duplicate message issue.
* Updated AnalysisEngineHibernateStatsTest.groovy to lower the number of entities loaded. That somehow improved.
* Removed MessageService.sendDirectMessageAndFlushToDatabase. THe regular sendDirectMessage now always flushes
* FirebaseService.sendMessage now does not save and flush the message anymore but it demands the message to be saved and flushed by its caller.

### DIFF
--- a/core/src/main/java/nu/yona/server/messaging/entities/Message.java
+++ b/core/src/main/java/nu/yona/server/messaging/entities/Message.java
@@ -4,6 +4,7 @@
  *******************************************************************************/
 package nu.yona.server.messaging.entities;
 
+import java.lang.reflect.InvocationTargetException;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.HashSet;
@@ -24,6 +25,7 @@ import nu.yona.server.crypto.pubkey.Decryptor;
 import nu.yona.server.crypto.pubkey.Encryptor;
 import nu.yona.server.entities.EntityWithId;
 import nu.yona.server.entities.RepositoryProvider;
+import nu.yona.server.exceptions.YonaException;
 import nu.yona.server.util.TimeUtil;
 
 @Entity
@@ -77,6 +79,25 @@ public abstract class Message extends EntityWithId
 		this.isSentItem = isSentItem;
 		this.replies = new ArrayList<>();
 		this.messagesInThread = new ArrayList<>();
+	}
+
+	/**
+	 * Copy constructor. See {@link nu.yona.server.messaging.entities.Message#duplicate()}
+	 * 
+	 * @param original Message to copy.
+	 */
+	protected Message(Message original)
+	{
+		this.decryptionInfo = original.decryptionInfo;
+		this.relatedUserAnonymizedId = original.relatedUserAnonymizedId;
+		this.threadHeadMessage = null;
+		this.messagesInThread = new ArrayList<>();
+		this.repliedMessage = null;
+		this.replies = new ArrayList<>();
+		this.messageDestination = null;
+		this.creationTime = original.creationTime;
+		this.isSentItem = original.isSentItem;
+		this.isRead = original.isRead;
 	}
 
 	public static MessageRepository getRepository()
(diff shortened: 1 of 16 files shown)


core/src/main/java/nu/yona/server/messaging/entities/Message.java
```diff
@@ -199,4 +220,34 @@ public void prepareForDelete()
...
+
+	/**
+	 * Creates a duplicate of the given messages, ready to be sent to a destination. Everything is identical to the original one,
+	 * except for:
+	 * <ul>
+	 * <li>The ID, which will be 0</li>
+	 * <li>References to other messages (e.g. replies). If this is ever necessary, it needs to be designed and built</li>
+	 * <li>Reference to the message destination. The duplicate is ready to be sent and does not have a destination yet.</li>
+	 * </ul>
+	 * For this to work, the message class needs to have a public copy constructor.
```
aristotelos: This currently has been only defined for possible direct messages, right? Maybe good to document that here.

core/src/test/java/nu/yona/server/messaging/service/MessageServiceTest.java
```diff
@@ -207,6 +207,7 @@ public void sendDirectMessage_userCreatedOnBuddyRequest_doesNotSendSms()
 	public void prepareMessageCollection_transferringDirectMessagesToAnonymousDestination_sendsFirebaseMessage()
 	{
 		Message message = Mockito.mock(Message.class);
+		when(message.duplicate()).thenReturn(message);
```
aristotelos: Can we not add some more asserts about what `prepareMessageCollection` does now?

core/src/main/java/nu/yona/server/messaging/entities/Message.java
```diff
@@ -79,6 +81,25 @@ protected Message(Optional<UUID> relatedUserAnonymizedId, boolean isSentItem)
...
 	}
 
+	/**
+	 * Copy constructor. See {@link nu.yona.server.messaging.entities.Message#duplicate()}
+	 * 
+	 * @param original Message to copy.
+	 */
+	protected Message(Message original)
+	{
+		this.decryptionInfo = original.decryptionInfo;
```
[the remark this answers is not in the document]
Bert-R: You're totally right. This is set when posting the message to a destination.

core/src/main/java/nu/yona/server/messaging/entities/Message.java
```diff
@@ -199,4 +220,34 @@ public void prepareForDelete()
...
 	protected abstract void decrypt();
+
+	/**
+	 * Creates a duplicate of the given messages, ready to be sent to a destination. Everything is identical to the original one,
+	 * except for:
+	 * <ul>
+	 * <li>The ID, which will be 0</li>
+	 * <li>References to other messages (e.g. replies). If this is ever necessary, it needs to be designed and built</li>
+	 * <li>Reference to the message destination. The duplicate is ready to be sent and does not have a destination yet.</li>
+	 * </ul>
```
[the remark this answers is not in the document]
Bert-R: Added a similar line

core/src/main/java/nu/yona/server/messaging/entities/Message.java
Bert-R: Makes sense.

core/src/test/java/nu/yona/server/messaging/service/MessageServiceTest.java
Bert-R: This change doesn't really change that, only the JPA-related behavior is changed, but that is hard to unit test. Given the urgency of this fix, I'm not going to spend time on this now.

Bert-R:
Pushed an update

PR #576, opened 2019-11-27
YD-662 Move creation time from User to UserPrivate

Bert-R:
This field is an identification risk. We only use it for management tasks, but for that we don't need the precise date, so we could as well round all creation dates to the first of the month and keep the real creation date encrypted inside UserPrivate.

By keeping it there, we can still see when the user was created and use that for decisions like determining the "earliest possible date" for activities.

The rounding only happens for users that use the app for more than 30 days. This way, we can still get insight in the usage statistics and see early leavers. The risk for these short-term users is lot smaller as they gather little data.

### DIFF
--- a/core/src/main/java/nu/yona/server/subscriptions/entities/User.java
+++ b/core/src/main/java/nu/yona/server/subscriptions/entities/User.java
@@ -40,7 +40,8 @@ public class User extends EntityWithUuid
 	@Column(unique = true)
 	private String mobileNumber;
 
-	private LocalDate roundedCreationDate;
+	// YD-666 Change to LocalDate when all users have migrated their data
+	private LocalDateTime roundedCreationDate;
 
 	private LocalDate appLastOpenedDate;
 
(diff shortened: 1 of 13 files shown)


core/src/main/java/nu/yona/server/subscriptions/entities/User.java
```diff
@@ -40,7 +40,8 @@
 	@Column(unique = true)
 	private String mobileNumber;
 
-	private LocalDateTime creationTime;
+	// YD-666 Change to LocalDate when all users have migrated their data
```
aristotelos: Why can't it immediately be a `LocalDate`?

Bert-R: Because we then cannot migrate the time part of it anymore in the migration step.

aristotelos: OK, the comment is about the `creationTime` and not about the `roundedCreationDate`?
Why don't we make `roundedCreationDate` a `LocalDate` instead of `LocalDateTime`?

Bert-R: It at least needs a bigger comment :)
This is how it works:
* This PR renames ``creationTime`` into ``roundedCreationDate``
* The migration step copies the date/time into the private (encrypted) data and rounds the time to 'start of day'
* If the user is active for more than 30 days, the date is set to the first of the month

After everyone has executed the migration step, the ``roundedCreationDate`` column won't contain any times (other than 00:00) anymore, so the data type can be changed without losing data.

Makes sense? If yes, I'll explain this in a comment in the source code.

aristotelos: Makes sense, I missed that the existing column was renamed.

PR #577, opened 2019-12-03
YD-652 Add app version info to logging context

Bert-R:
Added the information from the Yona-App-Version header to the MDC.

Along with this corrected the message IDs of the Dutch messages for PIN reset confirmation codes from "error.pin.reset.overwrite.confirmation" into "error.pin.reset.request.confirmation".

### DIFF
--- a/core/src/main/java/nu/yona/server/rest/ErrorLoggingFilter.java
+++ b/core/src/main/java/nu/yona/server/rest/ErrorLoggingFilter.java
@@ -97,18 +97,10 @@ public class ErrorLoggingFilter implements Filter
 			MDC.put(APP_VERSION_NAME_MDC_KEY, versionName);
 		}
 
-		private static Optional<YonaException> validateHeaders(HttpServletRequest request)
+		private static void assertValidHeaders(HttpServletRequest request)
 		{
-			try
-			{
-				Optional.ofNullable(request.getHeader(Constants.APP_VERSION_HEADER))
-						.ifPresent(LoggingContext::validateAppVersionHeader);
-			}
-			catch (YonaException e)
-			{
-				return Optional.of(e);
-			}
-			return Optional.empty();
+			Optional.ofNullable(request.getHeader(Constants.APP_VERSION_HEADER))
+					.ifPresent(LoggingContext::validateAppVersionHeader);
 		}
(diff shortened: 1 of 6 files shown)


core/src/main/java/nu/yona/server/rest/ErrorLoggingFilter.java
```diff
@@ -72,6 +86,52 @@ private static String getCorrelationId(HttpServletRequest request)
...
+		{
+			Optional.ofNullable(request.getHeader(Constants.APP_VERSION_HEADER))
+					.ifPresent(LoggingContext::validateAppVersionHeader);
+		}
+
+		private static void validateAppVersionHeader(String header)
+		{
+			String[] parts = header.split("/");
+			Require.that(parts.length == 3, () -> InvalidDataException.invalidAppVersionHeader(header));
+			assertValidOperatingSystem(parts[0]);
```
aristotelos: This puts quite strict requirements on the app version header. Can we test this in time that it does not block the app?

core/src/main/java/nu/yona/server/rest/ErrorLoggingFilter.java
```diff
@@ -72,6 +86,52 @@ private static String getCorrelationId(HttpServletRequest request)
 			return traceIdHeader;
 		}
 
+		private static void putAppVersionContext(String header)
+		{
+			String[] parts = header.split("/");
+			String operatingSystem = parts[0];
+			int versionCode = Integer.parseInt(parts[1]);
+			String versionName = parts[2];
+			MDC.put(APP_OS_MDC_KEY, operatingSystem);
```
aristotelos: Have we tested this?

core/src/main/java/nu/yona/server/rest/ErrorLoggingFilter.java
Bert-R: Yes, on the beta server

core/src/main/java/nu/yona/server/rest/ErrorLoggingFilter.java
Bert-R: Yes, through the Groovy test

PR #578, opened 2019-12-10
YD-668 Do not allow leading zeros in mobile nr

Bert-R:
Till now, we allowed numbers like +3106..., with a zero (could even be more than one) between the country code and the rest of the number. It looks like the mobile number confirmation logic generally doesn't work in that case, as only one user has a confirmed number of this format. However, if they use it to fetch a new device request, that fails. Similar issues exist when inviting buddies.

With this change, such numbers are not allowed anymore.


core/src/test/java/nu/yona/server/subscriptions/entities/BuddyTest.java
```diff
@@ -1,12 +1,11 @@
 /*******************************************************************************
- * Copyright (c) 2018, 2019 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License, v.
- * 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ * Copyright (c) 2018, 2019 Stichting Yona Foundation This Source Code Form is subject to the terms of the Mozilla Public License,
```
aristotelos: 2020 now? ;)

core/src/main/java/nu/yona/server/subscriptions/service/UserAssertionService.java
```diff
@@ -68,29 +70,40 @@ static void assertValidUserFields(UserDto user, UserService.UserPurpose purpose)
...
 		}
 		catch (NumberParseException e)
 		{
 			throw YonaException.unexpected(e);
 		}
 	}
 
+	private static void assertNoLeadingZeros(PhoneNumberUtil util, String mobileNumberStr, PhoneNumber mobileNumber)
+	{
+		if (!mobileNumberStr.equals(util.format(mobileNumber, PhoneNumberFormat.INTERNATIONAL).replaceAll("[ -]", "")))
```
aristotelos: This line is a bit hard to read for me. Does it really state that there are leading zeros? It seems that it is a very general validation...

core/src/test/java/nu/yona/server/subscriptions/entities/BuddyTest.java
Bert-R: For new PRs, yes, but this one was created in 2019

core/src/main/java/nu/yona/server/subscriptions/service/UserAssertionService.java
Bert-R: What it does is a "normalization" of the number: format the number in the international format and remove all decorations from it. Then the number should be identical to the original one.
Shall I add that in a comment?

Bert-R: Added the comment

PR #580, opened 2019-12-19
YD-669 Pass app info all the way

Bert-R:
It now goes from the app service to invoked services (e.g. the analysis service) and also to the async thread that sends the Firebase message, so it can be included in the error logging there.

Along with this:
* Improved the error logging for errors returned from an invoked service
* Included the token in the exception logging for Firebase issues
* Made the parameter order in tests consistent. Now the parameter order is always "parameters, headers", where "headers" can be omitted.

### DIFF
--- a/appservice/src/intTest/groovy/nu/yona/server/FirebaseTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/FirebaseTest.groovy
@@ -59,10 +59,10 @@ class FirebaseTest extends AbstractAppServiceIntegrationTest
 		def appVersionCode = 123
 		def appVersionName = "1.2 (I'm Rich)"
 		def headers = ["Yona-App-Version" : "$appOs/$appVersionCode/$appVersionName"]
-		appService.sendBuddyConnectRequest(richard, bobAndroid, true, headers)
+		appService.sendBuddyConnectRequest(richard, bobAndroid, true, [:], headers)
 
 		when:
-		def messageUrlBobAndroid = appService.getMessages(bobAndroid, headers, [:]).responseData._embedded."yona:messages".findAll{ it."@type" == "BuddyConnectRequestMessage"}[0]._links.self.href
+		def messageUrlBobAndroid = appService.getMessages(bobAndroid, [:], headers).responseData._embedded."yona:messages".findAll{ it."@type" == "BuddyConnectRequestMessage"}[0]._links.self.href
 		def responseBobAndroid = appService.getLastFirebaseMessage(bobAndroid.requestingDevice.firebaseInstanceId)
 
 		then:
@@ -184,7 +184,7 @@ class FirebaseTest extends AbstractAppServiceIntegrationTest
 		def headers = ["Yona-App-Version" : "$appOs/$appVersionCode/$appVersionName"]
 
 		when:
-		assertResponseStatusNoContent(appService.postAppActivityToAnalysisEngine(richard, richard.requestingDevice, AppActivity.singleActivity("Poker App", startTime, endTime), headers, [:]))
+		assertResponseStatusNoContent(appService.postAppActivityToAnalysisEngine(richard, richard.requestingDevice, AppActivity.singleActivity("Poker App", startTime, endTime), [:], headers))
 
 		then:
 		def responseRichard = analysisService.getLastFirebaseMessage(richard.requestingDevice.firebaseInstanceId)
--- a/appservice/src/intTest/groovy/nu/yona/server/FirebaseTest.groovy
+++ b/appservice/src/intTest/groovy/nu/yona/server/FirebaseTest.groovy
@@ -8,6 +8,9 @@ package nu.yona.server
 
 import static nu.yona.server.test.CommonAssertions.*
 
+import java.time.ZonedDateTime
+
+import nu.yona.server.test.AppActivity
 import nu.yona.server.test.Device
 import nu.yona.server.test.User
 
@@ -26,9 +29,9 @@ class FirebaseTest extends AbstractAppServiceIntegrationTest
 
 		then:
 		assertResponseStatusOk(responseRichard)
-		assert responseRichard.responseData.title == "Message received"
-		assert responseRichard.responseData.body == "Tap to open message"
-		assert messageUrlRichard.endsWith(responseRichard.responseData.data.messageId.toString())
+		responseRichard.responseData.title == "Message received"
+		responseRichard.responseData.body == "Tap to open message"
+		messageUrlRichard.endsWith(responseRichard.responseData.data.messageId.toString())
 
 		when:
 		def messageUrlBob = appService.getMessages(bob).responseData._embedded."yona:messages".findAll{ it."@type" == "BuddyConnectRequestMessage"}[0]._links.self.href
@@ -36,9 +39,9 @@ class FirebaseTest extends AbstractAppServiceIntegrationTest
 
 		then:
 		assertResponseStatusOk(responseBob)
-		assert responseBob.responseData.title == "Message received"
-		assert responseBob.responseData.body == "Tap to open message"
-		assert messageUrlBob.endsWith(responseBob.responseData.data.messageId.toString())
+		responseBob.responseData.title == "Message received"
+		responseBob.responseData.body == "Tap to open message"
+		messageUrlBob.endsWith(responseBob.responseData.data.messageId.toString())
 
 		cleanup:
 		appService.deleteUser(richard)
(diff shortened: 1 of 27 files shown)


core/src/main/java/nu/yona/server/rest/HeadersHolder.java
```diff
@@ -0,0 +1,51 @@
...
+		String value = headers.getFirst(Constants.APP_VERSION_HEADER);
+		if (value != null)
+		{
+			storedHeaders.put(name, value);
+		}
+	}
+
+	public void writeTo(HttpHeaders headers)
+	{
+		storedHeaders.forEach((n, v) -> headers.add(n, v));
```
yonabuild: MINOR Replace this lambda with a method reference. [rule](http://10.127.10.107:9000/coding_rules#rule_key=squid%3AS1612)

yonabuild:
SonarQube analysis reported 1 issue
* ![MINOR][MINOR] 1 minor

[MINOR]: https://sonarsource.github.io/sonar-github/severity-minor.png 'Severity: MINOR'

core/src/main/java/nu/yona/server/messaging/service/FirebaseService.java
```diff
@@ -78,19 +84,11 @@ public void sendMessage(String registrationToken, nu.yona.server.messaging.entit
...
-		{
-			logger.info("Firebase message not sent because Firebase is disabled");
-			// Store for testability
-			lastMessageByRegistrationToken.put(registrationToken, firebaseMessage);
-		}
+		// Sending takes quite a bit of time, so do it asynchronously
+		ThreadData threadData = asyncExecutor.getThreadData();
+		CompletableFuture
+				.runAsync(() -> asyncExecutor.initThreadAndDo(threadData, () -> sendMessage(registrationToken, firebaseMessage)))
+				.whenCompleteAsync((r, t) -> logIfCompletedWithException(t, registrationToken));
```
aristotelos: Does this `logIfCompletedWithException` not need an `initThreadAndDo` call?

core/src/main/java/nu/yona/server/CoreConfiguration.java
```diff
@@ -107,10 +106,10 @@ public CurieProvider curieProvider()
 	}
 
 	@Bean
-	public RestTemplate restTemplate(ObjectMapper objectMapper)
+	public static CustomScopeConfigurer customScopeConfigurer()
 	{
-		RestTemplate restTemplate = new RestTemplate();
-		restTemplate.setErrorHandler(new RestClientErrorHandler(objectMapper));
-		return restTemplate;
+		CustomScopeConfigurer configurer = new CustomScopeConfigurer();
```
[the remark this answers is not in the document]
aristotelos: Ah, I see that this is needed for thread local storage of the `HeadersHolder`

core/src/main/java/nu/yona/server/ThreadScope.java
```diff
@@ -0,0 +1,15 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v.2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at https://mozilla.org/MPL/2.0/.
+ *******************************************************************************/
+package nu.yona.server;
+
+import org.springframework.context.annotation.Scope;
+
+@Scope("thread")
+public @interface ThreadScope
```
aristotelos: This interface should be a bit better documented.

core/src/main/java/nu/yona/server/rest/TestController.java
```diff
@@ -153,31 +154,52 @@ private FirebaseMessageResourceAssembler createFirebaseMessageResourceAssembler(
...
 			try
 			{
-				Notification notification = (Notification) notificationField.get(message);
-				Map<String, String> data = (Map<String, String>) dataField.get(message);
+				Notification notification = (Notification) notificationField.get(messageData.firebaseMessage);
+				Map<String, String> data = (Map<String, String>) dataField.get(messageData.firebaseMessage);
 				String title = (String) titleField.get(notification);
 				String body = (String) bodyField.get(notification);
-				return new FirebaseMessageDto(title, body, data);
+				String appOs = getStringValueFromMdc(messageData, Constants.APP_OS_MDC_KEY).orElse(null);
```
aristotelos: Isn't it a bit strange to get this from MDC while it is present in `HeadersHolder`?

core/src/main/java/nu/yona/server/rest/HeadersHolder.java
```diff
@@ -0,0 +1,49 @@
...
+ * v.2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *******************************************************************************/
+package nu.yona.server.rest;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.springframework.http.HttpHeaders;
+
+public class HeadersHolder
```
aristotelos: Instead of just storing all headers, I would prefer to have a more speaking class like `AppInformationHolder`.

appservice/src/intTest/groovy/nu/yona/server/FirebaseTest.groovy
```diff
@@ -52,27 +55,37 @@ class FirebaseTest extends AbstractAppServiceIntegrationTest
 		User bobAndroid = addBob(true, "ANDROID", "nl-NL")
 		bobAndroid.emailAddress = "bob@dunn.com"
 		User bobIos = appService.addDevice(bobAndroid, "Bob's iPhone", "IOS", Device.SOME_APP_VERSION)
-		appService.sendBuddyConnectRequest(richard, bobAndroid)
+		def appOs = "ANDROID"
```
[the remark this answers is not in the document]
Bert-R: OK

core/src/main/java/nu/yona/server/rest/HeadersClientInterceptor.java
```diff
@@ -0,0 +1,31 @@
...
+package nu.yona.server.rest;
+
+import java.io.IOException;
+
+import org.springframework.http.HttpRequest;
+import org.springframework.http.client.ClientHttpRequestExecution;
+import org.springframework.http.client.ClientHttpRequestInterceptor;
+import org.springframework.http.client.ClientHttpResponse;
+
+public class HeadersClientInterceptor implements ClientHttpRequestInterceptor
```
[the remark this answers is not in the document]
Bert-R: Added a JavaDoc comment.

core/src/main/java/nu/yona/server/rest/Config.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.web.client.RestTemplate;
+import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import nu.yona.server.ThreadScope;
+
+@Configuration
+public class Config implements WebMvcConfigurer
```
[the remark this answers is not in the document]
Bert-R: It's a configuration (as indicated by the ``@Configuration`` annotation) with REST-related stuff (and that's indicated because it is in the ``nu.yona.server.rest`` package). Naming it ``WebMvcConfig`` would indicate a more narrow scope than that.

core/src/main/java/nu/yona/server/messaging/service/FirebaseService.java
Bert-R: I don't think so. I believe I have manually verified this. If it is lacking, we'll need to add this in a separate PR.

core/src/main/java/nu/yona/server/ThreadScope.java
Bert-R: Agreed. Added a comment.

core/src/main/java/nu/yona/server/rest/TestController.java
Bert-R: The test controller is here to verify the firebase messaging and related logging. For that reason, it takes it from the place LogBack would take it from as well: the MDC. Makes sense?

core/src/main/java/nu/yona/server/rest/HeadersHolder.java
Bert-R: What about ``PassThroughHeadersHolder``? It holds 'headers', not just 'information', so I'd like to be more specific. At the other hand, it doesn't just hold headers, but the headers that are to be passed through from the incoming requests to the outgoing ones.

Bert-R:
Pushed an update

core/src/main/java/nu/yona/server/rest/HeadersClientInterceptor.java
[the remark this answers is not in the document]
aristotelos: I still don't understand.... For what requests is it? Requests from the app service to the analysis service?

core/src/main/java/nu/yona/server/rest/HeadersHolder.java
aristotelos: I still would prefer a domain-centric name with something indicating this is app information instead of a solution-centric name containing something about headers...

core/src/main/java/nu/yona/server/rest/HeadersClientInterceptor.java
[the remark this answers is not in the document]
Bert-R: All requests from one service to another, so from app to batch and from app to analysis

core/src/main/java/nu/yona/server/rest/TestController.java
aristotelos: Wouldn't somebody understand it better when it is plainly got from the `HeadersHolder`? `MDC` is only for logging, while this method is about getting the app information, not the logging.

core/src/main/java/nu/yona/server/rest/HeadersHolder.java
Bert-R: The class provides the solution of holding headers that are passed through in HTTP requests. Not just app information, but HTTP headers. To me, the class name tells what it does. The earlier name said "it holds headers". The new name is specific about what headers it holds.

core/src/main/java/nu/yona/server/rest/TestController.java
Bert-R: No. The headers holder is just a container/holder to hold the headers that are passed through in HTTP requests. We use some information that's communicated through headers as "mapped diagnostic context" in logging. The controller enables a test to see that the right information made it into this diagnostic context. It did so using the pass through headers solution, but the integration test verifies the result not the means.

core/src/main/java/nu/yona/server/rest/HeadersServerInterceptor.java
```diff
@@ -13,6 +11,11 @@
 import org.springframework.lang.Nullable;
 import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
 
+/**
+ * Intercept all incoming HTTP requests, with the intend to store headers that are to be passed through in outgoing HTTP requests.
```
aristotelos: intend -> intent

core/src/main/java/nu/yona/server/rest/PassThroughHeadersHolder.java
```diff
@@ -0,0 +1,54 @@
...
+	private final Map<String, String> storedHeaders = new HashMap<>();
+
+	public void readFrom(HttpHeaders headers)
+	{
+		storeIfPresent(headers, Constants.APP_VERSION_HEADER);
+	}
+
+	private void storeIfPresent(HttpHeaders headers, String name)
+	{
+		String value = headers.getFirst(Constants.APP_VERSION_HEADER);
```
aristotelos: `Constants.APP_VERSION_HEADER` can be replaced by `name` here

Bert-R:
Pushed an update

PR #583, opened 2020-01-18
YD-675 Corrected prev links on day/week details

Bert-R:


appservice/src/intTest/groovy/nu/yona/server/ActivityTest.groovy
```diff
@@ -142,28 +154,28 @@ class ActivityTest extends AbstractAppServiceIntegrationTest
 		appService.deleteUser(richard)
 	}
 
-	def 'Retrieve activity report of previous week'()
+	def 'Retrieve activity report of previous weeks'()
```
Bert-R: I'm going to revert this change and instead make a test "Richard gets proper navigation links between day and week details". That'll be easier to understand.

Bert-R: Done.

core/src/main/java/nu/yona/server/analysis/entities/IntervalActivity.java
```diff
@@ -119,19 +121,29 @@ public ZonedDateTime getStartTime()
...
+		return hasPrevious(earliestPossibleDate, goal, getStartTime(), getTimeUnit());
 	}
 
 	public boolean hasNext()
 	{
 		return hasNext(getStartTime(), getTimeUnit());
 	}
 
-	public static boolean hasPrevious(Goal goal, ZonedDateTime startTime, TemporalUnit timeUnit)
+	public static boolean hasPrevious(LocalDate earliestPossibleDate, Goal goal, ZonedDateTime startTime, TemporalUnit timeUnit)
```
aristotelos: It may be good to add a unit test for this method as well

PR #584, opened 2020-01-28
YD-676 Clear Firebase ID if not registered anymore

Bert-R:
If the Firebase instance ID is not registered anymore, the Firbase web service returns a 404. In that case, we clear the Firebase instance ID on the DeviceAnonymized entity, so we stop trying to push notifications to that device.

Along with this:
* Added a method to find a DeviceAnonymized without having the UserAnonymized ID available. That ID would otherwise be used for logging, but that isn't worth carrying the UserAnonymized around.
* Used the Spring-way of asynchronous execution. This way, the thread is automatically initialized for Spring behavior.
* Fixed an issue in the Groovy Firebase tests that caused intermittent failures because the asynchronous execution wasn't completed before the Firebase message was fetched by the test
* Created a unit test for AsyncExecutor
* Made the AsyncExecutor clear the MDC and header holder upon completion of the task

### DIFF
--- a/core/src/main/java/nu/yona/server/messaging/service/FirebaseService.java
+++ b/core/src/main/java/nu/yona/server/messaging/service/FirebaseService.java
@@ -10,7 +10,7 @@ import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.CompletableFuture;
+import java.util.UUID;
 
 import javax.annotation.PostConstruct;
 
@@ -30,6 +30,7 @@ import com.google.firebase.messaging.Notification;
 
 import nu.yona.server.Constants;
 import nu.yona.server.Translator;
+import nu.yona.server.device.service.DeviceService;
 import nu.yona.server.exceptions.YonaException;
 import nu.yona.server.properties.YonaProperties;
 import nu.yona.server.util.AsyncExecutor;
@@ -39,6 +40,7 @@ import nu.yona.server.util.Require;
 @Service
 public class FirebaseService
 {
+	private final static String FIREBASE_ID_NOT_REGISTERED = "registration-token-not-registered";
 	private static final Logger logger = LoggerFactory.getLogger(FirebaseService.class);
 
 	private final Map<String, MessageData> lastMessageByRegistrationToken = new HashMap<>();
@@ -51,6 +53,9 @@ public class FirebaseService
 
 	@Autowired
 	private AsyncExecutor asyncExecutor;
+
+	@Autowired
+	private DeviceService deviceService;
 
 	@PostConstruct
 	private void init()
@@ -76,7 +81,7 @@ public class FirebaseService
 		}
 	}
 
-	public void sendMessage(String registrationToken, nu.yona.server.messaging.entities.Message message)
+	public void sendMessage(UUID deviceAnonymizedId, String registrationToken, nu.yona.server.messaging.entities.Message message)
 	{
 		// The message URL might seem useful notification payload, but that is not possible as messages can be sent from anonymous
 		// contexts, while the URL requires the user ID.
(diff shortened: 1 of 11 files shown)


core/src/main/java/nu/yona/server/messaging/service/FirebaseService.java
```diff
@@ -39,6 +40,7 @@
 @Service
 public class FirebaseService
 {
+	private final static String FIREBASE_ID_NOT_REGISTERED = "registration-token-not-registered";
```
yonabuild: MINOR Reorder the modifiers to comply with the Java Language Specification. [rule](http://10.127.10.107:9000/coding_rules#rule_key=squid%3AModifiersOrderCheck)

core/src/main/java/nu/yona/server/util/AsyncExecutor.java
```diff
@@ -38,10 +41,47 @@ public ThreadData getThreadData()
...
-		headersHolder.importFrom(threadData.headers);
-		action.run();
+		try (Handler handler = Handler.initialize(headersHolder, threadData))
+		{
+			try
+			{
+				action.run();
+				completionHandler.accept(Optional.empty());
+			}
+			catch (Throwable e)
```
yonabuild: MAJOR Catch Exception instead of Throwable. [rule](http://10.127.10.107:9000/coding_rules#rule_key=squid%3AS1181)

yonabuild:
SonarQube analysis reported 2 issues
* ![MAJOR][MAJOR] 1 major
* ![MINOR][MINOR] 1 minor

[MINOR]: https://sonarsource.github.io/sonar-github/severity-minor.png 'Severity: MINOR'

core/src/main/java/nu/yona/server/util/AsyncExecutor.java
Bert-R: Here we really want to catch Throwable

yonabuild:
SonarQube analysis reported 1 issue
* ![MAJOR][MAJOR] 1 major

core/src/test/java/nu/yona/server/util/AsyncExecutorTest.java
```diff
@@ -0,0 +1,131 @@
...
+		CountDownLatch doneSignal = new CountDownLatch(1);
+		DataCarrier actionHandlerData = new DataCarrier();
+		DataCarrier completionHandlerData = new DataCarrier();
+
+		service.execAsync(service.getThreadData(), () -> actionHandler(actionHandlerData, expectedData.exception),
+				(t) -> completionHandler(doneSignal, completionHandlerData, t));
+		doneSignal.await();
+
+		assertReturnedData(completionHandlerData, expectedData);
+		expectedData.exception = Optional.empty();
```
aristotelos: I am a bit puzzled by this last assert about action handler data. I think it is easier to create 3 tests instead of one, so that this line is not needed:

1. `execAsync_default_dataIsPresentInActionHandler`
2. `execAsync_completionWithError_dataIsPresentInCompletionHandlerThrowableIsAvailable`
3. `execAsync_completionSuccess_dataIsPresentInCompletionHandlerThrowableIsEmpty`

Bert-R: How about extending ``assertReturnedData`` with a parameter to indicate it's asserting action data or assertion data? Or even making two methods for it?

Bert-R: I've created two assertion methods. That makes it clear what's being asserted.

Bert-R:
Pushed an update

core/src/test/java/nu/yona/server/util/AsyncExecutorTest.java
aristotelos: That makes it better, but the situation where the action runs into an exception does not need an assert on the action data. So I would still favor three methods that assert a single thing.

Bert-R: I'don't see why we shouldn't assert the data of the action handler at the time it will fail. That's still valid to check. Sure, we could omit that entire assert, as it's covered by the first test already, but in the second test it needs to have the same values before it fails. Three tests doesn't add anything, as there are only two scenarios:
* The action handler succeeds
* The action handler fails

In either case, the action handler and the completion handler need to have all data and in the failure case, the completion handler also needs to have the exception.

PR #585, opened 2020-02-01
YD-677 Resolved multithreading issue pass-through headers

Bert-R:
The issue was that we were trying to inject a thread-scoped bean into a singleton bean. However, the injection on a singleton bean happens only once, so the same PassThroughHeaders instance was used on all threads.

Added a test that reproduces the issue, created a straightforward fix and removed the thread-scope stuff.


core/src/main/java/nu/yona/server/rest/PassThroughHeadersHolder.java
```diff
@@ -7,16 +7,26 @@
...
 /**
  * Holds the headers that are received in an incoming HTTP request, which needs to be passed-through in outgoing HTTP requests.
  * See {@link HeadersServerInterceptor} for the place where the incoming headers are stored in the holder and
  * {@link HeadersClientInterceptor} for the place where the headers are read.
  */
+@Component
 public class PassThroughHeadersHolder
 {
-	private final Map<String, String> storedHeaders = new HashMap<>();
+	private final ThreadLocal<Map<String, String>> threadLocal = new NamedThreadLocal<Map<String, String>>(
```
aristotelos: One thing I have experienced before with thread local is that for example IIS uses a thread pool and sometimes reuses a thread. Can we get the same issue with Apache? So should we not at least clear the holder on the `HeadersServerInterceptor` to be sure?

core/src/main/java/nu/yona/server/rest/TestController.java
```diff
@@ -52,6 +60,11 @@
 	@Autowired
 	private FirebaseService firebaseService;
 
+	@Autowired
+	private PassThroughHeadersHolder headersHolder;
+
+	private final CyclicBarrier barrier = new CyclicBarrier(2);
```
aristotelos: For clarity `passThroughHeadersRequestBarrier` is more speaking here

core/src/main/java/nu/yona/server/rest/TestController.java
```diff
@@ -145,6 +129,77 @@ private FirebaseMessageResourceAssembler createFirebaseMessageResourceAssembler(
 		return new FirebaseMessageResourceAssembler();
 	}
 
+	/**
+	 * Returns the headers stored in the {@link PassThroughHeadersHolder}. This method blocks till a second request is done, thus
+	 * enforcing multithreading.
+	 * 
+	 * @return the headers stored in the {@link PassThroughHeadersHolder}
+	 */
+	@GetMapping(value = "/passThrougHeaders")
```
aristotelos: `passThrougHeaders` -> `passThroughHeaders`

core/src/main/java/nu/yona/server/rest/TestController.java
```diff
@@ -145,6 +129,77 @@ private FirebaseMessageResourceAssembler createFirebaseMessageResourceAssembler(
...
+	@ResponseBody
+	public HttpEntity<PassThroughHeadersResource> getPassThroughHeaders()
+	{
+		Require.that(yonaProperties.isTestServer(),
+				() -> InvalidDataException.onlyAllowedOnTestServers("Endpoint /passThrougHeaders is not available"));
```
aristotelos: `passTrougHeaders` -> `passThroughHeaders`

core/src/main/java/nu/yona/server/rest/PassThroughHeadersHolder.java
Bert-R: Valid point. The threads can certainly be reused.

Bert-R: I'm going to clear it as part of ``readFrom`` and ``importFrom``.

core/src/main/java/nu/yona/server/rest/TestController.java
Bert-R: Sharp!

core/src/main/java/nu/yona/server/rest/TestController.java
Bert-R: Agreed.

core/src/main/java/nu/yona/server/rest/TestController.java
Bert-R: You're suffering from a similar issue :)
``passTrougHeaders`` -> ``passThrougHeaders`` -> ``passThroughHeaders``

Bert-R:
Pushed an update

PR #587, opened 2020-02-13
YD-679 Return 400 on MultipartException

Bert-R:
Added MultipartException to the list of exceptions that cause a 400 status to be returned.


Bert-R:
That's not worth it. If this ever starts to fail, we will get alerts for 500 errors that actually should be 400 errors. That will only anoy us a little, nothing else.

PR #588, opened 2020-02-15
YD-680 Include duration in raw activity report

Bert-R:
Instead of having to process the result in Excel, the data can now be interpreted immediately.


Bert-R:
Given this is a small change outside the regular API, which I need later this evening, I'm going to merge it myself.

PR #589, opened 2020-02-24
YD-681 Use new Spring forward header setting

Bert-R:
The old setting (server.use-forward-headers=true) was deprecated. We now use server.forward-headers-strategy=native. The old setting was deprecated to allow different strategies.


Bert-R:
Merging this very small change myself

PR #590, opened 2020-02-25
YD-682 Save the user anonymized

Bert-R:
after updating one of the anonymized devices


Bert-R:
Merging this very small change myself

PR #591, opened 2020-03-11
Upgrade helm charts for new API versions

kti-nicole:
Use new updated mariadb and openldap charts for local testing (when support_infrastructure: true) and update app deployments to new api version.

### DIFF
--- a/k8s/helm/yona/values.yaml
+++ b/k8s/helm/yona/values.yaml
@@ -184,15 +184,21 @@ apple:
   ca_certificate_file: apple/AppleWWDRCA.cer
 
 mariadb:
-  mariadbRootPassword: develop
-  mariadbUser: develop
-  mariadbPassword: develop
-  mariadbDatabase: yona
+  db:
+    user: develop
+    password: develop
+    name: yona
+  rootUser:
+    password: develop
   persistence:
-    storageClass: standard
-    size: 2Gi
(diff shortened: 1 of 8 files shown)


k8s/helm/yona/values.yaml
```diff
@@ -184,15 +184,21 @@ apple:
...
-    storageClass: standard
-    size: 2Gi
+    enabled: false
+  replication:
+    enabled: false
+  volumePermissions:
+    enabled: true
   user: develop
   password: develop
+  mariadbDatabase: yona
```
Bert-R: This one is now obsolete, right?

kti-nicole: Technically no, since this one is used for the deployments and the other tells the mariadb chart what database to create (unused in environments with already existing mysql servers), but I don't see why we can't unify them to be the same thing like it was before (the newer mysql chart changed). I'll update and push.

Bert-R: I see. I misinterpreted what caused the change. It's indeed good to unify it again.

kti-nicole:
> Looks good. Do we need to update the values files of the different environments before we can push this to production?

Yep, that should already be taken care of. Right now we have both versions set for all environments so hopefully it should just work, and we can remove the old ones once this goes out.

PR #596, opened 2020-03-20
YD-632 Implement check on supported country codes

Bert-R:


aristotelos (approved):
Looks great. I'm impressed by the list of country codes we actually support by default, is that intentional?

Bert-R:
Yona has been used in these countries and these numbers are today on our list of supported countries in Plivio, so I mirrored that.

Bert-R:
As the PR is approved, I'm going to merge it.

PR #597, opened 2020-03-24
YD-636 Temporary fix for flipping tests

Bert-R:
See the analysis on the ticket. The cause of the issue seems to be that updates to goals do not immediately reach the analysis engine service. Due to that, the next request might reach the analysis engine service before the cache update reaches it, causing the test to fail

Added a 0.5 second sleep after the goal creation, marked with a // TODO. If this indeed resolves it, we need to make up our mind on a definitive fix.


yonabuild:
Testing whether the GitHub pull request builder is configured properly

PR #599, opened 2020-03-26
YD-685 Gracefully handle activity on orphaned user anonymized

Bert-R:
A device might continue to use its VPN after the user overwrote their account. In that case, new goal conflict messages might be created for a user user anonymized that is orphaned and where the buddies already wiped the related messages and buddy entity. Such a new goal conflict message would lead to an inconsistency where the belonging buddy entity cannot be found.

To fix this, BuddyService.getBuddyUsersAnonymized now does not return buddy entities for which the corresponding buddy entity at the other side does not exist.

This has a minor side effect: earlier, if two users became buddies, the goal conflicts of the accepting user would be reported to the requesting user immediately after the accepting user accepted the request. Now this is delayed a little: goal conflicts of the accepting user are now reported to the requesting user after the requesting user processed the buddy connect response message of the accepting user.


aristotelos (approved):
Looks great, will merge.

PR #600, opened 2020-04-28
YD-686 Added length check for activity comments

Bert-R:
Earlier, there was no explicit check. Testing proved I could make a comment with 237 *s. In reality, this limit might vary due to the encryption. To stay on the safe side and still give plenty of room for a comment, the maximum is now set to 200 characters.

Along with this:
* Corrected a few error messages by taking out double curly braces
* Removed two duplicated error messages


appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
```diff
@@ -430,6 +432,45 @@ class ActivityCommentTest extends AbstractAppServiceIntegrationTest
...
+		assert budgetGoalNewsBuddyBob
+		def responseOverviewsBobAsBuddyAll = appService.getDayActivityOverviews(richard, richard.buddies[0], ["size": 14])
+		assertResponseStatusOk(responseOverviewsBobAsBuddyAll)
+		def responseDetailsBobAsBuddy = appService.getDayDetailsFromOverview(responseOverviewsBobAsBuddyAll, richard, budgetGoalNewsBuddyBob, 1, "Tue")
+		assertResponseStatusOk(responseDetailsBobAsBuddy)
+		def messageText = StringUtils.repeat("*", messageLength)
+		def messageJson = """{"message": "$messageText"}"""
+
+		then:
+		def response = appService.yonaServer.createResourceWithPassword(responseDetailsBobAsBuddy.responseData._links."yona:addComment".href, messageJson, richard.password)
```
aristotelos, line 460: We may extract a method `addComment` here

core/src/main/resources/messages/messages.properties
```diff
@@ -28,14 +28,13 @@ notification.message.body=Tap to open message
 error.invalid.request=Invalid request
 error.invalid.uuid=String ''{0}'' is not a valid UUID
 error.invalid.date=String ''{0}'' is not a valid date
-error.invalid.date.too.early=Date ''{0}}'' is too early. Earliest possible date is ``{1}}``
+error.invalid.date.too.early=Date ''{0}'' is too early. Earliest possible date is ``{1}``
```
aristotelos, line 31: Good catch... Another one: I see two ticks `''` vs. two back-ticks ````, which we may correct too

aristotelos (commented):
Looks great, added a suggestion

appservice/src/intTest/groovy/nu/yona/server/ActivityCommentTest.groovy
Bert-R: Good point. Done

core/src/main/resources/messages/messages.properties
Bert-R: There were a few more of these. Replaced them all

Bert-R:
Pushed an update

aristotelos (approved):

PR #601, opened 2020-05-20
YD-690 Added SMTP connection timeouts

Bert-R:
For both SMTP and SMTPS, all set to three minutes. From infinite to 3 minutes is a significant change. If need be, we reduce the values in a next round.


Bert-R:
Going to merge this. After the timeout has proven to work, the extra logging will be removed.

PR #610, opened 2021-01-12
YD-699 Correct ISO week number calculation

Bert-R:
The week number was wrongly calculated for the week of 30-Dec-2020 (2021-W01 instead of 2020-W53).

Along with this: added a delete for Bea that was forgotten earlier.


Bert-R (commented):
Going to merge this now, so the build can pass again.

PR #612, opened 2021-02-13
YD-695 Switch to Read Committed

Bert-R:
Now use Read Committed as transaction isolation level.

Introduced ``OverwriteUserTest.'Concurrent requests cause no errors'``. This consistently failed before switching the isolation level and now consistently passes. Generalized the method ``postNTimes`` of ``PinResetRequestTest``.

Logged YD-702 to remove the extra debug logging after the fix has proven itself for a couple of months.


aristotelos (approved):
Looks great!

PR #691, opened 2021-06-05
YD-708 Address Java compiler warnings

Bert-R:


aristotelos (approved):
Looks good, will merge. To me it is a shock that Java does not have true generic class literals :wink:

Bert-R:
> To me it is a shock that Java does not have true generic class literals

Yes, that's a pity. It's because of [type erasure](https://www.baeldung.com/java-type-erasure).

